PR #33, opened 2019-10-20
buildsystem improvements

swick:
* allow building either with shared libraries or static libraries
  (default) with -DBUILD_SHARED_LIBS=on/off
* improve running from within the build directory in linux using a
  wrapper shell script (bin/Gothic2Notr.sh)
* don't set CMAKE_BINARY_DIR so the game can always be run from the
  build directory
* allow installing
* some of the features require a changed MoltenTempest

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,8 @@ cmake_minimum_required(VERSION 2.8)
 project(Gothic2Notr)
 set (CMAKE_CXX_STANDARD 14)
 
-include_directories("Game")
+option(BUILD_SHARED_LIBS "Build shared libraries." OFF)
+option(BUILD_SHARED_MOLTEN_TEMPEST "Build shared MoltenTempest." ${BUILD_SHARED_LIBS})
 
 set(CMAKE_DEBUG_POSTFIX "")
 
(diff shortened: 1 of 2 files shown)


swick:
Why shared by default? Without these changes the only shared library is MoltenTempest, everything else is statically linked. When we change the default to ON with this patch all dependencies are getting linked dynamically. Is that what you want?

Try:
>  Is that what you want?
Yes - link behavior should not be changed at all

swick:
But right now *only* MoltenTempest is being linked dynamically, everything else is being linked statically. Is that intended?

swick:
New version pushed. Notable change:

* allow building individual dependencies as shared libraries (currently only MoltenTempest) by setting `-DBUILD_SHARED_${PROJECT_NAME}=on/off`

Try:
I've test new CMakeLists.txt
As-is build fails with linking -lvulkan1 to Gothic2Notr.exe - this is Tempest issue, fixed in fd247d3
Also need to add `-DBUILD_SHARED_MOLTEN_TEMPEST:BOOL=ON` appveyor.yml script
Try to rebase your patch on most recent version of OpenGothic, to pass buildcheck

swick:
rebased but the appveyor build fails. I don't know anything about submodules and appveyor.

Try:
OK, since CI is red already I'm going to merge this request and deal with CI separately

PR #35, opened 2019-10-30
build system cleanup part two

swick:
* compile shaders as part of the cmake build
* get rid of the bin/shader directory (containing prebuild shaders)
* move bin/data/ to data/
* move data/run.bat to scripts/Gothic2Notr.bat to mirror linux
* finally get rid of bin/
* copy scripts/Gothic2Notr.bat to binary directory to mirror linux
* copy data/ to binary directory
* adjusted appveyor.yml to generate the artifact from the binary
  directory

The result is that the binary directory now contains everything required
to start OpenGothic from there while the source directory stays
completely clean.


swick:
Alright, the CI is now working for windows and linux. The artifact for linux is working (but not running due to other issues), the artifact for Windows looks correct but I can't test it.

CMakeLists.txt
```diff
@@ -70,33 +68,196 @@ set(GLFW_BUILD_EXAMPLES OFF CACHE STRING "" FORCE)
...
 # MoltenTempest
 set(BUILD_SHARED_LIBS ${BUILD_SHARED_MOLTEN_TEMPEST})
 add_subdirectory(lib/MoltenTempest/Engine)
 
+
+## OpenGothic
+
+# sources
+# FIXME: recursive CMakeLists.txt instead of global listing
+set(OPENGOTHIC_SOURCES
```
Try: >  recursive CMakeLists.txt instead of global listing

OK, but why, not to keep wildcards for now?

CMakeLists.txt
```diff
@@ -70,33 +68,196 @@ set(GLFW_BUILD_EXAMPLES OFF CACHE STRING "" FORCE)
...
+    Game/world/interactive.cpp
+    Game/world/npc.cpp
+    Game/world/world.cpp
+    Game/world/worldsound.cpp
+    Game/camera.cpp
+    Game/gothic.cpp
+    Game/mainwindow.cpp
+    Game/resources.cpp)
+
+# shaders
```
Try: Since compilation of shader moved from *.bat, it make sense to move commentary for ubershading:
```
# Ubershader flags:
#   OBJ        - enable object matrix
#   SKINING    - animation skeleton
#   SHADOW_MAP - output is shadowmap
#   ATEST      - use alpha test
#   PFX        - use color modulation
```

appveyor.yml
```diff
@@ -7,7 +7,7 @@ configuration:
 
 before_build:
   # - sh: sudo apt-get install libvulkan-dev
-  - sh:  sudo apt-get --yes install libgl-dev
+  - sh:  sudo apt-get --yes install libgl-dev libvulkan-dev
```
Try: Was this a reason for weird link/rpath error lately on appveyor?
I've tried many times to reproduce it locally on Ubuntu18 virtual machine, with no success.

swick: Yeah. Not sure if the SDK was supposed to provide the .so somehow but installing to the system works for sure.

CMakeLists.txt
swick: Alright, will revert for now.

CMakeLists.txt
swick: Sure thing.

swick:
Done.

PR #41, opened 2020-01-23
looting speed now increases when holding action button

da3m0nsec:
Items looted increase x10 each 10 loots.

### DIFF
--- a/Game/ui/inventorymenu.cpp
+++ b/Game/ui/inventorymenu.cpp
@@ -230,16 +230,17 @@ void InventoryMenu::mouseDownEvent(MouseEvent &e) {
     if(r.isEquiped())
       player->unequipItem(r.clsId()); else
       player->useItem    (r.clsId());
-    }
-  else if(state==State::Chest || state==State::Trade || state==State::Ransack) {
-    takeTimer.start(100);
-    onTakeStuff();
-    }
-  adjustScroll();
   }
+  else if(state==State::Chest || state==State::Trade || state==State::Ransack) {
+    takeTimer.start(200);
+    onTakeStuff();
+  }
+  adjustScroll();
+}
 
 void InventoryMenu::mouseUpEvent(MouseEvent&) {
   takeTimer.stop();
+  takeCount=0;
   }
 
 void InventoryMenu::mouseWheelEvent(MouseEvent &e) {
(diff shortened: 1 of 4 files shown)


da3m0nsec:
Sure! Will work on it next week when I have some spare time :)

da3m0nsec:
Called ```e.accept()``` on ```InventoryMenu::keyDownEvent``` because keyUpEvents weren't getting triggered on traders for not being accepted:
```cpp
void MainWindow::keyDownEvent(KeyEvent &event) {
   if(inventory.isActive()){
      inventory.keyDownEvent(event);
      if(event.isAccepted()){
         uiKeyUp=&inventory;
         return;
         }
   }
   ...
   }
void MainWindow::keyUpEvent(KeyEvent &event) {
   if(uiKeyUp==&inventory){
       inventory.keyUpEvent(event);
       if(event.isAccepted())
         return;
       }
   }
   ...
   }
```
Maybe that's not the intended use but it's the workaround I found.

Game/ui/inventorymenu.cpp
```diff
@@ -204,11 +205,30 @@ void InventoryMenu::keyDownEvent(KeyEvent &e) {
...
+    lootMode = LootMode::Stack;
+    takeTimer.start(200);
+    onTakeStuff();
+    }
   adjustScroll();
   update();
   }
 
 void InventoryMenu::keyUpEvent(KeyEvent &e) {
+  if(e.key==KeyEvent::K_Space || e.key==KeyEvent::K_Z || e.key==KeyEvent::K_X){
```
Try: Is this check truly necessary?
The keyUpEvent will be called, only if previous keyDownEvent was accepted - it's save to call takeTime.stop() unconditionally

Game/ui/inventorymenu.cpp
```diff
@@ -308,34 +330,59 @@ InventoryMenu::PageLocal &InventoryMenu::activePageSel() {
...
+void InventoryMenu::onTakeStuff() { 
+  uint8_t itemCount = 0;
   auto& page = activePage();
-  auto& sel  = activePageSel();
-  if(sel.sel>=page.size())
+  auto& sel = activePageSel();
+  if(sel.sel >= page.size())
     return;
   auto& r = page[sel.sel];
+  if(r.isEquiped()) {
```
Try: This check won't work as expected, if r.count()>1.

da3m0nsec: Yeah.. It can move all the stack until r.count()==1, so it doesn't touch _the one_ equipped. For sure the best would be to separate them in two stacks but don't know how.

Try:
Remove isEquiped check for now. I think, it's better to implement this check as separate PR

PR #46, opened 2020-02-03
fixed possible overflow InventoryMenu

da3m0nsec:
Now Item::count() returns the same type as Daedalus::GEngineClasses::C_Item.amount

### DIFF
--- a/Game/world/npc.cpp
+++ b/Game/world/npc.cpp
@@ -2014,7 +2014,7 @@ void Npc::buyItem(uint32_t id, Npc &from, uint32_t count) {
   if(price*count>int32_t(invent.goldCount())) {
     owner.script().printCannotBuyError(*this);
     if(int32_t(invent.goldCount())/price >= 1){
-      buyItem(id,from,uint32_t(invent.goldCount())/price);
+      buyItem(id,from,int32_t(invent.goldCount())/price);
       }
     return;
     }
--- a/Game/world/npc.cpp
+++ b/Game/world/npc.cpp
@@ -2014,7 +2014,7 @@ void Npc::buyItem(uint32_t id, Npc &from, uint32_t count) {
   if(price*count>int32_t(invent.goldCount())) {
     owner.script().printCannotBuyError(*this);
     if(int32_t(invent.goldCount())/price >= 1){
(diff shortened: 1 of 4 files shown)


Game/world/npc.cpp
```diff
@@ -2014,7 +2014,7 @@ void Npc::buyItem(uint32_t id, Npc &from, uint32_t count) {
   if(price*count>int32_t(invent.goldCount())) {
     owner.script().printCannotBuyError(*this);
     if(int32_t(invent.goldCount())/price >= 1){
-      buyItem(id,from,int32_t(invent.goldCount())/price);
+      buyItem(id,from,uint32_t(invent.goldCount())/price);
```
Try: Unrelated change.
Right now I'm working on Werror+Wconversion and already touch this function

da3m0nsec:
Certainly there is a problem for using both uint32_t and size_t.
If itemCount would be a size_t, it would be casted down to uint32_t anyway on setCount().
Do you think it is OK like this? Or should we use size_t on addItem/moveItem/buyItem/sellItem?

Try:
> Certainly there is a problem for using both uint32_t and size_t.

True, my current thought about this problem:
  * size_t for arrays
     Reason: this is 'native' type for array indexing in run-time
  * size_t for 'amount' of stuff.  
     Reason: often can be used as array index
  * size_t for script-symbols.
     Reason: symbols are stored in array
  * uint32_t for serialization of size_t 
     Reason: compatibility between x86 and x64 saved games
  * int32_t - any integer from script side
     Reason: compatibility with original gothic 
  * ZString - any string from script side
     Reason: special type; copy on write
  * int32_t prices(amount of gold/ore)
     Reason: integer from script side
  * int64_t for native time 
     Reason: 64 - large; sign - easier math
  * gtime for game-time
     Reason: special type
  * int - any UI related integer, widow coordinates for instance
     Reason: easier math
  * float - any floating point, unless double is really needed
     Reason: right now, double is never used in code anyway

Right now I'm working on -Wconversion + Werror, to make type checking more automatic.
On next step I'll add some coding guide with info about types + clang format(or something like it)

Try:
Upd: master is green now - you can safely rebase your change

PR #54, opened 2020-04-25
51 case sensitiveness issue on linux

### LINKED ISSUE #51: Case sensitiveness issue on linux
Hello.

First of all: Wow. This is an amazing project. Thank you for this.

I tried to compile and start the project on a linux machine.
After installing _CMake_ 3.12 from a third party repo (kitware) and installing _glslang-tools_ it built very smooth with only one warning so far. Nice.

After starting a new game, it turned out, that the player controls were'nt working. I debugged a bit and discovered that
```cpp
std::unique_ptr<IniFile>                baseIniFile;
std::unique_ptr<IniFile>                iniFile;
```
couldn't be found. Therefore all control keys were unset.
The reason, why the files could'nt be found, was the following:
In my installation (German CDs: Gothic II Normal + Addon) the folder and ini file was named in different case:
"_System/Gothic.INI_" instead of "_system/Gothic.ini_".
After I renamed the folder and the file, the player controls has been working.

For a general solution, I digged myself down in the code and made the following observations:

1. I found pieces of code, that might be required to solve the problem in the following places:
- Game/utils/fileutil.h: exists
- Game/utils/fileext.h: hasExt, exchangeExt
- Game/gothic.h: caseInsensitiveSegment
- lib/MoltenTempest/Engine/utility/dir.h: scan

2. String conversions are done at various places and the usage seems to have been mixed up sometimes (I've seen so far: char*, std::string, std::u16string, unicode 8 conversion being used as std::wstring). That's just a subjective impression.

I want to offer you the following changes, but want to ask for your opinions beforehand:
- Implementation of dedicated String classes, that are inheriting from the std::string, std::wstring, etc. and implement support functions, that are required at different places (toUpper, toLower, indexOf, equals, replace, trim, conversion from string/wstring/u16string,u32string to each other string).
- Implementation of a path class, that would be an specialization of the upper listed String class, that handles often required functions: getFilename, getExtension, getFilenameWithExtention, isAbsolute, isRelative, dirUp or concatenation of paths.
- After that an explorer class might be useful, to search for files / directories recursively or not and case sensitive or not. This class might be implemented with std::expermimental::fs features that should be available at >= c++14.

What do you think about that approach ? Do you want to have it done that way ?
The way of not specializing the std::string and using more static support functions might be a good approach, too. It would look more like the std::string instead of yet another custom util string.

I'm curious about your opinions :+1:

ErrorFlexXx:
Fixes #51.

Tested working on Linux XUbuntu 18.04.4 LTS.

### DIFF
--- a/Game/gothic.cpp
+++ b/Game/gothic.cpp
@@ -78,8 +78,8 @@ Gothic::Gothic(const int argc, const char **argv){
     throw std::logic_error("gothic not found!"); //TODO: user-friendly message-box
     }
 
-  baseIniFile.reset(new IniFile(gpath+u"system/Gothic.ini"));
-  iniFile    .reset(new IniFile(u"Gothic.ini"));
+  baseIniFile.reset(new IniFile(nestedPath({u"System",u"Gothic.ini"},Dir::FT_File)));
+  iniFile.reset(new IniFile(nestedPath({u"Gothic.ini"},Dir::FT_File)));
 
   // check actually for gothic-1, any questionable case is g2notr
   if(gpath.find(u"Gothic/")==std::string::npos && gpath.find(u"gothic/")==std::string::npos)
--- a/Game/utils/fileutil.h
+++ b/Game/utils/fileutil.h
@@ -4,5 +4,5 @@
 
 namespace FileUtil {
   bool exists(const std::u16string& path);
-  };
+  }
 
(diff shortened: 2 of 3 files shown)


Try:
Sorry I've already upload a one-line fix...
No need to change next path:
```
iniFile    .reset(new IniFile(u"Gothic.ini")); // this suppose to be *.ini form PWD, with open-gothic specific settings
```
I'll merge the change anyway, to get  changed 27ae483 and adc05f0

PR #58, opened 2020-04-26
Add no debug flag to disable FPS counter

d10sfan:
This is mainly an addition to the possible flags to disable the FPS counter (debug mode). Wasn't sure if the best option would be to by default have it on, since this is an in development project, or vice versa. It appears to work well, with and without the flag, tested on Linux.

### DIFF
--- a/Game/gothic.cpp
+++ b/Game/gothic.cpp
@@ -51,8 +51,8 @@ Gothic::Gothic(const int argc, const char **argv){
     else if(std::strcmp(argv[i],"-nomenu")==0){
       noMenu=true;
       }
-    else if(std::strcmp(argv[i],"-nodebug")==0){
-      noDebug=true;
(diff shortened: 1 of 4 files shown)


Try:
Hi, @d10sfan and thanks for PR!

Can you, please change a switch name to something more concrete? nodebug --> nofrate for example? (nofrate, because original game uses `frate` command to enable fps)
Can you, clarify why you want to have this switch?

d10sfan:
Sure, I've updated it to be nofrate instead and re-tested. 

This was mainly providing a way to play the game without the FPS overlay, whether someone didn't want an FPS counter while playing or wanted to use a 3rd party overlay (like the steam overlay for example, launching it as a non-steam game). I was planning to look into the possibility of making vulkan overlays work, at least for the steam one in the future.

Try:
Merged!

Issue #59, opened 2020-04-26
[Linux] Case sensitiveness while loading music


ErrorFlexXx:
Some sound files can't be found on linux systems, due to case sensitiveness issues.
As the music filenames are coming from Deadalus Scripts, I guess, the engine should deal with that issue.

Example output:
```
unable to load sound: "nw_daystd_A0.sgt"
unable to load sound: "nw_dayfgt.sgt"
unable to load sound: "nw_daystd_A0.sgt"
unable to load sound: "nw_dayfgt.sgt"
```
Files exist here: `Gothic II/_work/Data/Music/NewWorld/NW_DayStd_A0.sgt`.

As a proposal, I made some changes that carry out the Gothic::nestedPath feature to the OpenGothic/utils/fileUitls part and enable the musicLoader to use the case sensitive file search, too.
I'll create a PR for that change, so that it can be discussed in there.

ErrorFlexXx:
Fixed by #61.

PR #61, opened 2020-04-26
[Linux] Case sensitiveness while loading music

ErrorFlexXx:
As announced in issue #59 here is the pull request to enable the engine to load music files regardless the case sensitiveness.


Try:
Thanks for PR!
Merged!

PR #64, opened 2020-04-27
Implemented playercontrol weapon.

ErrorFlexXx:
* Implementation of playercontrol weapon, which toggles between last active weapon and no weapon.
  If the last active weapon is no longer available, a fallback weapon is chosen (Spell -> Range -> Melee)

* Fixed Bug: [FIXME]: Player got freezed, if he was swimming and a weapon shall be drawn.
  movealgo.cpp seems to cover this case. Investigation needed why the player freezes anyway.
  Current workaround: check in playercontrol, if the player is swimming and prevent WeponAction changes in that case.


Try:
I mess-up with git, while doing merge. The code is actually merged by 4ad6a0f.

In same commit I've also changed check of Npc::isSwim, fix nullptr crash in playercontrol.cpp

ErrorFlexXx:
Tested the current master with the fix. Hero doesn't freeze any longer.

PR #68, opened 2020-04-30
Fixed type missmatch to prevent possible linker error

vic-ju:
C_MusicTheme is originally defined as a struct. Clang warned that changing it may cause problems with windows ABI.


Try:
Hi, @vic-ju, and thanks for PR !

C_MusicTheme issue was already fixed as part of a3fec6ef2 (msvc adaptation)

PR #69, opened 2020-04-30
Fixed hero swims with drawn weapons.

ErrorFlexXx:
If the hero starts swimming with a drawn weapon, he won't put it back while starting to swim.
The code somehow states, that the weapon shall be deactivated with no animation.
This doesn't seem to work.

In the original game the weapon magically deactivates without an animation.
This pull request changes the behavior to actually play the put away animations while starting to swim. The animation executes overlayed with the swim animation and it looks pretty good.
In my opinion this is much cleaner than in the original game and I would do it with animation.

### DIFF
--- a/Game/game/movealgo.cpp
+++ b/Game/game/movealgo.cpp
@@ -647,7 +647,7 @@ void MoveAlgo::setInWater(bool f) {
 
 void MoveAlgo::setAsSwim(bool f) {
   if(f)
-    npc.closeWeapon(true);
+    npc.closeWeapon(false);
   if(f)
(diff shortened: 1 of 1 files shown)


Try:
Hi, @ErrorFlexXx !

If player walks in water animation look good, but if player jumps in water,then animation doesn't look correct.
Also I've upload a fix for Npc::closeWeapon(false) implementation - e2c4637, you can rebase.

ErrorFlexXx:
Ok, then I would propose to leave it closing without an animation, as long as #70 isn't solved.

PR #71, opened 2020-04-30
Sneak Implementation

ErrorFlexXx:
- [x] Added sneak key codec reading and added action for playercontroler
- [x] Added toggleSneakMode function to playercontroller.
- [x] Added animation solving for sneak animations.
- [x] Added Bodystate BS_SNEAK, so script NPCs can react to.
- [x] Added canSenseNpc feature to use npcs bodystate sneak for sense_hear detection.
- [x] Allow activation of sneak mode only, if the hero has learned it.
- [x] Fix bug: NPCs react to sneaking, even if CanSee returns false.

I actually created the pull request to ask, if someone has an idea, why NPCs react to sneaking, even if they definitely can't see me. NPCs that are occupied on mobsis cancle their tasks to turn around and react to the sneaking.
I read though the deadalous scripts and maybe I'm blind, but I cant see whats missing here.

### DIFF
--- a/Game/game/playercontrol.cpp
+++ b/Game/game/playercontrol.cpp
@@ -223,7 +223,7 @@ void PlayerControl::toggleSneakMode() {
   if(w==nullptr || w->player()==nullptr)
     return;
   auto pl = w->player();
-  if(0<pl->canSneak())
+  if(pl->canSneak())
     pl->setWalkMode(WalkBit(pl->walkMode()^WalkBit::WM_Sneak));
   }
 
--- a/Game/game/playercontrol.cpp
+++ b/Game/game/playercontrol.cpp
@@ -103,8 +103,13 @@ void PlayerControl::onKeyPressed(KeyCodec::Action a) {
     return;
     }
 
-  if(a==Action::Walk){
+  if(a==Action::Walk) {
     toogleWalkMode();
+    return;
+    }
+
+  if(a==Action::Sneak) {
+    toggleSneakMode();
     return;
     }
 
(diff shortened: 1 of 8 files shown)


ErrorFlexXx:
Ok. The Observe player routine got active from the asses player perception, that got active regardless of the view direction of the npc.

I can't understand the parameter name `freeLos` - Does it mean free line of sight ?
I have trouble to get the meaning in the following context:
```cpp
if(!freeLos){
    float dx  = x-tx, dz=z-tz;
    float dir = angleDir(dx,dz);
    float da  = float(M_PI)*(angle-dir)/180.f;
    if(double(std::cos(da))<=ref)
      if(!w->ray(x,y+180,z, tx,ty,tz).hasCol)
        ret = ret | SensesBit::SENSE_SEE;
    } else {
    // TODO: npc eyesight height
    if(!w->ray(x,y+180,z, tx,ty,tz).hasCol)
      ret = ret | SensesBit::SENSE_SEE;
    }
```
So a ray is done regardless of the parameter freeLOS. Instead only the `field of view` check is dependent on the parameter.

Maybe I just didn't understand the meaning of the abbreviation.

Try:
> I can't understand the parameter name freeLos - Does it mean free line of sight ?

This parameter is there to differentiate between `Npc_CanSeeNpc` and `Npc_CanSeeNpcFreeLOS`.
if, freeLos is true, then function ignores npc view direction.

Try:
Tested - works great.
Add small change to commit: change return value of `canSneak` from int to bool

PR #72, opened 2020-05-01
Added instructions to build on linux

herzenschein:
I checked these dependencies on two clean server VMs.
This should make it easier/faster for contributors to start working on the linux build.
Even if the README.md is not the proper place for this, this info can be added to a BUILD.md or to the wiki later.
I think adding instructions to run it on linux (since Gothic 2 is played through WINE/Proton anyway) would help too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Core gameplay is done, you can complete first chapter, as well as all addon cont
 
 ##### Build it for Linux
 1. Install dependencies for Ubuntu 20.04:
+
 `sudo apt install git cmake g++ glslang-tools libvulkan-dev libasound2-dev libx11-dev`
 
 2. Clone this repo, including submodules:
--- a/README.md
+++ b/README.md
@@ -10,11 +10,28 @@ The goal of this project is to make feature complete Gothic client-app compatibl
 [![Build status](https://ci.appveyor.com/api/projects/status/github/Try/opengothic?svg=true)](https://ci.appveyor.com/project/Try/opengothic)  
 Core gameplay is done, you can complete first chapter, as well as all addon content for any guild. 
 
-##### How to play
+##### Install on Windows
 1. Install original gothic game from CD/Steam/GOG/etc  
 *you have to install original game, since OpenGothic does not have any game assets or game scripts as built-in*
 2. [Download latest stable build](https://github.com/Try/opengothic/releases/latest)
 3. run '/OpenGothic/bin/Gothic2Notr.exe -g "C:\Program Files (x86)\Path\To\Gothic II"'
+
+##### Build it for Linux
+1. Install dependencies:
+
+* for Ubuntu:
+`sudo apt install cmake g++ doxygen glslang-tools libgl-dev`
+
+* for openSUSE:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,12 +10,29 @@ The goal of this project is to make feature complete Gothic client-app compatibl
...
 
-##### How to play
+##### Install on Windows
 1. Install original gothic game from CD/Steam/GOG/etc  
 *you have to install original game, since OpenGothic does not have any game assets or game scripts as built-in*
 2. [Download latest stable build](https://github.com/Try/opengothic/releases/latest)
 3. run '/OpenGothic/bin/Gothic2Notr.exe -g "C:\Program Files (x86)\Path\To\Gothic II"'
 
+##### Build it for Linux
+1. Install dependencies:
```
Try, line 20: I want to simplify this to:
1. Install dependencies:
`cmake g++ gcc-c++ doxygen glslang-tools libvulkan-dev libasound2-dev`

README.md
```diff
@@ -10,12 +10,29 @@ The goal of this project is to make feature complete Gothic client-app compatibl
...
+
+* for Ubuntu:
+`sudo apt install cmake g++ doxygen glslang-tools libgl-dev`
+
+* for openSUSE:
+`sudo zypper in git cmake gcc-c++ glslang-devel doxygen Mesa-devel`
```
Try, line 26: Doxygen for opensuse? You don't really need it to compile code, right?

Library `libasound2-dev` also reqired to build sound.
`libvulkan-dev` is not in list, what is odd for Vulkan project.
`libgl-dev`, 'Mesa-devel' - I thought, redundant dependency from those libraries was fixed a sometime ago. Do we still need those?

README.md
herzenschein: I'm not sure it's a good idea. Now that I think about it, I think the opposite should be done: specifying that this is not only for Ubuntu, but for Ubuntu 20.04. The same with openSUSE, for which I used Tumbleweed.

README.md
herzenschein: I'll test this section on a clean VM later, but it's true that with these build dependencies the cmake command does not output "not found".
On Tumbleweed I believe Mesa-devel pulls all the required openGL libraries, but I'll test it again.

Try: I mean: since project not using opengl, there not suppose to be any gl-packages, right?
The engine requires glslangValidator, but it suppose to be included in vulkan-sdk,

README.md
Try: Motivation for one list is simplicity. It 3 time easier to maintain one list instead of three.
Another thing, what I'm worry about is approach. Testing on each type of linux is very time consuming, why not just build a package list based on actual dependencies?

ErrorFlexXx:
For me on XUbuntu the glslangValidator is included   in the package glslang-tools in a third party graphic driver repo. It belongs to opengl build tools.
On Arch Linux, I saw it as glslang package.
I prefer installing packages from repositories instead of binary installers, so I chose those glslang package and libvulkan-dev and it is working for me.

herzenschein:
I removed the opensuse build dependencies and left only those for ubuntu. I just confirmed it builds, so it can be merged.

Try:
Merged!

PR #73, opened 2020-05-02
Fix broken sit ani at vectors.

### LINKED ISSUE #43: Npc_CanSeeNpc returns wrong value, when npc sitting on a bench.
Seem like Npc_CanSeeNpc should use BIP01 rotation, to calculate eye-dir vector.
B_AssessTalk :
```
		if(Npc_CanSeeNpc(self,other))
		{
			AI_StartState(self,ZS_Talk,0,"");
		}
		else
		{
			if(Npc_IsInState(self,ZS_ObservePlayer))
			{
				AI_Standup(self);
			};
			AI_StartState(self,ZS_Talk,1,"");
		};
```

ErrorFlexXx:
Fixes #43.
Update of the readme file (hints for common Gothic installation paths and 
changed sneak implementation status: Done.

There are two animations with broken atVectors (pointing in the opposite direction).

Affected animations:
- S_BENCH_S1
- S_THRONE_S1

All other animations aren't affected by this.

### DIFF
--- a/Game/graphics/mdlvisual.cpp
+++ b/Game/graphics/mdlvisual.cpp
@@ -491,15 +491,24 @@ Tempest::Vec3 MdlVisual::displayPosition() const {
   return {0.f,0.f,0.f};
   }
 
-Vec3 MdlVisual::lookAtVector(const char *nodeName) const
+bool MdlVisual::boneRotationY(const char *nodeName, float& deg) const
 {
-  if(nullptr==skeleton) {
-    return Vec3(0.f, 0.f, 0.f);
-  }
-  size_t nodeId = skeleton->findNode(nodeName);
-  Matrix4x4 trafo = skInst->tr[nodeId];
-  Vec3 atVector = Vec3(trafo.at(0,0),trafo.at(0,1),trafo.at(0,2));
-  return atVector;
+  size_t nodeId = size_t(-1);
+
+  if(nullptr!=skeleton)
+    nodeId = skeleton->findNode(nodeName);
+
+  if(size_t(-1)!=nodeId) {
+    auto p = pos;
+    p.mul(pose().tr[nodeId]);
+    float atVecX = p.at(2,0);
+    float atVecZ = p.at(2,2);
+    if(0.f!=atVecX) {
+      deg = static_cast<float>(atan(atVecZ / atVecX)) * 180.f / static_cast<float>(M_PI);
+      return true;
+      }
+    }
+  return false;
 }
 
 const Animation::Sequence* MdlVisual::continueCombo(Npc& npc, AnimationSolver::Anim a,
--- a/Game/graphics/mdlvisual.cpp
+++ b/Game/graphics/mdlvisual.cpp
@@ -491,8 +491,7 @@ Tempest::Vec3 MdlVisual::displayPosition() const {
   return {0.f,0.f,0.f};
   }
 
-float MdlVisual::viewDirection() const
-{
+float MdlVisual::viewDirection() const {
   auto p = pos;
   if(nullptr!=skeleton) {
     size_t nodeId = skeleton->findNode("BIP01");
@@ -501,8 +500,8 @@ float MdlVisual::viewDirection() const
     }
   float rx = p.at(2,0);
   float rz = p.at(2,2);
-  return static_cast<float>(atan2(rz,rx)) * 180.f / static_cast<float>(M_PI);
-}
+  return float(std::atan2(rz,rx)) * 180.f / float(M_PI);
+  }
 
 const Animation::Sequence* MdlVisual::continueCombo(Npc& npc, AnimationSolver::Anim a,
                                                     WeaponState st, WalkBit wlk)  {
--- a/Game/graphics/mdlvisual.cpp
+++ b/Game/graphics/mdlvisual.cpp
@@ -491,24 +491,17 @@ Tempest::Vec3 MdlVisual::displayPosition() const {
   return {0.f,0.f,0.f};
   }
 
-bool MdlVisual::boneRotationY(const char *nodeName, float& deg) const
+float MdlVisual::viewDirection() const
 {
-  size_t nodeId = size_t(-1);
-
-  if(nullptr!=skeleton)
-    nodeId = skeleton->findNode(nodeName);
-
-  if(size_t(-1)!=nodeId) {
-    auto p = pos;
-    p.mul(pose().tr[nodeId]);
-
-    float rx = p.at(2,0);
-    float rz = p.at(2,2);
-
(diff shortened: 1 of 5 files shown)


Game/world/npc.cpp
```diff
@@ -3088,6 +3088,10 @@ SensesBit Npc::canSenseNpc(float tx, float ty, float tz, bool freeLos, float ext
   if(!freeLos){
     float dx  = x-tx, dz=z-tz;
     float dir = angleDir(dx,dz);
+    if((bodyState()&BodyState::BS_SIT)!=0 &&    //If we are sitting with
+       (visual.pose().isInAnim("S_BENCH_S1") || //this ani on a bench or
```
Try, line 3092: I don't think hard-coding animation for this case is a correct approach.
As mention in issue:
> Seem like Npc_CanSeeNpc should use BIP01 rotation, to calculate eye-dir vector

ErrorFlexXx: I don't know what BIP01 rotation is.
I'm not that familar with the animation system in detail, yet.
Is it hard to do ?

I know, this here is just a quick fix to get it working as fast as possible ;-)
Shame on me.

Try: In Gothic, every bone has a name, "BIP01" is one of them.

It's possible to query a bone index by calling `Skeleton::findNode("BIP01")`
Once you havea `id`, it's possible to fetch a transformed bone matrix:
`Matrix4x4 tr = skInst->tr[id]`
Most-likely this matrix suppose to be used, to deduce a 'real' rotation of npc.

You can find more details about bone-usage in mdlvisual.cpp

ErrorFlexXx: Ok, if I broke it down correctly, here is my proposal how to get the head rotation:
npc [Npc] ->visual [MdlVisual]->head [Mesh] -> protoMesh() -> Node.transform
- Extract rotations from ObjectMatrix to get an atVector (sub Translation, unapply scale, build rotation matrix, convert to quaternion, get rotations as vector).
If I understand the Bone animation things correctly, the ObjectMatrix of the Head Mesh should include all transformations from the Base meshes, or ? Otherwise this might have to be done for all Base meshes, too, to get the final rotation of the head.

Downside of this approach, even if this is more correct, it's expensive and might introduce "bugs" during gameplay (npcs field of view changes). NPCs that might not have seen the hero before, can now see him related to the animations of the character. But, it might not be a problem.

What do you thing about that ?

ErrorFlexXx: But I guess, as we would have to change the code to get the angle in CanSenseNpc anyway, the temporary workaround with the animations wouldn't be forgotten to be removed later.
So as a temporary workaround, it could do the job, maybe flagged with a fixme comment to replace it with the real head rotation, if implemented.

Try: You don't have to do so much work to fetch the transform, because they already calculated:
`Pose::tr' is a ready-to use matrices, in object space. Then to get a view vector:
```c++
  auto p = pos;
  if(boneId<pose.tr.size())
    p.mul(pose.tr[boneId]);
  Vec3 viewDir = Vec3(p.at(0,0),p.at(0,1),p.at(0,2));
``` 
Note: I don't know for sure which axis is used as a 'forward', but most-likely it's X-axis.

Not about possible bugs: idea is to pick a 'root' bone, aka BIP01, to use instead of 'BIP01 HEAD'. That way implementation can be more reliable.

ErrorFlexXx:
For me, it works.

ErrorFlexXx:
So current state is: The calculated rotation is wrong.

I print out the difference between Npc::rotation() and the now calculated BIP01 Bone roatation and from my understanding, the difference should be close to zero in most cases, except the sit animation.
If I follow an NPC that walks around, I see the difference between the rotation changes. It doen't look like pos * tr would represent the overall rotation of the character.

Example Log for Khorinis Temple square:
```
Info: Bone rotation diff: 180.005
Info: Bone rotation diff: 0.31506
Info: Bone rotation diff: 0.98671
Info: Bone rotation diff: 0.00502777
Info: Bone rotation diff: 180.005
Info: Bone rotation diff: 0.00405121
Info: Bone rotation diff: 180.005
Info: Bone rotation diff: 180.005
Info: Bone rotation diff: 0.00501633
Info: Bone rotation diff: 0.00502777
Info: Bone rotation diff: 0.00502014
Info: Bone rotation diff: 180.005
Info: Bone rotation diff: 0.00405121
```

Game/graphics/mdlvisual.cpp
```diff
@@ -491,15 +491,24 @@ Tempest::Vec3 MdlVisual::displayPosition() const {
...
-  return atVector;
+  size_t nodeId = size_t(-1);
+
+  if(nullptr!=skeleton)
+    nodeId = skeleton->findNode(nodeName);
+
+  if(size_t(-1)!=nodeId) {
+    auto p = pos;
+    p.mul(pose().tr[nodeId]);
+    float atVecX = p.at(2,0);
```
ErrorFlexXx, line 504: Interesting, too:
If I change this to 
```cpp
x = p.at(0,0)
z = p.at(0,2)
```
which should be the right vector, I get a difference around 90° for all NPCs (including the one with the bench animation).

ErrorFlexXx:
Now it's working. I compared the rotation() from the Npc with the bone ones like above.
For all standing NPCs the difference is around 0-5° (dependent on the current animation) and for all Bench sitting guys at 180-184° (checked 3 different NPCs sitting on a bench).
Seems stable.

Game/graphics/mdlvisual.cpp
```diff
@@ -491,6 +491,26 @@ Tempest::Vec3 MdlVisual::displayPosition() const {
   return {0.f,0.f,0.f};
   }
 
+bool MdlVisual::boneRotationY(const char *nodeName, float& deg) const
```
Try, line 494: Design with return code for this function doesn't look good to me.
Can we use a `pos` matrix as-is, if bone not found?

Game/world/npc.cpp
```diff
@@ -3088,7 +3088,9 @@ SensesBit Npc::canSenseNpc(float tx, float ty, float tz, bool freeLos, float ext
   if(!freeLos){
     float dx  = x-tx, dz=z-tz;
     float dir = angleDir(dx,dz);
-    float da  = float(M_PI)*(angle-dir)/180.f;
+    float boneRotation = rotation(); //Fallback is NPC rotation.
+    (void)visual.boneRotationY("BIP01", boneRotation);
```
Try, line 3092: This line look confusing, as suggested above `boneRotationY` can perform fallback by itself, so return code is not needed.

ErrorFlexXx:
Yes, looks much cleaner now :+1:

PR #79, opened 2020-07-03
Update Gothic2Notr.sh

Sebbestune:
The lib can now be reached (and game started) even when running from e.g. /build/openGothic instead of /bin.


Sebbestune:
This would work as long as the name of the main folder remains "OpenGothic". When thinking about it a more desirable approach might be something that allows a name change as well.

PR #80, opened 2020-07-03
Update Gothic2Notr.sh

Sebbestune:
A simpler version that do not require an unchanged project name. This works when testing to start the game directly from build folder in Linux.

### DIFF
--- a/scripts/Gothic2Notr.sh
+++ b/scripts/Gothic2Notr.sh
@@ -1,8 +1,17 @@
 #!/bin/bash
+
+function cdroot()
+{
+  cd ..; 
+  echo ${PWD}
+}
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-export LD_LIBRARY_PATH="$DIR/lib:$LD_LIBRARY_PATH"
+
+
+export LD_LIBRARY_PATH=(cdroot)+"/lib:$LD_LIBRARY_PATH"
 if [[ $DEBUGGER != "" ]]; then
-  exec $DEBUGGER --args "$DIR/bin/Gothic2Notr" "$@"
+  exec $DEBUGGER --args "$DIR/Gothic2Notr" "$@"
 else
-  exec "$DIR/bin/Gothic2Notr" "$@"
+  exec "$DIR/Gothic2Notr" "$@"
 fi


scripts/Gothic2Notr.sh
```diff
@@ -1,8 +1,17 @@
...
+  echo ${PWD}
+}
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-export LD_LIBRARY_PATH="$DIR/lib:$LD_LIBRARY_PATH"
+
+
+export LD_LIBRARY_PATH=(cdroot)+"/lib:$LD_LIBRARY_PATH"
 if [[ $DEBUGGER != "" ]]; then
-  exec $DEBUGGER --args "$DIR/bin/Gothic2Notr" "$@"
```
Try, line 5: CMake build script suppose to put all outputs in /opengothic, including all *.so, that means `export LD_LIBRARY_PATH=$DIR:$LD_LIBRARY_PATH"`, is enough and `cdroot` is not required at all

Sebbestune:
Hi @Try !
For me I don't get the /bin/ structure after a fresh compile under (Manjaro 20.0.3) Linux. 
image

The files I can see are: Gothic2Notr, Gothic2Notr.sh and some library files directly inside build/opengothic. One step up we can see lib and other files and folders. 

When running:

`./Gothic2Notr -g "/run/media/sebastian/4212589A125894AF/Program Files (x86)/Steam/steamapps/common/Gothic II/"` 
I get
`./Gothic2Notr: error while loading shared libraries: libMoltenTempest.so: cannot open shared object file: No such file or directory`

And running:
`./Gothic2Notr.sh -g "/run/media/sebastian/4212589A125894AF/Program Files (x86)/Steam/steamapps/common/Gothic II/"`
I get
`./Gothic2Notr.sh: rad 7: /home/sebastian/Dokument/OpenGothic/build/opengothic/bin/Gothic2Notr: Filen eller katalogen finns inte` (file or catalogue does not exist)

When I tried using the lib folder in the parent directory everything runs like it should. Maybe there is some other issue such as that /bin/ should in fact be generated also in the Linux build and the lib files should not result in "cannot open". Then this change can be discarded and we can look at how to generate a valid structure.

See this video for more details: https://youtu.be/az2XaKjKxBg
0:00 - Git clone project
5:03 - Make project
14:27 - Showcasing this problem
19:08 - OpenGothic running (The issues with SIGABRT seen before seem to have been because of a mounting issue since I have my gothic files on a windows partition and opengothic on a linux partition on a dual booted system. I did not see them before but of course they show up when I am recording :P) 
Gothic shows up perfectly for me but not in the recording unfortunately.

Try:
> The files I can see are: Gothic2Notr, Gothic2Notr.sh and some library files directly inside build/opengothic

Yes, that what I was talking about: in this catalog you have everything you need to run the game. 
So, in principle to run the game you can:
```bash
cd build/opengothic
LD_LIBRARY_PATH=.
./Gothic2Notr -g "/path/to/gothic/"
```
And Gothic2Notr.sh only wrap this 2 lines, no complex logic, like `cdroot` is needed

Sebbestune:
Aha! Now I see what you mean and cdroot would in this case not be needed as "../" would be sufficient. I think I was stuck on the cdroot idea mostly because I primarily had a different solution in mind and then didn't really mentally detach from that solution. 

Changing the file to:
```
#!/bin/bash

DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
export LD_LIBRARY_PATH="../lib:$LD_LIBRARY_PATH"
if [[ $DEBUGGER != "" ]]; then
  exec $DEBUGGER --args "$DIR/Gothic2Notr" "$@"
else
  exec "$DIR/Gothic2Notr" "$@"
fi
```
...works well. 

Also at least for me the files in the same folder does not seem to work. I have only gotten the program to start using the ".sh" file.
image

I used to have this problem also when using the .sh file until I changed it, but after testing again today it seems like all issues are gone just by removing the bin/ part of the script. 

I will change the PR to only remove the /bin. 

Thanks for the patience!

PR #83, opened 2020-07-18
Fixed paths Gothic2Notr.sh

smaTc:
Output directory does not have subdirectories; removed `/bin` and `/lib` as all libraires are in current `$DIR`


Try:
Thanks for the fix!
Merged

PR #84, opened 2020-08-10
config file support added

smaTc:
Added support for a basic config file ( called `gothic2.conf`) for the command line options. Must be in executable directory.

gothic2.conf Example:

```
gamePath=/path/to/game
noMenu=false
noFramerate=true
windowMode=false
ramboMode=true
renderer=vulkan
```

### DIFF
--- a/Game/gothic.cpp
+++ b/Game/gothic.cpp
@@ -17,15 +17,76 @@
 #include "utils/installdetect.h"
 #include "utils/fileutil.h"
 #include "utils/inifile.h"
+#include <string>
+#include <fstream>
+#include <iostream>
+
 
 using namespace Tempest;
 using namespace FileUtil;
 
 Gothic::Gothic(const int argc, const char **argv) {
-  if(argc<1)
-    return;
 
-  for(int i=1;i<argc;++i){
+  std::ifstream runningConfig("gothic2.conf");
+
+  if (runningConfig.is_open() && runningConfig.good()) {
+    std::cout << "Config file found!" << std::endl;
+    std::string line;
+    while (std::getline(runningConfig,line)){
+      std::string option;
+      std::string value;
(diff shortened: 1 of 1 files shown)


Try:
Hi, @smaTc and thanks for PR!

I love the idea, to move command line settings into a file, but unfortunately this implementation won't fit project.
There are 2 major things to take into account:
1. OpenGothic already has *.ini file parser - no need to introduce a new format
2. There is already support for settings file in executable path:
   1. OpenGothic read 'Gothic.ini', from exe-directory
   2. OpenGothic read 'Gothic.ini', from /path/to/gothic/system
   3. Values in 'Gothic.ini', from exe-directory has a priority

My suggestion how to change your path:
1. change from custom parser to Gothic::baseIniFile
2. reserve a section in *.ini, for example `[OPEN_GOTHIC]`
3. fetch settings, like: `baseIniFile->getS("[OPEN_GOTHIC]", "game_path")`

smaTc:
Hey @Try. OK. I am not sure which instance is the right one for the ini File in the executable directory or how I correctly instatiate a new ini file. Can you elaborate what to do a bit further?

Try:
Hi, @smaTc !
About correct ini-file: simple answer: `Gothic::baseIniFile`
This is the one is corresponding for "Gothic.ini" in  executable path.

smaTc:
@Try  Well then I understood your first answer correctly. I tried that and it only gives me `segmentation fault (core dumped)` because at this point `Gothic::baseIniFile` is not initialized and the validation via `validateGothicPath()` stops execution before initialization. That is why I asked how to instantiate a new *.ini File correctly, because I don't want to break anything by changing to much without knowing the consequences.

Try:
Closing this PR - since it's not progressing

PR #88, opened 2020-10-17
allow up/down arrows for dialog selection

Nindaleth:
up/down arrows are used (in addition to W/S keys) for changing of the active selection in most of the GUI, with the exception of dialog option selection; this PR adds that possibility


Try:
Hi @Nindaleth !
Thanks for PR, merged!

PR #94, opened 2020-10-22
enable CrashLog::dumpStack outside Windows

Nindaleth:
This seems like the cleanest diff right now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,7 @@ endif()
 if(WIN32)
   target_link_libraries(${PROJECT_NAME} shlwapi DbgHelp)
 elseif(UNIX)
-  target_link_libraries(${PROJECT_NAME} -lpthread)
+  target_link_libraries(${PROJECT_NAME} -lpthread -ldl)
 endif()
 
 # ZenLib
@@ -103,6 +103,8 @@ target_link_libraries(${PROJECT_NAME} GothicShaders)
 
 if(NOT MSVC)
   target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wconversion -Wno-strict-aliasing -Werror)
+  # -rdynamic is needed to allow obtaining meaningful backtraces
+  target_link_options(${PROJECT_NAME} PRIVATE -rdynamic)
 endif()
 
 if(WIN32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,8 +103,10 @@ target_link_libraries(${PROJECT_NAME} GothicShaders)
 
 if(NOT MSVC)
   target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wconversion -Wno-strict-aliasing -Werror)
-  # -rdynamic is needed to allow obtaining meaningful backtraces
-  target_link_options(${PROJECT_NAME} PRIVATE -rdynamic)
+  if(UNIX)
+    # -rdynamic is needed to allow obtaining meaningful backtraces
+    target_link_options(${PROJECT_NAME} PRIVATE -rdynamic)
+  endif()
 endif()
 
 if(WIN32)
--- a/Game/utils/crashlog.cpp
+++ b/Game/utils/crashlog.cpp
@@ -136,10 +136,13 @@ std::string CrashLog::demangle(const void* frame, const char* symbol) {
     }
   // couldn't demangle, return the unchanged symbol
   return symbol;
+  #else
+  return "";
   #endif
   }
 
 void CrashLog::tracebackLinux(std::ostream &out) {
+  #ifdef __LINUX__
   // inspired by https://gist.github.com/fmela/591333 (BSD)
   void *callstack[64];
   char **symbols = NULL;
@@ -163,6 +166,7 @@ void CrashLog::tracebackLinux(std::ostream &out) {
       }
     }
     free(symbols);
+  #endif
   }
 
 void CrashLog::writeSysInfo(std::ostream &fout) {
--- a/Game/utils/crashlog.cpp
+++ b/Game/utils/crashlog.cpp
@@ -7,6 +7,11 @@
 #include <csignal>
 #include <fstream>
 #include <cstring>
+#ifdef __LINUX__
+#include <execinfo.h> // backtrace
+#include <dlfcn.h>    // dladdr
+#include <cxxabi.h>   // __cxa_demangle
+#endif
 
 #include <dbg/frames.hpp>
 #include <dbg/symbols.hpp>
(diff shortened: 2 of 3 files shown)


Try:
Overall change look good to me, but can you please clarify a use-case?

Nindaleth:
Well, it could at least save a file with info which signal killed the app and the little bit of system information.

I agree it would be better to extend edd-dbg to work on Linux, so that the `#ifdef` could be dropped. Does it compile and work on OSX/macOS, since edd-dbg advertises support for it?

Nindaleth:
Edd Dawson said on his website that porting edd-db from OSX to Linux should be fairly easy, so I've tried it. I copied the osx dir into linux dir and fixed all compilation warnings and linker warnings. It shouldn't have been able to link on OSX because the library was never provided, not sure about the missing `-ldl`.

[WIP] because it doesn't really dump the stack yet (see below) and I don't know yet what's missing:

```
---crashlog(SIGSEGV)---
GPU: AMD RADV/ACO POLARIS10 (LLVM 10.0.1)
[no call frames available]

Segmentation fault (core dumped)
```

The important changes are into [`debugger_attached`](https://github.com/Try/OpenGothic/pull/94/files#diff-c97c1358aaff342e83471a2c33dc693981ca59307ecbcfcf378f6147b06aefe8R22) and [`frame_ptr_check` constructor](https://github.com/Try/OpenGothic/pull/94/files#diff-de9771f08675223b5b10a426bd09d182412b10ae040bbc988eb9c88fd972490dR49).

Try:
> OSX to Linux should be fairly easy, so I've tried it. 

That's one way to do it, as alternative I want point out to linux 'execinfo.h': 
https://www.tutorialspoint.com/how-to-automatically-generate-a-stacktrace-when-a-gcc-cplusplus-program-crashes

Nindaleth:
Thanks for the hint, the execinfo.h is promising. I've got some basic results already, but it's not demangled yet.

Nindaleth:
Pushed an acceptable initial version. It's much better than nothing, but there are two weaknesses:
* hypothetical for now: the demangling using `__cxa_demangle` is non-standard, but GCC and LLVM support it (not much I can do about that right now, might need some ifdefs in case other niche compilers don't support). I didn't find explicit confirmation or denial, but I've seen some hints that ICC supports this too.
* actual weakness: source filename and line number not present in the output

Here's an example output (forced NULL pointer dereference in inventory window code):
```
---crashlog(SIGSEGV)---
GPU: AMD RADV POLARIS10 (ACO)
#1: InventoryMenu::keyUpEvent(Tempest::KeyEvent&) - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(_ZN13InventoryMenu10keyUpEventERN7Tempest8KeyEventE+0x39) [0xa3b145]
#2: MainWindow::keyUpEvent(Tempest::KeyEvent&) - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(_ZN10MainWindow10keyUpEventERN7Tempest8KeyEventE+0x1b2) [0x9e3ce0]
#3: Tempest::EventDispatcher::dispatchKeyUp(Tempest::Widget&, Tempest::KeyEvent&, unsigned int) - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest15EventDispatcher13dispatchKeyUpERNS_6WidgetERNS_8KeyEventEj+0x125) [0x7fd462b8bb93]
#4: Tempest::SystemApi::dispatchKeyUp(Tempest::Window&, Tempest::KeyEvent&, unsigned int) - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest9SystemApi13dispatchKeyUpERNS_6WindowERNS_8KeyEventEj+0x2d) [0x7fd462b9115d]
#5: Tempest::X11Api::implProcessEvents(Tempest::SystemApi::AppCallBack&) - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest6X11Api17implProcessEventsERNS_9SystemApi11AppCallBackE+0x4e8) [0x7fd462b869b2]
#6: Tempest::X11Api::implExec(Tempest::SystemApi::AppCallBack&) - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest6X11Api8implExecERNS_9SystemApi11AppCallBackE+0x46) [0x7fd462b864c0]
#7: Tempest::SystemApi::exec(Tempest::SystemApi::AppCallBack&) - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest9SystemApi4execERNS0_11AppCallBackE+0x27) [0x7fd462b912b7]
#8: Tempest::Application::exec() - /home/black_fox/src/OpenGothic/build/opengothic/libMoltenTempest.so(_ZN7Tempest11Application4execEv+0x1b) [0x7fd462b88faf]
#9: main - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(main+0x121) [0x9dfd77]
```

First four stack frames are skipped to avoid printing the signal handling parts - the unskipped trace starts like this:
```
#1: CrashLog::tracebackLinux(std::ostream&) - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(_ZN8CrashLog14tracebackLinuxERSo+0x36) [0xa47d56]
#2: CrashLog::dumpStack(char const*) - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(_ZN8CrashLog9dumpStackEPKc+0xb2) [0xa47a30]
#3: /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr() [0xa475fc]
#4: /lib64/libc.so.6(+0x3da60) [0x7f89b0a2aa60]
#5: InventoryMenu::keyUpEvent(Tempest::KeyEvent&) - /home/black_fox/src/OpenGothic/build/opengothic/Gothic2Notr(_ZN13InventoryMenu10keyUpEventERN7Tempest8KeyEventE+0x39) [0xa3b145]
(...)
```

Game/utils/crashlog.cpp
```diff
@@ -92,25 +97,78 @@ void CrashLog::dumpStack(const char *sig) {
...
   fout << "---crashlog(" <<  sig << ")---" << std::endl;
   writeSysInfo(fout);
+#ifdef __WINDOWS__
   traceback.log(db, fout);
+#elif __LINUX__
+  tracebackLinux(fout);
 #endif
   }
 
+std::string CrashLog::demangle(const void* frame, const char* symbol) {
```
Try, line 126: Using std::string in crash-log is risky: string do heap-allocation, while heap can be already corrupted. 
May be just emplace this code into `void CrashLog::tracebackLinux`

CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ target_link_libraries(${PROJECT_NAME} GothicShaders)
 
 if(NOT MSVC)
   target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wconversion -Wno-strict-aliasing -Werror)
+  # -rdynamic is needed to allow obtaining meaningful backtraces
+  target_link_options(${PROJECT_NAME} PRIVATE -rdynamic)
```
Try, line 107: `if(UNIX)` ?
-rdynamic seems not working in mingw

Try (changes requested):

Game/utils/crashlog.cpp
Nindaleth: I agree, the reworked approach should prevent these allocations.

CMakeLists.txt
Nindaleth: It's good there's a CI attached :-) Added UNIX check, should be much more specific.

Nindaleth:
I've removed the [WIP] - the review comments have been hopefully resolved, please take another look.

Try:
Looks good to me - merged.

PR #95, opened 2020-10-22
close inventory also with backspace

Nindaleth:
Currently, the inventory can be opened with Tab and Backspace, but can be closed only with Tab. This adds support for Backspace too.

### DIFF
--- a/Game/ui/inventorymenu.cpp
+++ b/Game/ui/inventorymenu.cpp
@@ -315,7 +315,7 @@ void InventoryMenu::keyRepeatEvent(KeyEvent& e) {
 void InventoryMenu::keyUpEvent(KeyEvent &e) {
   takeTimer.stop();
   lootMode = LootMode::Normal;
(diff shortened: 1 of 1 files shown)


Try:
This patch doesn't look correct to me:
In game inventory-button is not hard-coded and can be changed by user.

For this case `keycodec.tr(e)==KeyCodec::Inventory` should be used.

Nindaleth:
Sorry for no activity, I'm in a process of moving. I'll get back to the PRs once the dust settles.

Nindaleth:
I think I understand the translation a bit now - the INI file has between 0 and 2 keycodes stored for one action - in my case Tab and Backspace for Inventory window - those values I could check manually using one of the `KeyCodec::keys` init lists. Then the `.is()` in keycodec.h is used to check whether the pressed keycode matches any of the up to 2 configured ones and if so, the appropriate action is returned. 

After I realized that the INI file is taken from the path given in -g parameter, not as a part of OpenGothic, it only took a bit of time to find the `.is()` function to confirm how the process works.

I've updated and rebased the change.

Try:
Cool! Not it's one-line patch :)

PR #96, opened 2020-11-08
 Remove BYPRODUCTS from cmake custom commands

ChristophHaag:
Some fixes I had to add to build on Linux, and with ninja (`cmake -G Ninja`)


Try:
Merged; engine submodule already updated in separate commit

PR #108, opened 2020-12-19
inventory: Crash Bugfix

DHertrich:
Fixed inventory keyDownEvent to allow for equipping
and using items with ctrl key. Would otherwise may
result in segfault crash because calling loot logic.

### DIFF
--- a/Game/ui/inventorymenu.cpp
+++ b/Game/ui/inventorymenu.cpp
@@ -294,12 +294,26 @@ void InventoryMenu::keyDownEvent(KeyEvent &e) {
     lootMode = LootMode::Stack;
     takeTimer.start(200);
     onTakeStuff();
-    }
-  else if(keycodec.tr(e)==KeyCodec::ActionGeneric) {
-    lootMode = LootMode::Normal;
-    takeTimer.start(200);
-    onTakeStuff();
-    }
+    } else if (keycodec.tr(e) == KeyCodec::ActionGeneric) {
+
+      auto &p = activePage();
+      auto &sel = activePageSel();
+      if (sel.sel >= p.size()) {
+          return;
+      }
+      auto &item = p[sel.sel];
+      if (state == State::Equip) {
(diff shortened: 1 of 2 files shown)


Try:
Hi, @DHertrich !

Can you clarify what was the root cause of crash? In case if `SIGSEGV` there must be null pointer or bad access somewhere, right?

DHertrich:
Unfortunately it crashes at a more or less random time afterwards (we're talkink some function calls later), so I think it causes a crash on a different thread. I suspect it has something to do with dropItem but I am not sure. Could you try if you can reproduce the crash by pressing left ctrl in the inventory yourself?

Try:
> Could you try if you can reproduce the crash by pressing left ctrl in the inventory yourself?

I've tried many times already, with different compilers - no luck reproducing. 
Since we don't know the root cause, there is a risk to only hide real issue be this patch.

May be this is physic engine issue and not inventory? Can you reproduce problem in case if `Item::setPhysicsEnable` disabled?

DHertrich:
> May be this is physic engine issue and not inventory? Can you reproduce problem in case if Item::setPhysicsEnable disabled?

Yes, correct. Commenting out line 2419 in npc.cpp
`it->setPhysicsEnable(*owner.physic());`
will make the game stop crashing.

Still weird that you can't reproduce. Here is my toolchain configuration:

Screenshot 2020-12-22 170031

Switching the architecture from x86 to amd64 did not change the behaviour.

Interestingly, my game crashes with a SEGSEV too, when approaching cavalorn who fights the goblins. This might be related, as it seemingly crashes the moment he kills a mob, which should make the goblin drop its weapon right?

Try:
> This might be related, as it seemingly crashes the moment he kills a mob, which should make the goblin drop its weapon right?

Yes, look like the real problem is a object-drop and dynamic-physic simulation

Try:
Closing, because weapon-drop was fixed separatly

PR #110, opened 2020-12-26
feat: inventory original buttons actions

okkindel:
Originally, in the game, the action key was used to "use" an item, and the jump key was used to throw it away. Thought it might be worth recreating.


Try:
Thanks for PR, merged!

Issue #97, opened 2020-11-08
Key interpretation


hama3254:
I noticed that the game does use the left and right rotate keys not as the original.

You can not use the rotate keys `RotateL` and `RotateR` (from `keyLeft` and `keyRight`) to navigate in the inventor and they are also do not work for fighting.
Both works fine with the `Left` and `Right` (from `keyStrafeLeft` and `keyStrafeRight`) 

Should it be that way?

Try:
Hi, @hama3254 !

> Should it be that way?

As far as I know vanilla gothic2 behaves exactly same way, isn't it?

Kirides:
> As far as I know vanilla gothic2 behaves exactly same way, isn't it?

yes. it does.
By default in G2 the arrow keys map to up, down, rotate left, rotate right and do not allow you to move cells left/right if pressing left or right arrow key.

hama3254:
> > As far as I know vanilla gothic2 behaves exactly same way, isn't it?
> 
> yes. it does.
> By default in G2 the arrow keys map to up, down, rotate left, rotate right and do not allow you to move cells left/right if pressing left or right arrow key.

For me this was never the case, if i take a look at my German Steam Version from Notr this is how it acts : [YouTube](https://www.youtube.com/watch?v=p5UAdOXaGDk)
i used A and D to turn for ever since mouse support in Gothic was not great and i can move the inventory with that exact keys.

okkindel:
I think, we can close now this issue, fixed in 6be4ca17f9e6b3bd89e02ccb1a953e3e311370c9.

PR #111, opened 2020-12-27
feat: original buttons

okkindel:
I think I looked at all the gothic controls and its implementation in the engine. I corrected all the changes I noticed, to match the original. A list of changes:

- in the inventory you can use rotating buttons to go left / right in the items table
- lockpicking can by provided with rotating buttons
- swinging the sword left and right should be possible after holding down the action button and rotating buttons
- the Y rotation of the player during the dive is possible using the forward and backward buttons
- closing documents (stone plates, letters) should be possible with the inventory button, otherwise you can close the inventory with the document remaining on the screen
- closing the trade screen with the inventory key


okkindel:
Also resolves https://github.com/Try/OpenGothic/issues/97

Try:
@okkindel , thanks for PR!
Merged.

PR #112, opened 2020-12-31
fix 'class-memaccess' error

christianrauch:
fixes the compiler error:
```
[...]/OpenGothic/Game/graphics/skeletalstorage.cpp:113:16: error: \u2018void* memset(void*, int, size_t)\u2019 clearing an object of non-trivial type \u2018struct SkeletalStorage::TImpl<48>::Block\u2019; use assignment or value-initialization instead [-Werror=class-memaccess]
  113 |     std::memset(m,0,sizeof(Matrix4x4)*bonesCount);
      |     ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Game/graphics/skeletalstorage.cpp
+++ b/Game/graphics/skeletalstorage.cpp
@@ -110,7 +110,9 @@ struct SkeletalStorage::TImpl : Impl {
   void   free(const size_t objId, const size_t bonesCount) override {
     freeList.free(objId, bonesCount);
     auto m = &obj[objId];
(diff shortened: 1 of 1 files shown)


Try:
Hi @christianrauch and thanks for PR!

`-Wclass-memaccess` is awkward warning, in this case it's more like a compiler bug, because block is trivially-copyable class.
I'll prefer to fix this by adding `static_cast<void*>` to `m` - I'll upload this fix as separated commit

PR #131, opened 2021-03-21
Marvin Mode: Insert command and rework

Eulenmensch93:
Implement the gothic 2 console command "insert" with NPC or item name as parameter.

Rework of the marvin class to allow commands with dynamic parameters.

### DIFF
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -6,132 +6,156 @@
 #include "world/objects/npc.h"
 #include "camera.h"
 #include "gothic.h"
+#include "utils/stringutil.h"
 
-Marvin::Marvin() {
-  cmd = std::vector<Cmd>{
-    {"cheat full",        C_CheatFull},
-
-    {"camera autoswitch", C_CamAutoswitch},
-    {"camera mode",       C_CamMode},
-    {"toogle camdebug",   C_ToogleCamDebug},
-    {"toogle camera",     C_ToogleCamera},
-    };
+Marvin::Marvin(Gothic& gothic)
+: gothic( gothic )
+, commandHandlers()
+{
+  commandHandlers["cheat full"] = CommandFunction(&Marvin::handleCheatFull);
+  commandHandlers["camera autoswitch"] = CommandFunction(&Marvin::handleCameraAutoSwitch);
+  commandHandlers["camera mode"] = CommandFunction(&Marvin::handleCameraMode);
+  commandHandlers["toggle camdebug"] = CommandFunction(&Marvin::handleToggleCamDebug);
+  commandHandlers["toggle camera"] = CommandFunction(&Marvin::handleToggleCamera);
+  commandHandlers["insert"] = CommandFunction(&Marvin::handleInsert);
   }
 
-Marvin::CmdVal Marvin::recognize(const std::string& v) {
-  char clean[256] = {};
+void Marvin::autoComplete(std::string& inputString) {
+  inputString = StringUtil::trimFront(inputString, ' ');
+  CommandMap::iterator it;
 
-  for(size_t i=0, r=0; r<255 && i<v.size(); ++i) {
-    char c = v[i];
-    if('A'<=c && c<='Z')
-      c = char(c+'a'-'A');
-    if(r==0 && c==' ')
-      continue;
-    if(c==' ') {
-      while(v[i+1]==' ')
-        ++i;
+  size_t matchCount = 0;
+  std::string matchCommand;
+
+  for(it = commandHandlers.begin(); it != commandHandlers.end(); it++) {
+    const std::string command(it->first);
+
+    if(inputString.length() < command.length()) {
+      std::string restOfCommand( command.substr(inputString.length()));
+      std::string::size_type spacePos = restOfCommand.find(" ");
+      std::string commandToken(command);
+
+      if( spacePos != std::string::npos ) {
+        commandToken = command.substr(0, inputString.length() + spacePos);
       }
 
-    clean[r] = c;
-    ++r;
-    }
 
-  size_t cmdLen = std::strlen(clean);
-  size_t prefix = cmdLen;
-  if(prefix==0)
-    return C_None;
-
-  while(prefix>0) {
-    if(clean[prefix]==' ')
-      break;
-    --prefix;
-    }
-
-  size_t  cnt           = 0;
-  size_t  suggestionLen = 0;
-  Cmd     suggestion;
-
-  for(auto& i:cmd) {
-    const size_t len = std::strlen(i.cmd);
-    if(len<prefix)
-      continue;
-    if(std::memcmp(clean,i.cmd,prefix)!=0)
-      continue;
-
-    if(cmdLen<=len && std::memcmp(clean,i.cmd,cmdLen)==0) {
-      size_t len = cmdLen;
-      while(i.cmd[len]!='\0' && i.cmd[len]!=' ')
-        ++len;
-      if(len!=suggestionLen || std::memcmp(suggestion.cmd,i.cmd,len)!=0) {
-        suggestion    = i;
-        suggestionLen = len;
-        cnt++;
+      if(commandToken.substr(0,inputString.length()) == inputString) {
+        if(matchCommand != commandToken) {
+          matchCommand = commandToken;
+          matchCount++;
         }
-      }
-    }
-
-  if(cnt==0)
-    return C_Invalid;
-
-  if(cnt!=1)
-    return C_Incomplete;
-
-  if(suggestion.cmd[cmdLen]=='\0')
-    return CmdVal{suggestion,cmdLen};
-
-  suggestion.type = C_Incomplete;
-  for(size_t i=cmdLen;; ++i) {
-    if(suggestion.cmd[i]=='\0')
-      return CmdVal{suggestion,cmdLen};
-    if(suggestion.cmd[i]==' ' || i>=suggestionLen)
-      return CmdVal{suggestion,cmdLen};
-    }
-  }
-
-void Marvin::autoComplete(std::string& v) {
-  auto ret = recognize(v);
-  if(ret.cmd.type==C_Incomplete&& ret.cmd.cmd!=nullptr) {
-    for(size_t i=ret.cmdOffset;; ++i) {
-      if(ret.cmd.cmd[i]=='\0')
-        return;
-      if(ret.cmd.cmd[i]==' ') {
-        v.push_back(' ');
-        return;
-        }
-      v.push_back(ret.cmd.cmd[i]);
       }
     }
   }
 
-bool Marvin::exec(Gothic& gothic, const std::string& v) {
-  auto ret = recognize(v);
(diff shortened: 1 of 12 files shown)


Game/world/objects/npc.cpp
```diff
@@ -342,6 +342,11 @@ bool Npc::setPosition(const Tempest::Vec3& pos) {
   return setPosition(pos.x,pos.y,pos.z);
   }
 
+Tempest::Vec3 Npc::getPosition() const
```
Try, line 345: Duplicate of `Npc::position()`

Game/world/world.cpp
```diff
@@ -791,6 +795,20 @@ void World::addSound(const ZenLoad::zCVobData& vob) {
     }
   }
 
+bool World::addItemOrNpcBySymbolName (const std::string& name, const Tempest::Vec3& at) {
```
Try, line 798: This function is used only in single case (in marvin) - should be inlined, to keep World-class cleaner

Game/world/world.cpp
```diff
@@ -791,6 +795,20 @@ void World::addSound(const ZenLoad::zCVobData& vob) {
...
+    size_t id = script().getSymbolIndex(name);
+
+    if( addNpc(id, at) != nullptr ) {
```
Try, line 801: Type of symbol can be deducted by `script().getSymbol(id).instance.instanceOf()`

Game/marvin.cpp
```diff
@@ -6,132 +6,130 @@
 #include "world/objects/npc.h"
 #include "camera.h"
 #include "gothic.h"
+#include "utils/stringutil.h"
+
+Marvin::Marvin(Gothic& gothic)
+: gothic( gothic )
+, commandHandlers()
```
Try, line 13: Nitpick: useless line

Game/marvin.cpp
```diff
@@ -6,132 +6,130 @@
...
+  commandHandlers["camera mode"] = CommandFunction(&Marvin::handleCameraMode);
+  commandHandlers["toggle camdebug"] = CommandFunction(&Marvin::handleToggleCamDebug);
+  commandHandlers["toggle camera"] = CommandFunction(&Marvin::handleToggleCamera);
+  commandHandlers["insert"] = CommandFunction(&Marvin::handleInsert);
+  }
 
-Marvin::Marvin() {
-  cmd = std::vector<Cmd>{
-    {"cheat full",        C_CheatFull},
+void Marvin::autoComplete(std::string& inputString) {
```
Try, line 23: This code do very-much same as `Marvin::recognize` use to preform, but in veery expensive manner.
OK, this path is not time-critical, but why downgrade algorithm?

Game/marvin.cpp
```diff
@@ -6,132 +6,130 @@
...
-  if(cnt!=1)
-    return C_Incomplete;
+bool Marvin::exec(const std::string& inputString) {
+  bool ret = false;
+  size_t matchCount = 0;
+  std::string matchCommand("");
 
-  if(suggestion.cmd[cmdLen]=='\0')
-    return CmdVal{suggestion,cmdLen};
+  std::string trimedInputString = StringUtil::trimFront(inputString, ' ');
```
Try, line 62: The command recognition seems to be duplicated across 2 functions: `Marvin::exec` and `Marvin::autoComplete`

Game/ui/consolewidget.h
```diff
@@ -27,11 +27,11 @@ class ConsoleWidget : public Tempest::Widget {
 
     Gothic&                   gothic;
 
-    Tempest::UiOverlay*       overlay    = nullptr;
-    const Tempest::Texture2d* background = nullptr;
+    Tempest::UiOverlay*       overlay;
```
Try, line 30: Why remove inline initialization?

Game/utils/stringutil.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <locale>
+
+
+std::string StringUtil::toLower(const std::string& str) {
+  std::string ret(str);
+  
+  for(std::string::size_type i = 0; i < ret.length(); i++)
+  {
+    ret[i] = std::tolower(ret[i], std::locale::classic());
```
Try, line 11: Behavior of  `tolower` is undefined for non ancii-z input. Meanwhile, `str` is utf8 string.

Game/utils/stringutil.cpp
```diff
@@ -0,0 +1,72 @@
...
+  
+  return ret;
+  }
+
+std::string StringUtil::toUpper(const std::string& str) {
+  std::string ret(str);
+  
+  for(std::string::size_type i = 0; i < ret.length(); i++)
+  {
+    ret[i] = std::toupper(ret[i], std::locale::classic());
```
Try, line 22: Same as in line 11

Game/world/worldobjects.cpp
```diff
@@ -523,22 +523,41 @@ Bullet& WorldObjects::shootBullet(const Item& itmId,
...
+  Item* item = nullptr;
+  Tempest::Vec3 pos;
+  Tempest::Vec3 dir;
+  const WayPoint* waypoint = owner.findPoint(at);
+
+  if(waypoint != nullptr) {
+    pos = {waypoint->x, waypoint->y, waypoint->z};
+    dir = {waypoint->dirX, waypoint->dirY, waypoint->dirZ};
+  } else {
+    pos = {0,0,0};
```
Try, line 535: Vec3 is initialized as {0,0,0} by default, no need for else-path

Game/utils/stringutil.cpp
```diff
@@ -0,0 +1,72 @@
...
+  }
+  
+  return str.substr(0,i);
+  }
+
+std::string StringUtil::trim(const std::string& str, char c) {
+  std::string ret;
+  
+  ret = trimFront(str, c);
+  ret = trimBack(str, c);
```
Try, line 56: Previous value of `ret` was never used. `ret = trimBack(ret, c);` ?

Game/world/objects/npc.cpp
Eulenmensch93: Good point, probably missed it because Npc::position() has auto as return type in the .h file.
Will remove the commit introducing the new function.

Game/world/world.cpp
Eulenmensch93: OK will do that, I am not really a friend of inline functions, but I can understand the argument :)

Game/world/world.cpp
Eulenmensch93: Nice, I will change that.

Game/marvin.cpp
Eulenmensch93: You mean the commandHandlers init?
I am not sure if it is strictly necessary, but I once had the problem that I did not init a map in the constructor which lead to a crash while inserting elements.
Also I like to see in the constructor what the initial state of all the member variables are, this also avoids the constructor variable ordering warning.

Game/marvin.cpp
Eulenmensch93: Truth be told, I did not really understand what the recognize method really does.
And because I was already changing the the way the data is stored I just replaced the code.

This was also not the first version I implemented that day, the first implementation was much more clean but did not have the same result as your original implementation.

I will look into the recognize function once more and try to understand exactly whats happening and maybe I can re-use it.
 I was not sure if it can handle commands with parameters like the "insert" command?
Where only the command can be recognized, but not the rest of the string.

Game/marvin.cpp
Eulenmensch93: Possibly, will need to look into this as part of the discussion about the recognize function.

Game/ui/consolewidget.h
Eulenmensch93: Just because I am not a friend of inline stuff in the .h files, the .h file in my opinion only describes the interface.
And all the other variables are already initialized in the constructor of the object, why not those two?

Game/utils/stringutil.cpp
Eulenmensch93: std::string stores the data as char, so on most platforms it holds 8bit, but as I understand it will still work with utf8, because of the way the data is encoded.

But you are probably right that the std::tolower will not do what we want it to do.
Did the first few lines of the recognice function do a to upper or lower by adding and subtracting 'A' and 'a'?
And this not also be utf8 incompatible?

Game/world/worldobjects.cpp
Eulenmensch93: OK, good to know, will be removed.

Game/utils/stringutil.cpp
Eulenmensch93: Copy paste error, sorry.

Eulenmensch93:
I would like to create a new branch to fix all the discussed points, because I can just delete one commit using git rebase.
Is this OK or is this a bad idea for some reason on github? (Sorry, I am new to github)
I want to avoid to clutter your commit history with useless commits which are reverted before they get pulled and stuff like this.

Game/world/objects/npc.cpp
Eulenmensch93: One more question, why use the -> syntax in the .h file with auto return type?
Wouldn't it be easier to just write `Tempest::Vec3 position() const;`?
Or is it possible that other types are returned here?

And sorry if this is a stupid question, Its the first time that I see that syntax.

Try:
> Is this OK or is this a bad idea for some reason on github? 

Yes, it's actually a good call :)

Game/world/world.cpp
Eulenmensch93: I was thinking about why I dislike inline functions and variable initialization.
The reason is that I like to open the .h file to have a look at the interface of a class or object, and in most cases I do not care about the implementation details at that moment. So inline code will only make the interface harder to read in my opinion.
Also I think it will trigger a rebuild of all the files that include that interface if the inline function implementation is changed which would otherwise not be necessary.

But I am aware that this is probably my own personal preference and style so I have no problem with changing it if you pref ere the things to be inline :)

pkurras: While you might be right about the interface in VS there is CTRL+SHIFT+M for collapsing the Code and it looks clean again. Inlining functions - if the compiler decides to follow the request to inline - has the positive effect of removing the function prologue and epilogue from the assembly improving it for being integrated in performance relevant loops and for functions that are tiny and where the stack frame would nearly double the binary size of the function without inlining them. It adds to the size of the calling function however. So whether you like them is out of the question. It's a valid tool for a specific problem just like anything in C++, but in most cases it doesn't change that much for regular code. Good examples for inlining are math library member functions like vector math and stuff like that. They are called incredibly often in games so it does make a difference.

Eulenmensch93: Thanks for the detailed explanation, I only knew that it sometimes can improves compiler optimization.

Game/world/world.cpp
Eulenmensch93: Does not work as "insert ice_dragon" will return instance Daedalus::IC_None.
Is this intended behavior?

PR #132, opened 2021-03-23
Marvin Mode insert command support

Eulenmensch93:
- Add new addItem methods.
- Implement addItemOrNpcBySymbolName as inline method to be able to spawn items and npcs by name in marvin mode insert command.
- Adjust marvin mode recognize method to support commands with dynamic parameters.
- Implement marvin mode insert command.

### DIFF
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -147,9 +147,9 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
 
         if( world == nullptr || player == nullptr ) {
           return false;
-        } else{
-          return addItemOrNpcBySymbolName(world, arguments, player->position());
         }
+
+        return addItemOrNpcBySymbolName(world, arguments, player->position());
       }
 
       return false;
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -137,7 +137,6 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
       return true;
       }
     case C_Insert: {
-      bool ret = false;
       World* world = gothic.world();
       Npc* player = gothic.player();
 
@@ -147,13 +146,13 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
         std::string arguments = v.substr(spacePos + 1);
 
         if( world == nullptr || player == nullptr ) {
-          ret = false;
+          return false;
         } else{
-          ret = addItemOrNpcBySymbolName(world, arguments, player->position());
+          return addItemOrNpcBySymbolName(world, arguments, player->position());
         }
       }
 
-      return ret;
+      return false;
       }
     }
 
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -15,6 +15,7 @@ Marvin::Marvin() {
     {"camera mode",       C_CamMode},
     {"toogle camdebug",   C_ToogleCamDebug},
     {"toogle camera",     C_ToogleCamera},
+    {"insert",            C_Insert},
     };
   }
 
@@ -134,6 +135,25 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
       if(auto c = gothic.camera())
         c->setToogleEnable(!c->isToogleEnabled());
       return true;
+      }
+    case C_Insert: {
+      bool ret = false;
+      World* world = gothic.world();
+      Npc* player = gothic.player();
+
+      std::string::size_type spacePos = v.find(" ");
+
+      if( spacePos != std::string::npos ) {
+        std::string arguments = v.substr(spacePos + 1);
+
+        if( world == nullptr || player == nullptr ) {
+          ret = false;
+        } else{
+          ret = world->addItemOrNpcBySymbolName(arguments, player->position());
+        }
+      }
+
+      return ret;
       }
     }
 
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -67,6 +67,11 @@ Marvin::CmdVal Marvin::recognize(const std::string& v) {
         suggestionLen = len;
         cnt++;
         }
+      } else if(prefix == len) {
+        cmdLen        = prefix;
+        suggestion    = i;
+        suggestionLen = 0;
+        cnt++;
       }
     }
 
--- a/Game/marvin.cpp
+++ b/Game/marvin.cpp
@@ -149,7 +149,7 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
         if( world == nullptr || player == nullptr ) {
           ret = false;
         } else{
-          ret = world->addItemOrNpcBySymbolName(arguments, player->position());
+          ret = addItemOrNpcBySymbolName(world, arguments, player->position());
         }
       }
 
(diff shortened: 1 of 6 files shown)


Eulenmensch93:
Please pay extra attention on the commit https://github.com/Try/OpenGothic/pull/132/commits/dfe4efbf5303808f90329e28c758f3317295c7f4, I think this should work but its hard to see if I introduce side effects here.

Discussed points should now be fixed.

Eulenmensch93:
By the way, while testing I noticed that when I try to spawn a ice_dragon, a empty health bar would appear but no dragon.
I pulled your current master branch before creating this branch, either I broke something while create my new branch, or something happend in your master branch.

Game/world/world.h
```diff
@@ -175,6 +176,19 @@ class World final {
     void                 addFreePoint  (const Tempest::Vec3& pos, const Tempest::Vec3& dir, const char* name);
     void                 addSound      (const ZenLoad::zCVobData& vob);
 
+    inline bool addItemOrNpcBySymbolName (const std::string& name, const Tempest::Vec3& at) {
```
Try, line 179: That is not what I meant by saying "inline" :)
My point is: there is not reason for this method to exist in `World` class - this code can be just moved into `marvin.cpp` under C_Insert case

Game/marvin.h
```diff
@@ -25,6 +25,8 @@ class Marvin {
       C_CamMode,
       C_ToogleCamDebug,
       C_ToogleCamera,
+      // Insert
```
Try, line 28: Nitpick: this comment doesn't give any new info - can be removed

Game/marvin.cpp
```diff
@@ -130,6 +136,25 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
         c->setToogleEnable(!c->isToogleEnabled());
       return true;
       }
+    case C_Insert: {
+      bool ret = false;
```
Try, line 140: `ret` can be removed: if world is null or something - you can just early-out by `return false` straight

Try (changes requested):
Apart from `addItemOrNpcBySymbolName` looks quite good.
There is no auto-complete for item/npc names, this is how it was intended?

Game/world/world.h
```diff
@@ -175,6 +176,19 @@ class World final {
...
+      bool ret = false;
+      size_t id = script().getSymbolIndex(name);
+
+      if( addNpc(id, at) != nullptr ) {
```
Try, line 183:
```
if(script().getSymbol(id).instance.instanceOf(Daedalus::IC_Npc))
  addNpc(..)
if(script().getSymbol(id).instance.instanceOf(Daedalus::IC_Item))
  addItem(..)
```

Game/marvin.cpp
Eulenmensch93: Its a habit from work where we have to be MISRA complient to a certain extend.
And one MISRA rule is that early exit is not allowed, which has its benefits for example you dont need to set breakpoints every where and you are certain that the code  before the last return is executed every time if you have cleanup code at the end of the function for example.

But no problem will change that.

Game/world/world.h
Eulenmensch93: This does not work, I tried that. Mosters seem to be IC_None for some reason.

Game/marvin.h
Eulenmensch93: I thought you are using those comments to separate groups of commands, like camera specific and other commands.
If I just delete the line it will look like the insert command is part of the camera commands.
So I will just remove the comment and leave a empty line for visual distinction.

Game/world/world.h
Eulenmensch93: Ahhhhh, that makes muuuuuuuch more sense.
OK will do that, sorry for the misunderstanding.

Eulenmensch93:
> Apart from `addItemOrNpcBySymbolName` looks quite good.
> There is no auto-complete for item/npc names, this is how it was intended?

Yes that is intended for now as I don't have any Idea how to incorporate that into the recognize algorithm.
I am not even sure that the original game supports that?

Game/world/world.h
Try: Right, `instanceOf` work only on already created instances.
Anyway I think, that try-and-fail is not a great approach to instantiate object by name. 

I take a quick look on Daedalus to see what is possible to do here:
```
  auto& obj = vm.getDATFile().getSymbolByName(name); // get's you a symbol props
  // obj.properties.elemProps.type = EParType_Instance, for instances
  // obj.properties.elemProps.type = EParType_Class, for classes
  auto& cls = vm.getDATFile().getSymbolByIndex(s.parent); // get's you a parent object or prototype
  cls.name=="C_NPC" // for npc
```
So my suggestion is to unroll s.parent chain until parent=size_t(-1), or C_NPC/C_ITEM is found.

Game/marvin.cpp
```diff
@@ -130,8 +136,38 @@ bool Marvin::exec(Gothic& gothic, const std::string& v) {
...
+      Npc* player = gothic.player();
+
+      std::string::size_type spacePos = v.find(" ");
+
+      if( spacePos != std::string::npos ) {
+        std::string arguments = v.substr(spacePos + 1);
+
+        if( world == nullptr || player == nullptr ) {
+          return false;
+        } else{
```
Try, line 150: no need to `else`, after return.

Try (commented):
> Yes that is intended [ no auto-complete for item/npc names ]

OK, I'll do it after pr will be merged; and yes - this was in original game

Game/marvin.cpp
AdamBilisic: I think you have read that incorrectly. return is in if statement. If the condition fails it goes to else without ever hitting return. I think code is correct.

Try: @AdamBilisic, yes this code is correct, but `else` operator still doesn't affect anything:
if( {true} ) - early-out, by return
if( {false} ) - just continue forward, without else

AdamBilisic: Ah yeah so you meant redundant else because if already returns value out of the function. Yes this is personal preference but I like it too. So desired outcome is

 if( world == nullptr || player == nullptr ) {
    return false;
 }
 return addItemOrNpcBySymbolName(world, arguments, player->position());

Eulenmensch93: Sorry, had a lot todo the last few days.
Should be fixed now.

Game/world/world.h
Eulenmensch93: Yea I also don't like the current implementation, its more a workaround than a implementation.
I will take a look at this again when I have some more time.

I started to look into the code when I started to create this pull request, and for some issues, so I currently do not have much of a understanding of the code especially about the properties which could be used and how everything works together.

Could take some time till I get to it, but thanks for looking into it.

Try:
Hi, @Eulenmensch93 !
Kindly asking how it's going: on my side PR marked as "request changes", just want to make sure, that you not waiting on me.
Also if you shot on time, we can merge patch as-is and I'll do requested changes by my self - that also fine.

Try:
since the was no response, decision is:
merge as-is and fix object creation later

Eulenmensch93:
> since the was no response, decision is:
> merge as-is and fix object creation later

Sorry for not responding, I was not waiting for you, but I did not get to have a look into it.
I am happy that you merged the changes thank you very much.
Hopefully I can get back to the project at some point.

Thank you for your time and this learning experience.

PR #135, opened 2021-04-14
UI improvements

kuzy000:
* Dialogues
  - [x] RMB to skip phrase
  - [x] Esc down (instead of up) to skip phrase
  - [x] Skip fade out animation when phrase is skipped
* Inventory
  - [x] RMB to close
  - [x] Esc down (instead of up) to close inventory (or trade)
  - [x] Move selection left when there is no space to move up
  - [x] Move selection right when there is no space to move down
  - [x] Maintain selection when switching pages
  - [x] Show selection even if inventory is empty

I checked all the changes in the original game.

### DIFF
--- a/Game/ui/dialogmenu.cpp
+++ b/Game/ui/dialogmenu.cpp
@@ -345,6 +345,13 @@ void DialogMenu::startTrade() {
   dlgTrade=false;
   }
 
+void DialogMenu::skipPhrase() {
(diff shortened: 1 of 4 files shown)


Try:
Hi, @kuzy000 !

Thanks for PR, I've tested it - everything works quite nicely!
Special thanks for very 'clean' pull-request :)

Issue #127, opened 2021-03-05
Flathub package?


Newbytee:
I think that for easy cross-distribution distribution on Linux having this packaged on [Flathub](https://flathub.org/home) would be nice. This could also potentially give this some extra publicity.

Try:
Hi, @Newbytee !

I'm not sure about this Flathub-thing, since I'm not a Linux user. Mark this ticket as help-need to collect opinions.

Generally, first-thing first: Probably *.deb package and binary-releases goes first...

Newbytee:
> I'm not sure about this Flathub-thing, since I'm not a Linux user. Mark this ticket as help-need to collect opinions.

I might take this on myself. My main concern is really that as previously mentioned Flatpak packages tend to be focused on GUI rather than CLI, and as such users might be confused if they install OpenGothic from Flathub and don't see any .desktop entry in their menus. One workaround for this could be to e.g. write a shell script that launches in a terminal and asks the user for a path to their installation and then launches OpenGothic.

> Generally, first-thing first: Probably *.deb package and binary-releases goes first...

Understandable, but I also think this might be better in terms of reach. Flatpak packages (which Flathub is a repository for) work on any Linux distribution (provided they provide support for Flatpak themselves, which all big ones do)

Newbytee:
If I make a Flatpak package, what would you want the app ID to be? Usually you'd do the domain name in reverse followed by the name of the program, so in this case it could be com.github.OpenGothic. As you can see there's quite a lot of packages that use com.github as "domain": https://github.com/flathub?q=com.github&type=&language=&sort=

Either way, I thought I'd ask you before I actually go through with that name. Do you have a different domain I should use, or is that fine?

Try:
>  com.github.OpenGothic

Yes, this name is OK to me

DragonSWDev:
@Newbytee >com.github.OpenGothic

Examples you posted also include GitHub user name in ID so shouldn't it be something like com.github.Try.OpenGothic?

Newbytee:
> @Newbytee
> 
> > com.github.OpenGothic
> 
> Examples you posted also include GitHub user name in ID so shouldn't it be something like com.github.Try.OpenGothic?

Yeah, I realised that after writing but didn't update my comment since Try already had replied. I agree that this would be better as otherwise it sounds like it's made by GitHub themselves.

DragonSWDev:
Good to know. It would be nice to get OpenGothic available on Flathub.

[...]

PR #143, opened 2021-05-04
Add Appstream XML

Newbytee:
This is the first part to implementing https://github.com/Try/OpenGothic/issues/127. Appstream XML is something used for Linux store frontends to display information/metadata about packages and is, believe it or not, quite universally supported. While I'm adding this for the purposes of publishing to Flathub (which mandates that every package has Appstream XML), this is something e.g. a .deb or .rpm package could make use of as well. You can read more about the format here: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

Unless I run into something while implementing the actual Flathub "build script", this will probably be all the necessary work in this repository to get OpenGothic published on Flathub. The "build script" (it's not really a script) will be housed in a repository in Flathub's organisation, like all the other packages there: @flathub. There are other things that would be nice to have for a Flathub package, like a GUI launcher, but they are out of scope for this pull request and would require significantly more work and collaboration.

Also, this validates fine:

    [neboula@shock OpenGothic]$ flatpak run org.freedesktop.appstream-glib validate packaging/opengothic.appdata.xml
    packaging/opengothic.appdata.xml: OK

### DIFF
--- a/packaging/opengothic.appdata.xml
+++ b/packaging/opengothic.appdata.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>io.github.Try.OpenGothic</id>
+  <name>OpenGothic</name>
+  <summary>Open source reimplementation of Gothic 2 Notr</summary>
+  <description>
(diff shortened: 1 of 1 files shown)


Try:
looks good to me - merged

PR #147, opened 2021-05-07
Allow specifying Gothic catalog interactively

Newbytee:
This is useful for environments where it's not feasible to expect users to invoke the game from a command line in such a way that they can specify the parameters they want, e.g. Flatpaks from Flathub.

I can't get latest master to build at all so I haven't tested this there, but it does build against v0.37. Also worth noting is that I've only tested this on Linux, and there are some encoding conversions that may be problematic on Windows?

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Don't expect DirectX11 mod to work, since technicaly it's not a mod. But Project
 
 ##### Command line arguments
 * -g specify gothic game catalog
+* -i specify gothic game catelog interactively
 * -nomenu - skip main menu
 * -nofrate - disable FPS display in-game
 * -w <worldname.zen> - startup world; newworld.zen is default
--- a/game/gothic.cpp
+++ b/game/gothic.cpp
@@ -6,6 +6,8 @@
 #include <zenload/zCMesh.h>
 #include <cstring>
 #include <cctype>
+#include <codecvt>
+#include <locale>
 
 #include "game/definitions/visualfxdefinitions.h"
 #include "game/definitions/sounddefinitions.h"
@@ -31,6 +33,14 @@ Gothic::Gothic(const int argc, const char **argv) {
       ++i;
       if(i<argc)
         gpath.assign(argv[i],argv[i]+std::strlen(argv[i]));
+      }
+    else if(std::strcmp(argv[i],"-i")==0) {
+      ++i;
+      std::cout << "Enter the path to your Gothic installation:\n";
+      std::wstring_convert<std::codecvt_utf8_utf16<char16_t>,char16_t> convert;
+      std::string gothicPath;
+      std::cin >> gothicPath;
+      gpath.assign(convert.from_bytes(gothicPath));
       }
     else if(std::strcmp(argv[i],"-save")==0){
       ++i;


Try:
Btw, haven't try, but there is a file-dialog in bullet3 examples:
```c++
int X11OpenGLWindow::fileOpenDialog(char* filename, int maxNameLength)
{
	int len = 0;
	FILE* output = popen("zenity --file-selection --file-filter=\"*.urdf\" --file-filter=\"*.sdf\"  --file-filter=\"*.obj\"  --file-filter=\"*.*\"", "r");
	if (output)
	{
		while (fgets(filename, maxNameLength - 1, output) != NULL)
		{
			len = strlen(filename);
			if (len > 0)
			{
				filename[len - 1] = 0;
				printf("file open (length=%d) = %s\n", len, filename);
			}
		}
		pclose(output);
	}
	else
	{
		printf("Error: fileOpenDialog no popen output, perhaps install zenity?\n");
	}
	XCloseDisplay(m_data->m_dpy, m_data->m_win);
	return len;
}
Look pretty doable to me
```

Newbytee:
> I'm not 100%, but codecvt is not mandatory thing for compiler to support. In my practice it works on Windows, but doesn't on android. Anyway - there is a <Tempest/TextCodec> for that.

I see. Thanks.

> Also, it would be better I think, to implement as a FileDialog, or something like.

I agree.

> On Windows/Mac it's pretty straight, but can be tricky on Linux
> Btw, haven't try, but there is a file-dialog in bullet3 examples:

I think it would be better if the file dialogue was done using [xdg-desktop-portal](https://github.com/flatpak/xdg-desktop-portal). This would probably be trickier to implement, but the upside is that it works even in sandboxed situations (like Flatpak where I'm currently just breaking open the sandbox) and that it uses a GTK picker if you're using a GTK environment, Qt picker if you're using a Qt environment, and so on. This would mean that D-Bus becomes a dependency (or optional dependency) on Linux, but I don't think that's a problem? D-Bus is pretty ubiquitous on Linux (it's used for handling all kinds of things — notifications, for example).

I can look into implementing this for Linux, but I don't really have any Windows or macOS systems to test on, so I'd probably leave those no-op so you (or someone else) can implement them.

---

This leaves me with more questions however. Should the file picker appear immediately on launch, or should there be a dialogue box informing the user why it's asked to pick a folder? Should it have more of a form-style look where you pick a folder and then press okay, like some sort of mini-launcher? In the latter case we'd need some sort of toolkit which I understand this does not have right now? What are your thoughts?

Newbytee:
This implementation of xdg-desktop-portal could be useful for getting a file/directory picker on Linux: https://github.com/flatpak/libportal

Newbytee:
> > xdg-desktop-portal.

> Looks like sort-of wrapper to me. Is is truly necessary, to show dialog-box?

It's not strictly necessary, but using it has two main benefits (that I know of):

 1. It works in sandboxed environments that support xdg-desktop-portal, such as Flatpak and Snap. Essentially, how this works is that the game only gets to see the directory the user picked and nothing else from the main file system, so the game cannot do anything it shouldn't. Granted, this isn't super relevant for a game like this as it does not have any network components (to my knowledge?), but I personally still like to only provide applications with the absolute minimum amount of permissions that they need to function properly. Note that at least in the case of Flatpak we could allow the game to always see common installation paths for Gothic in the Flatpak manifest.
 2. It uses the user's preferred type of file picker. As you may know, in Linux there's more than one type of desktop environment, and different ones have different file pickers. Without xdg-desktop-portal you either have to write separate code paths for each different one, or just support one of them. Often applications just default to using the GTK file picker, which works, but isn't ideal for people who use e.g. KDE Plasma that uses Qt instead of GTK. Also, works with both X11 and Wayland (the sample code you sent above is X11-specific, though Wayland does include X11 compatibility).

Try:
Agreed - xdg-desktop-portal can be added for linux build of OpenGothic

Newbytee:
> Agreed - xdg-desktop-portal can be added for linux build of OpenGothic

I might add that what I'm currently waiting for in regards to implementing this is for this to happen: https://github.com/flatpak/libportal/issues/36

Currently libportal uses the GPL-2.0 licence which from what I understanding means that if OpenGothic depends on libportal OpenGothic needs to be relicenced as GPL-2.0 which I don't want and I don't imagine you want that either, so hopefully it will be relicenced soon.

See https://opensource.stackexchange.com/a/1641 — this is about GPL-3.0, but I think the same applies to GPL-2.0.

pm4rcin:
@Newbytee isn't the issue fixed with [this commit](https://github.com/flatpak/libportal/commit/726b4c765928bf734321cb64e6f7ac8cbe6ac1f2).

Newbytee:
> @Newbytee isn't the issue fixed with [this commit](https://github.com/flatpak/libportal/commit/726b4c765928bf734321cb64e6f7ac8cbe6ac1f2).

It is. In fact, I started looking into implementing this via libportal yesterday. It's not ready yet though.

pm4rcin:
@Newbytee any progress on that? I'd not ask but it's been over a year so I wonder if you still remember about that and if you are still interested in doing it?

Newbytee:
> @Newbytee any progress on that? I'd not ask but it's been over a year so I wonder if you still remember about that and if you are still interested in doing it?

Hey again. I tried looking into libportal, but I'm not sure how to integrate it properly into OpenGothic in a good manner since it needs glib's main loop running and stuff like that. My current idea is to instead make a small separate launcher application that prompts the user to pick a directory and then passes that to the OpenGothic executable. This way I don't have to bother with making glib and whatnot an optional dependency of OpenGothic. I should probably close this MR.

Newbytee:
Closing this. See https://github.com/Try/OpenGothic/issues/127#issuecomment-1262610208.

PR #193, opened 2021-09-13
Added 'k' key -> warp forward functionality in marvin

houkama:


houkama:
Oops, lines 336 and 337 in playercontrol.cpp should be removed/commented so the player isn't healed when pressing k

PR #208, opened 2021-11-07
[Fix] Bug fixes for UI

bwajsowicz:
Description
A little patch for a couple of UI bugs I've found.

- It's no longer possible to open any menu type (MENU_MAIN, MENU_LOG etc.)
  while the game is loading or saving.
- Removed possibility to open the inventory while the dialog menu is active.

### DIFF
--- a/game/mainwindow.cpp
+++ b/game/mainwindow.cpp
@@ -372,6 +372,11 @@ void MainWindow::keyRepeatEvent(KeyEvent& event) {
   }
 
 void MainWindow::keyUpEvent(KeyEvent &event) {
+  // Don't process any key up events if the game is loading!
+  if (Gothic::inst().checkLoading()!=Gothic::LoadState::Idle) {
+    return;
+    }
+
   if(uiKeyUp==&video){
     video.keyUpEvent(event);
     if(event.isAccepted())
--- a/game/ui/dialogmenu.cpp
+++ b/game/ui/dialogmenu.cpp
@@ -520,6 +520,13 @@ void DialogMenu::mouseWheelEvent(MouseEvent &e) {
   }
 
 void DialogMenu::keyDownEvent(KeyEvent &e) {
+  // Explicitly accept this even so inventory won't open
+  // as long as dialog menu is active.
+  if(e.key==Event::K_Back) {
+    e.accept();
+    return;
+    }
+
   if(state==State::Idle || trade.isOpen()!=InventoryMenu::State::Closed){
     e.ignore();
     return;


Try:
> It's no longer possible to open any menu type (MENU_MAIN, MENU_LOG etc.)

While it's inline with original, have option to pause game upfront is nice sometimes. Can you leave only a `MENU_MAIN` option?

> Removed possibility to open the inventory while the dialog menu is active.

Here a screenshoot of steam version of the game. Pretty nice feature actually)
изображение

bwajsowicz:
> While it's inline with original, have option to pause game upfront is nice sometimes. Can you leave only a MENU_MAIN option

I think it's kinda useless to have this option while the game is loading/saving. I'll stay with the legacy behaviour. Maybe let's left MENU_MAIN option only for debug build?

> Here a screenshoot of steam version of the game. Pretty nice feature actually)
That's weird - I can't do the same on the Steam version. I had to spam the backspace key while the dialog menu is opening to get the same effect as yours - can't open the inventory when dialog menu is fully active.

Try:
> Maybe let's left MENU_MAIN option only for debug build?

Yep, that's fine.

>  I had to spam the backspace/tab key while the dialog menu

Vanilla is not 100% reliable at this, but it inventory is nice to have in dialog: player can switch spells/weapons before the fight; equip/remove rings and such.

Fabiosek:
> Pretty nice feature actually)

Nope.  This causes a lot of problems. This has been fixed in the latest version of Systempack 2.0 (Union).

Try:
Hi, @Fabiosek !
It seem this inventory case is arguable. In that case can you provide some details, on what kind of problems it causes?

On my side I can say: 
1. it's nice to have an option to check(and change) equipment, before battle; 
2. and to drop the gold in the right time

Fabiosek:
> 
> 
> Hi, @Fabiosek ! It seem this inventory case is arguable. In that case can you provide some details, on what kind of problems it causes?
> 
> On my side I can say:
> 
>     1. it's nice to have an option to check(and change) equipment, before battle;
> 
>     2. and to drop the gold in the right time

2. This way you can cheat without the codes. This allows you to rob traders (check out Lord Saradoc's videos). A good game should not have such bugs. Therefore, in Union or Systempack, a lot of such stupidities from the original have been corrected (especially in Union, the newer version of systempack)

1. I don't know if you've played Gothic mods, but it's very easy to bug them that way. It is enough for you to have a dialogue that requires having a specific item. If you don't give the item back right away, you can activate dialogue this way and quickly discard the item, which will "do nothing" to the NPC. This is strong cheating. Once I described it more broadly, but I do not have this content anymore, and I do not want to repeat myself. You do as you think. But recreating Gothic 1: 1 + his mistakes is not a good idea.

Try:
> discard the item, which will "do nothing" to the NPC

Found in vanilla scripts this case: dragons do not exchange `Eye of Innos` atomically. Moving to formal code-review then.

game/ui/dialogmenu.cpp
```diff
@@ -520,6 +520,13 @@ void DialogMenu::mouseWheelEvent(MouseEvent &e) {
   }
 
 void DialogMenu::keyDownEvent(KeyEvent &e) {
+  // Explicitly accept this even so inventory won't open
+  // as long as dialog menu is active.
+  if(e.key==Event::K_Back) {
```
Try, line 525: Inventory can be reassigned to different key

game/mainwindow.cpp
```diff
@@ -372,6 +372,11 @@ void MainWindow::keyRepeatEvent(KeyEvent& event) {
   }
 
 void MainWindow::keyUpEvent(KeyEvent &event) {
+  // Don't process any key up events if the game is loading!
+  if (Gothic::inst().checkLoading()!=Gothic::LoadState::Idle) {
+    return;
+    }
+
```
Try, line 379: should allow MENU_MAIN and F3

Try (changes requested):

Try:
Discard pull request, since there is no response to requested changes.
Inventory + Dialog interaction is already covered by ffb73cad

PR #209, opened 2021-11-13
Add Clang compatibility

lmichaelis:
This PR adds support for the Clang compiler. The build mostly works out-of-the box, however the `-Wno-format-truncation` is not supported by Clang and an additional error was being generated by `DamageCalculator::rangeDamage(Npc&, Npc& const Bullet&, const CollideMask)` where an integer enum value was multiplied with a floating point value.

I have fixed both of these issues. There is another [issue](https://github.com/Try/Tempest/pull/28) related to Clang compatibility in Tempest however, which needs to be merged too in order to properly support Clang.


Try:
Merged.
Have a little worry, about possible regressions in the future, since clang in not part of CI. But I guess it's OK, for now

Issue #214, opened 2021-11-21
Implement "Gothic 1 controls" switch in settings


Try:
Currently game always works in G1 controls flow: each attack-move requires 2 keys (action + w/s/a/d)
Need to implement G2 version as well

fhanrath:
I'd like to work on this issue (and already did so), but am not able to assign myself. Could you please assign me and [this](https://github.com/Try/OpenGothic/pull/217) draft?

PR #217, opened 2021-11-22
implemented gothic 2 controls

fhanrath:
Merge for https://github.com/Try/OpenGothic/issues/214

I noticed that the running attack seems to have a bug (even without this change).
In the gothic 1 control scheme the running attack sometimes doesn't work properly. Especially if you already did a normal attack.
In the gothic 2 control scheme the running attack always doesn't work properly.

When you try to initiate the running attack, the hero stops and does a normal attack which seems to hit twice.

I tried to find the issue in `npc.cpp doAttack`, but couldn't find anything related to the differentiation between a running and a normal attack.

Maybe it would be best to fix this problem in this branch, but I'm a bit lost right now.

The other attacks seem to work like vanilla gothic 2 with gothic 2 controls.


fhanrath:
I had a git misconfiguration and opened a new merge

PR #218, opened 2021-11-22
allow usage of mouse 4 and 5

fhanrath:
A small change that allows to set mouse 4 and 5 as bindings.
I wasn't sure about the hex numbers so I just incremented the second 0.

Seems to work fine.


fhanrath:
I had a git misconfiguration and opened a new merge

PR #220, opened 2021-11-23
allow usage of mouse 4 and 5

fhanrath:
A small change that allows to set mouse 4 and 5 as bindings.
I wasn't sure about the hex numbers so I just incremented the second 0.

Seems to work fine.

### DIFF
--- a/game/utils/keycodec.cpp
+++ b/game/utils/keycodec.cpp
@@ -67,8 +67,8 @@ std::initializer_list<KeyCodec::M_Key> KeyCodec::mkeys = {
   {Tempest::Event::ButtonLeft,    0x0c02},
   {Tempest::Event::ButtonMid,     0x0e02},
   {Tempest::Event::ButtonRight,   0x0d02},
-  {Tempest::Event::ButtonForward, 0x0e12},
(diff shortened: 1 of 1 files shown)


game/utils/keycodec.cpp
```diff
@@ -337,6 +339,14 @@ void KeyCodec::keyToStr(Tempest::Event::MouseButton k, char* buf, size_t bufSz)
     std::strncpy(buf,"MOUSE RIGHT",bufSz);
     return;
     }
+  if(k==Tempest::Event::ButtonForward) {
+    std::strncpy(buf,"MOUSE 5",bufSz);
+    return;
+    }
+  if(k==Tempest::Event::ButtonBack) {
+    std::strncpy(buf,"MOUSE 4",bufSz);
```
Try, line 347: In steam version of the game this key is `MOUSE X1` and `MOUSE X2` and code is `0x0f02`,`0x1002` (tested with UNION installed)
We should aligned to it

Try (changes requested):
Generally GTM, need to do a minor change (see comment)

Try:
Merged. Thanks for PR!

PR #223, opened 2021-11-24
implemented the marvin 'o' command

fhanrath:
Integrated the comments

### DIFF
--- a/game/world/world.cpp
+++ b/game/world/world.cpp
@@ -121,7 +121,9 @@ void World::setPlayer(Npc* npc) {
   if(npc==nullptr)
     return;
   npcPlayer->setProcessPolicy(Npc::ProcessPolicy::AiNormal);
-  npcPlayer->resumeAiRoutine();
+  if (!npcPlayer->isDead()) {
+    npcPlayer->resumeAiRoutine();
+  }
   
   npc->setProcessPolicy(Npc::ProcessPolicy::Player);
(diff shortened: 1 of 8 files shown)


game/world/world.cpp
```diff
@@ -117,6 +117,22 @@ void World::insertPlayer(std::unique_ptr<Npc> &&npc, std::string_view waypoint)
     game.script()->setInstanceNPC("HERO",*npcPlayer);
   }
 
+void World::setPlayer(Npc* npc) {
+  if(npc==nullptr)
+    return;
+  npcPlayer->setProcessPolicy(Npc::ProcessPolicy::AiNormal);
+  npcPlayer->resumeAiRoutine();
```
Try, line 124: That is still deferred, now - just wrapped :)
You can have design, like:
```c++
void Npc::resumeAiRoutine(){ /* actual work */ }
```
and
```c++
case AI_ContinueRoutine:
  resumeAiRoutine();
  break;
```

fhanrath: Thanks for your patience and help :)

Try:
Find a bug, while testing the last revision: if your character, `npc1` to name it, is dead, and you take control of npc2, then `resumeAiRoutine` will resurrect `npc`.
Other thing seems to be fine.

Try:
Merged, thanks!

PR #284, opened 2022-08-21
Enable quickload

matthiakl:
Enables F9 quickload feature.


Try:
Hi, @matthiakl , thanks for PR!

Generally code look good, but I'm not sure, if this is a good feature to have: `F8` and `F9` are too close, what makes misc-kicks quite annoying.
Do you know, if there is anything in vanilla to workaround this problem?

matthiakl:
Quicksave is F5, so during normal play this should not be a problem. F8 is only used in marvin mode, so I don't think miss-clicking is a problem here.

Try:
Right, fix code-style and merged

PR #285, opened 2022-08-21
Read enabled .mod files from ini file

matthiakl:
Currently all files in Data/ and Data/modvdf are loaded. This prevents having multiple mods installed. With this change, only the .mod files listed in the ini file (with -game:\<inifile>) are loaded.

>  [FILES]
vdf=FunnyMod.mod (Gothic 1)
VDF=FunnyMod.mod (Gothic 2)
List of all .mod files created for the mod. If this is more than one, they are listed separated by spaces (FunnyMod1.mod FunnyMod2.mod ...). GothicStarter first copies all *.mod files from .../Data/ to .../Data/modvdf/ before starting a modification and then copies all specified files from .../Data/modvdf/ to .../Data. The value "VDF" must exist (even if it is empty) for GothicStarter to accept the INI. In G2, the special value "!<invalid>" can be used to explicitly disable reading by the GothicStarter. During development a VDF list is most useful. It is very important with G2 that the parameter GohticGame.mod is not here! This leads to the fact that all changes to the scripts are ignored. 

https://wiki.worldofgothic.de/doku.php?id=ini-datei

### DIFF
--- a/game/gothic.cpp
+++ b/game/gothic.cpp
@@ -116,7 +116,7 @@ Gothic::Gothic() {
         modvdfs.push_back(mod);
       }
     }
-  Resources::loadAssets(modvdfs);
+  Resources::loadVdfs(modvdfs);
 
   if(wrldDef.empty()) {
     if(version().game==2)
--- a/game/gothic.cpp
+++ b/game/gothic.cpp
@@ -6,6 +6,8 @@
 #include <zenload/zCMesh.h>
 #include <cstring>
 #include <cctype>
+#include <vector>
+#include <sstream>
 
 #include "game/definitions/visualfxdefinitions.h"
 #include "game/definitions/sounddefinitions.h"
--- a/game/gothic.cpp
+++ b/game/gothic.cpp
@@ -6,9 +6,6 @@
 #include <zenload/zCMesh.h>
 #include <cstring>
 #include <cctype>
-#include <locale>
-#include <sstream>
-#include <vector>
 
 #include "game/definitions/visualfxdefinitions.h"
 #include "game/definitions/sounddefinitions.h"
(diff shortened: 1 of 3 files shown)


matthiakl:
https://ci.appveyor.com/project/Try/opengothic/builds/44531279/job/k1k02xtv7qyil1hc#L1688
Any idea why this fails to build on macos?

Try:
In error log:
> instantiation of undefined template 'std::ctype<char16_t>'

It seems MacOS doesn't support ctype for anything else than a `char`. so using `std::getline` is not an option. Generally it woruld be nice to avoid `basic_stringstream`, completely

game/gothic.cpp
```diff
@@ -100,13 +103,22 @@ Gothic::Gothic() {
...
 
+  std::vector<std::u16string> modvdfs;
   if(modFile!=nullptr) {
     wrldDef = modFile->getS("SETTINGS","WORLD");
     size_t split = wrldDef.rfind('\\');
     if(split!=std::string::npos)
       wrldDef = wrldDef.substr(split+1);
     plDef = modFile->getS("SETTINGS","PLAYER");
+
+    std::u16string vdf = TextCodec::toUtf16(std::string(modFile->getS("FILES","VDF")));
```
Try, line 114: As already mentioned in comments:
would be nice avoid `basic_stringstream` completely. I've tried to read cppreference, still can't say what is mandatory and what part of streams are optional in library :D 

Since you have string value - you can simply "for-loop" it.

game/gothic.cpp
```diff
@@ -100,13 +103,22 @@ Gothic::Gothic() {
...
+    std::basic_stringstream<char16_t> stringstream { vdf };
+    while (std::getline(stringstream, vdf, char16_t(' '))) {
+      if(!vdf.empty())
+        modvdfs.push_back(vdf);
+      }
     }
+  Resources::loadAssets(modvdfs);
```
Try, line 121: Nitpick: `Resources::loadVdfs`, to not confuse with actual game-assets

game/resources.cpp
```diff
@@ -102,9 +103,23 @@ Resources::Resources(Tempest::Device &device)
...
   }
+  }
 
+void Resources::loadAssets(const std::vector<std::u16string>& modvdfs) {
   std::vector<Archive> archives;
-  detectVdf(archives,Gothic::inst().nestedPath({u"Data"},Dir::FT_Dir));
+  inst->detectVdf(archives,Gothic::inst().nestedPath({u"Data"},Dir::FT_Dir));
+
+  // Remove all mod files, that are not listed in modvdfs
+  archives.erase(std::remove_if(archives.begin(), archives.end(),
```
Try, line 113: Nitpick: `std::erase_if(archives, ..`

matthiakl: `std::erase_if` is C++20 (https://en.cppreference.com/w/cpp/container/vector/erase2), you currently have C++17 set in CMakeLists. Do you want to update this now?

Try: Right, we need to update first. So, let's just leave `remove_if` for now then

Try:
Merged, thanks for PR!

PR #317, opened 2022-09-15
Play stow animation if a drawn weapon is unequipped

thokkat:
As a  followup to [9f01e8f](https://github.com/Try/OpenGothic/commit/9f01e8fdca346b32e0195c78733104cbc56cbe7e).

### DIFF
--- a/game/game/inventory.cpp
+++ b/game/game/inventory.cpp
@@ -356,6 +356,9 @@ bool Inventory::unequip(size_t cls, Npc &owner) {
   }
 
 void Inventory::unequip(Item *it, Npc &owner) {
+  if (it==activeWeapon())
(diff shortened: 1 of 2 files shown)


Try:
Hi. @thokkat !

I've tested the patch: doen't work as-is, if inventory gets open, when character just start to drawing the weapon. Easy reproduction:
click `2` to activate bow, and before bow ready - open inventory immediately.

thokkat:
Thanks for testing. I pushed a fix.

Still messing with git, this merge before looks wrong. Tell me if need to do something.

Try:
I've found extra bug: when in jump and weapon active, summoning inventory menu causes animation to stop completely. Fixed inline, by checking for `isStanding`

Issue #310, opened 2022-09-13
marvin F8 increases your weapon strength if used to revive yourself


thokkat:
Was wondering why i can one-shot everything :D. Don't know if it is original/intended behavior, at least i wasn't able to reproduce it in original NotR.

PR #323, opened 2022-09-18
Set correct Weapon Damage after dead/downstate

thokkat:
If you get killed or become unconscious while wielding a e.g. 1H weapon, every other 1H weapon you have or pick up will have  the dropped weapon's damage added.

For testing start a new game and grab a branch from the goblins at Cavalorn's place. Now let the bandit kill you, revive with F8 and  pick-up your branch or get a new one from the goblins. If you now attack the bandit, weapon damage is now 20 instead of 10.

fixes https://github.com/Try/OpenGothic/issues/310


Try:
Tested - works correct

Issue #120, opened 2021-02-01
Ladder doen't work


Try:
Ladder is not recognized as interactive object; climbing probably also doesn't work.
![изображение](https://user-images.githubusercontent.com/122894/106525304-11ddc600-64e4-11eb-87c6-2b20f7c9fd7f.png)

PR #325, opened 2022-09-21
Ladder Progress

thokkat:
Climbing upstairs works now. Might require some adaptions once downstairs climbing is possible.

https://github.com/Try/OpenGothic/issues/120


Try:

I have had a quick look at the code, one thing that is suspicious:
```
- void Interactive::nextState(Npc& npc)
```
Why this function is no longer needed? What about other interactive objects?

PS: will look more closely at the end of the week

thokkat:
It was only used in the processLadder function, which is now the ladderUp/Wait/Down functions.

Try:
Testing right now(Jarkentar only):
  can't climb down - ladder is not interactive. Can be unrelated issue
  when holding UpKey, upon climbing, there is a point when key-repeat stops working, at begin at climb

game/world/objects/interactive.cpp
```diff
@@ -215,6 +215,24 @@ void Interactive::tick(uint64_t dt) {
...
+  if (isLadder()) {
+    if (ladderFreeze && state>-1 && state<stateNum+1)
+      return;
+    if (state==stateNum && !reverseState) {
+      auto&  pl=*p->user;
+      auto  pos = pl.position();
+      float rot = pl.rotationRad();
+      float s   = std::sin(rot), c = std::cos(rot);
+      Tempest::Vec3 dp(s,0.0f,-c);
+      pos += dp*6.f*float(dt);
```
Try, line 228: `6.f` is arbitrary, right? While it's achieving the purpose, I have droughts about correctness. 
On previous iteration, (back at `nextState` days), I found that position of npc changes after detaching from ladder. We should probably exam closely animations here.

game/game/playercontrol.cpp
```diff
@@ -855,13 +868,17 @@ void PlayerControl::processPickLock(Npc& pl, Interactive& inter, KeyCodec::Actio
...
-    pl.setInteraction(nullptr);
+  if(key==KeyCodec::Back && !isPressed(KeyCodec::Forward)) {
+    inter.nextLadderMove(false,true);
     return;
     }
-
-  if(key==KeyCodec::Forward) {
-    inter.nextState(pl);
+  if(key==KeyCodec::Forward && !isPressed(KeyCodec::Back)) {
+    inter.nextLadderMove(false);
```
Try, line 876: It's probably better to stick with `nextState` notation, as regular MOBSI are also allowed to have multiple states.
Not in use in vanilla, but can be in mods.

game/game/playercontrol.cpp
```diff
@@ -855,13 +868,17 @@ void PlayerControl::processPickLock(Npc& pl, Interactive& inter, KeyCodec::Actio
   }
 
 void PlayerControl::processLadder(Npc& pl, Interactive& inter, KeyCodec::Action key) {
-  if(key==KeyCodec::Back) {
-    pl.setInteraction(nullptr);
+  if(key==KeyCodec::Back && !isPressed(KeyCodec::Forward)) {
+    inter.nextLadderMove(false,true);
```
Try, line 872: `false, true` - is not good from readability point of view.
It's probably better to handle key-repeat in `implMoveMobsi` routine.

game/world/objects/interactive.cpp
thokkat: Currently it's identical to marvin k. I tested some ladders in the G1 swamp camp and one only worked with 6f.


Hmm i have no other idea than "cheating" the character upwards. I only remember that devs said ladders were a nightmare so they left them out of G2 :D.  My test ladder in vanilla G1 funnily didn't work too and dropped the character back down at the top. 

I look into some other ladders in vanilla to see whats happening there

game/game/playercontrol.cpp
thokkat: Sure

game/game/playercontrol.cpp
thokkat: I kinda had that feeling but couldn't come up with something better, will change.

game/world/objects/interactive.cpp
```diff
@@ -215,6 +215,24 @@ void Interactive::tick(uint64_t dt) {
...
+      pl.setInteraction(nullptr);
+      pl.setPosition(pos);
+      return;
+      }
+    if (state==0 || state==stateNum)
```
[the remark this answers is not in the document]
thokkat: Actually this doesn't happen for me. Can it be keyboard related? Anyway you are right, changing this to -1, why else should it be excluded in the first if statement.

thokkat:
Yes non-interactive ladder upstairs was already the case. After some tries there's sometimes an interactive object found but not at the correct place, The climbing animation is played but the character doesn't move.

thokkat:
When you climb up, reach the end of the ladder and press the action key after the climbing animation ends, you are now again attached to the ladder but a bit shifted. For most ladders i tested that point looks like the vanilla endpoint after climbing is finished, but could also be coincidental.

Try:
Hi, @thokkat !
Sorry, for late response, bit to much code review in flight :)
for debugging, you can use `F9`/`F10`, to have frame-by-frame debugging:
F9 - pause
F10 - next frame

Setting MAX_FPS to low values in gothic.ini file can also be useful.

thokkat:
@Try, everything works so far. Only bug remaining is the stuttering issue. At the end of an animation the character is lifted for like 2-3 frames. This causes a bit of camera shaking sometimes after climbing upstairs is finished.

game/world/objects/interactive.cpp
```diff
@@ -291,22 +306,49 @@ void Interactive::implTick(Pos& p, uint64_t /*dt*/) {
...
 
-void Interactive::nextState(Npc& npc) {
-  const int prev = state;
-  if(!setAnim(&npc,Anim::In))
-    return;
-  setState(std::min(stateNum,state+1));
-  loopState = (prev==state);
+void Interactive::nextState(Npc& npc, const uint8_t st) {
+  wait = !st;
+  switch(st){
```
Try, line 311: `enum` ?

game/graphics/mdlvisual.cpp
```diff
@@ -591,9 +591,9 @@ const Animation::Sequence* MdlVisual::startAnimAndGet(Npc& npc, AnimationSolver:
     auto inter = npc.interactive();
     const Animation::Sequence *sq = solver.solveAnim(inter,a,*skInst);
     if(sq!=nullptr){
-      if(skInst->startAnim(solver,sq,comb,BS_MOBINTERACT,Pose::NoHint,npc.world().tickCount())) {
+      auto bs=sq->name.find("T_LADDER")==0 ? BS_CLIMB : BS_MOBINTERACT;
```
Try, line 594: There is no need to inspect animation names. Not to mention they are unreliable. 
`inter->isLadder()`

game/game/movealgo.cpp
```diff
@@ -44,10 +44,20 @@ void MoveAlgo::tickMobsi(uint64_t dt) {
     return;
 
   auto dp  = animMoveSpeed(dt);
-  if(!npc.interactive()->isLadder())
+  auto pos = npc.position();
+  if(npc.interactive()->isLadder()) {
+    auto rotY=-npc.rotationY();
+    if (std::abs(rotY)>0.001 && std::abs(dp.y)>0.001) {
```
Try, line 50: What this code do? I've tested without it - no notifiable difference.

game/world/objects/npc.cpp
```diff
@@ -409,7 +409,7 @@ void Npc::setDirectionY(float rotation) {
   if(rotation<-90)
     rotation = -90;
   rotation = std::fmod(rotation,360.f);
-  if(!mvAlgo.isSwim())
+  if(!mvAlgo.isSwim() && !(interactive()!=nullptr && interactive()->isLadder()))
```
Try, line 412: I don't mind such code(why only ladder tho?), but is this case should be already handled in `PlayerControl`

game/world/spaceindex.cpp
```diff
@@ -102,7 +102,7 @@ void BaseSpaceIndex::implFind(Vob** v, size_t cnt, uint8_t depth,
 
   auto mid = cnt/2;
   auto pos = v[mid]->position();
-  auto qR  = (R+500.0);//v[mid]->extendedSearchRadius());
+  auto qR  = (R+675.0);//v[mid]->extendedSearchRadius());
```
Try, line 105: Ideally need to index each attach point individually - can be a separated ticked

game/world/objects/npc.cpp
```diff
@@ -2975,7 +2975,7 @@ bool Npc::rotateTo(float dx, float dz, float step, bool noAnim, uint64_t dt) {
 
   const auto sgn = std::sin(double(da)*M_PI/180.0);
   if(sgn<0) {
-    setAnimRotate( 1);
+    setAnimRotate(1);
```
Try, line 2978: nitpick: space

game/world/objects/interactive.h
```diff
@@ -151,6 +151,7 @@ class Interactive : public Vob {
     int                          stepsCount = 0;
 
     int32_t                      state         = -1;
+    bool                         wait          = false;
```
Try, line 154: Can it be moved to PlayerControl ? Having this for every interactive object is unnecessary

Try (commented):
Look very good in game already - left some comments, need to cleanup code a little

game/world/objects/interactive.cpp
thokkat: I was reluctant because i thought it's unnecessary to have it in class definition only for ladder. But since you said `nextState` could be used for regular Mobsis in mods as well it' makes sense. And not as ugly as raw numbers of course.

game/graphics/mdlvisual.cpp
thokkat: `isLadder()` gave an incomplete class error and i didn't know how to deal with it. Should have stated that in the first place. Including the header fixed it.

game/game/movealgo.cpp
thokkat: For tilted ladders. Character fails to properly climb up at top. And if the ladder is attached to a wall, the character climbs into the wall if climbing down.

game/world/objects/npc.cpp
thokkat: No if at least one of  `isSwim()` or `isLadder()` is true the statement returns false and `angleY` is set. I need it to have access to the tilt factor in `tickMobsi` routine in `movealgo.cpp`.

game/world/objects/npc.cpp
thokkat: I removed the space. You want to keep it?

game/world/objects/interactive.h
thokkat: I could use a variable that already exists like `isLockCracked`. It even does the same just for chests and it should be safe since interactive can't be chest and ladder at the same time.

game/graphics/mdlvisual.cpp
Try: > isLadder() gave an incomplete class error and i didn't know how to deal with it.

Can you show error itself?

game/world/objects/interactive.h
Try: Idea is to have cleaner code by keeping player interaction stuff in `PlayerControl` and common stuff (npc and player) in `Interactive`.

game/graphics/mdlvisual.cpp
thokkat:
```
/home/tt/git/OpenGothic/game/graphics/mdlvisual.cpp:594: error: invalid use of incomplete type ‘class Interactive’
/home/tt/git/OpenGothic/game/graphics/mdlvisual.cpp: In member function ‘const Animation::Sequence* MdlVisual::startAnimAndGet(Npc&, AnimationSolver::Anim, uint8_t, WeaponState, WalkBit)’:
/home/tt/git/OpenGothic/game/graphics/mdlvisual.cpp:594:20: error: invalid use of incomplete type ‘class Interactive’
  594 |       auto bs=inter->isLadder() ? BS_CLIMB : BS_MOBINTERACT;
      |                    ^~
```

Try: `#include "interactive.h"`

game/world/objects/interactive.h
thokkat: I just don't see how. For chest 
```
  if(needToLockpick(npc)) {
      ...
      return; // chest is locked - need to crack lock first
      }

```

is checked to prevent playing the chest opening animation too early. The same is needed for ladder to check if player has up or down key pressed.

game/world/objects/npc.cpp
Try: Right, but what this code is for? `angleY` is set to be a swimming rotation, I don't see how this fit into laddering

game/game/movealgo.cpp
Try: Let's not do it for now - in test Jarkendar ladder works most of the time. Top-most ladder sometimes works poorly, but it same as in vanilla.

thokkat: They shouldn't be affected by it. There are only a few in G1.

game/world/objects/npc.cpp
thokkat: For tilted ladders. I need to know how much the character has to move forward and when to apply it.

thokkat:
Some footage so you know what i mean with tilted ladders.

https://user-images.githubusercontent.com/100468436/196058888-7322be56-b4a3-44c4-9f90-22f1c0900549.mp4

https://user-images.githubusercontent.com/100468436/196058889-ef350bce-50c7-4850-b7e7-9f0c1d4cfbc1.mp4

https://user-images.githubusercontent.com/100468436/196058957-8b7a87f7-e868-42f1-a1c2-f9fff06d2cd2.mp4

game/world/objects/npc.cpp
Try: Oh, now I think I'm getting it. Yep `angleY` is correct tool in such case.
Curious: is there any titled ladder in the game? Gothic1?

game/world/objects/interactive.h
Try: > Is this intended?

No, it's not. There is keyPress - ketRepeat - keyRelease workflow, should work same for any key-type.

thokkat:
I made the aforementioned changes. Tell me if there*s more to do.

thokkat:
@Try, reminder for review, now that phoenix has been merged ;) . G2 works as expected. In G1 there's now a general bug so animations aren't played.

game/game/playercontrol.cpp
```diff
@@ -196,12 +198,13 @@ void PlayerControl::tickFocus() {
...
 
-  if(!ctrl[Action::ActionGeneric])
+  if(!ctrl[Action::ActionGeneric] || Gothic::inst().world()->player()->interactive()!=nullptr)
     return;
 
   auto focus = currentFocus;
   if(focus.interactive!=nullptr && interact(*focus.interactive)) {
-    clearInput();
+    if (!focus.interactive->isLadder())
+      clearInput();
```
Try, line 207: What is the intent of this code? sorry, This related to key-repeat stuff, right?

game/game/playercontrol.cpp
```diff
@@ -196,12 +198,13 @@ void PlayerControl::tickFocus() {
   currentFocus = findFocus(&currentFocus);
   setTarget(currentFocus.npc);
 
-  if(!ctrl[Action::ActionGeneric])
+  if(!ctrl[Action::ActionGeneric] || Gothic::inst().world()->player()->interactive()!=nullptr)
```
Try, line 201: 1. nullptr checks - there is no guaranies , that player always will be around
2. What is the purpose if this change?

game/game/movealgo.cpp
Try: This is quite awkward code still :(
Why `std::abs`? There is seemingly no divisions downstream or something. And `cos/sin` should be way faster than the `if`

AFAIR this is G1 only, right? Can we decouple this change into another PR. This `*=-2` fell like fine tuning; `Vec3(s,0,-c)` - in  other places it's usually `Vec3(c,0,s)`

game/ui/inventorymenu.cpp
```diff
@@ -105,7 +105,7 @@ void InventoryMenu::close() {
   }
 
 void InventoryMenu::open(Npc &pl) {
-  if(pl.isDown() || pl.isMonster() || pl.isInAir() || pl.isSlide() || (pl.interactive()!=nullptr))
+  if(pl.isDown() || pl.isMonster() || pl.isInAir() || pl.isSlide() || pl.interactive()!=nullptr)
```
Try, line 108: nitpick: code-style only change - not related to PR

Try (commented):
movealgo trigonometry is still suspicious. Other  than that look good.

game/game/playercontrol.cpp
thokkat: It's for G1. In vanilla, if you interact with ladder and keep action key pressed you stay attached to ladder. Without this change you are released.

game/game/playercontrol.cpp
thokkat: Again for G1 controls. Some ladders showed buggy behavior when interacting at top. This change prevents interaction, if player already interacts.

Try: Should this kind of check then be placed into `Focus World::findFocus(const Npc &pl, const Focus& def)`?
if player already interacts with something, then search can simply early-out.

game/game/movealgo.cpp
thokkat: I only want to test if both values are nonzero, is there a better way? If `rotY=0` only a zero vector is added, so this check could be left out. `std::abs(dp.y)>0.001` is needed to know if the character is moving and don't add anything if not.

`Vec3(s,0,-c)` stuff is copied from `marvinF8/K` routines. Factor `2` is needed to make downstairs climbing look nice and yeah tilted ladders occur only in G1.

game/ui/inventorymenu.cpp
thokkat: So this should be a separate PR?

Try: No - I'll fix it as separated commit, as followup

thokkat:
@Try, I cleaned up the `abs` stuff. `Vec3(c,0,s)` moves the character leftwards and i only found it in `camera.cpp`. `Vec3(s,0,-c)` is copied from `marvinF8/K` routines,

Try:
Merged, thanks!

PR #326, opened 2022-09-22
Allow open inventory while walking

thokkat:
A small update to https://github.com/Try/OpenGothic/pull/317.  Currently if you press the inventory key while walking the character just stops instead of opening the inventory.


Try:
Hi, @thokkat ! Thanks for yet another PR :)
Tested, merged.

Issue #309, opened 2022-09-13
Sengrath's corpse has no crossbow in the inventory


thokkat:
...thus you can't finish Udar's quest. In original NotR you can loot his trader inventory instead of the usual "loot inventory" and get the crossbow.

Try:
Some thoughts:

    Sengrath is a merchant
    Merchant inventory is cleared automatically in ZS_DEAD
    Mission items (ITEM_MISSION) should not be cleared, but his cross-box is not marked as ITEM_MISSION

My current view on this: if npc is killed by script ZS_DEAD should not be called then. Need to check with vanilla here: does Sengraths goods are still on him?

thokkat:
From youtube Let's plays it looks like he has his normal inventory and everything you sold to him but not his sword

Try:
> and everything you sold to him but not his sword.

Atm(after [b80f2b0](https://github.com/Try/OpenGothic/commit/b80f2b0615842e2644f0d813b86d9c19d43e935f)) OpenGothic will show sword as well. It's likely that equipped stuff should not be shown at all (`Inventory::Iterator::skipHidden`), but need to verify

thokkat:
Looks good 
![Screenshot_20220925_160902](https://user-images.githubusercontent.com/100468436/192148072-3115ab56-7762-4a7c-9a13-2b5308950d94.png)

PR #327, opened 2022-09-24
Add Sengrath's crossbow to his inventory if he's dead

thokkat:
... so you can solve the corresponding quest. I attached a save file to verify. Also removed unnecessary parentheses.

 [save_slot_1.zip](https://github.com/Try/OpenGothic/files/9638752/save_slot_1.zip)

fixes https://github.com/Try/OpenGothic/issues/309

Ah just noticed this can be abused if not the entire inventory is looted. Looking into it.

### DIFF
--- a/game/ui/inventorymenu.cpp
+++ b/game/ui/inventorymenu.cpp
@@ -104,7 +104,7 @@ void InventoryMenu::close() {
   }
 
 void InventoryMenu::open(Npc &pl) {
-  if(pl.isDown() || pl.isMonster() || pl.isInAir() || pl.isSlide() || (pl.interactive()!=nullptr))
+  if(pl.isDown() || pl.isMonster() || pl.isInAir() || pl.isSlide() || pl.interactive()!=nullptr)
     return;
   if(pl.weaponState()!=WeaponState::NoWeapon) {
     pl.stopAnim("");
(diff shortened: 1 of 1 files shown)


Dehela:
I don't know game internals so I might be not right here, but I think it's a wrong way to fix the issue by adding the item to the body like that, hardcoding in game engine code (if anything, it should be done by fixing game script). Better way of fixing it is figuring out why the item is missing, because there are good chances that similar issue is happening in some other places too, so better resolve the root of the problem than relying on quickfix(es).

game/ui/inventorymenu.cpp
```diff
@@ -145,6 +145,9 @@ bool InventoryMenu::ransack(Npc &pl, Npc &tr) {
   auto it = tr.inventory().iterator(Inventory::T_Ransack);
   if(!it.isValid())
     return false;
+  // fix quest "Lost in Darkness"
+  if (tr.displayName()=="Sengrath" && tr.isDead())
+    tr.addItem(7674,1);
```
PolyMeilex, line 150: I'm not really Gothic expert but this looks suspicious at best. What if there is a character named Sengrath in some mod, did the original game have random quest specific code scattered in the engine code?

To me it sounds like there is some sort of bug in handling of this specific quest, that should be fixed rather than hardcoding a workaround into the engine.

I might be completely wrong tho, I'm just a random bystander.

Try:
Hi, @thokkat !

Have to rejecting this PR, due to obvious reasons.

Some summary thoughts:
* Sengrath is a merchant
* Merchant inventory is cleared automatically in `ZS_DEAD` 
* Mission items (`ITEM_MISSION`) should not be cleared, but his cross-box is not marked as `ITEM_MISSION`

My current view on this: if npc is killed by script `ZS_DEAD` should not be called then. Need to check with vanilla here: does Sengraths goods are still on him?

PR #328, opened 2022-09-25
Unequip Sengrath's sword

thokkat:
The unequip routine is not called upon his death, so do it manually.
https://github.com/Try/OpenGothic/issues/309

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1141,8 +1141,10 @@ void Npc::changeAttribute(Attribute a, int32_t val, bool allowUnconscious) {
 
   if(a==ATR_HITPOINTS) {
     checkHealth(true,allowUnconscious);
-    if(aiPolicy==AiFar || aiPolicy==AiFar2)
+    if(aiPolicy==AiFar || aiPolicy==AiFar2) {
       aiState.started = true;
+      invent.unequipWeapons(owner.script(),*this);
+      }
     }
   }
 


Try:
Why do you think it should be unequipped? Can you confirm that with original game?
Note: original game also has somewhat inconsistent state, when npc is dead, but his weapons are equipped and inaccessible to player.

thokkat:
Just to remove red equipped rectangle when looting. When beginning chapter 4 AI_UnEquipWeapons is called for all the newly spawned dead dragon slayers but not for Sengrath and the removed npc's like Milten.

Try:
> When beginning chapter 4 AI_UnEquipWeapons is called for all the newly spawned dead dragon slayers but not for Sengrath and the removed npc's like Milten.

This looks to me more like script issue. If behavior is consistent with original, than we should do anything on engine side

thokkat:
You mentionend `Inventory::Iterator::skipHidden`  in the issue before. Maybe also add equipped weapons to it since sword is not visible in original.

Try:
Can you test this kind of code? :
```c++
  if(type==T_Ransack) {
    // while(at<it.size() && ((it[at]->isEquiped() && it[at]->isRune()) || (it[at]->isEquiped() && it[at]->isArmour()))) {
    while(at<it.size() && it[at]->isEquiped()) {
      ++at;
      }
```
Unfortunately can't test on my side: vanilla game not working again for me - so can't assure match :(

thokkat:
In vanilla weapons are unequipped if a npc is knocked down, which is already the case in OpenGothic. So only mismatchs could be npc's killed by script. In my testing at least everything looks correct.

The only character i found was the dead militia Den, but this is not directly related to the code change. Script gives him this item `EquipItem(self,ItMw_1h_Vlk_Dagger);`. It's shown in OpenGothic because it's not equipped, while In vanilla it's equipped and therefore not shown. He has an old coin instead which in turn isn't present in OpenGothic.

thokkat:
Fixed in https://github.com/Try/OpenGothic/commit/5c19d609c992c5a64557acffb8010b8c4427e154.

PR #330, opened 2022-09-29
Remove kuba-zip from submodules

Newbytee:
This causes issues with cloning OpenGothic in some setups, and I don't think it was intended to be left here.


Try:
Right, since we are using minizip, kuba is not needed

Newbytee:
> Right, since we are using minizip, kuba is not needed

Additionally, the gitlink inside of the lib/ directory was removed previously, so it wouldn't have been cloned regardless.

PR #332, opened 2022-10-03
allow using Enter in addition to LMB in journal and inventory

Nindaleth:
Some game UI parts only support the left mouse button click and not the Enter button for certain actions, making the control a bit more cumbersome on Steam Deck than it needs to be. This PR allows to show quest details and to interact with items in the inventory using the Enter key. This improves user experience when playing on Steam Deck.

Since neither the original nor OpenGothic support gamepad controls, the Steam Deck uses Steam Input to map the physical buttons to simulated keyboard/mouse button presses. In the input mapping assigned by default, the A button is mapped to Enter key.


Try:
Hi, @Nindaleth !

Looks good, thanks!

PR #335, opened 2022-10-08
Adress some swimming/diving related issues

thokkat:
Fixes the following bugs:

- Camera stays under the character after diving to water surface.
- ~~After flying out of water by help of marvin F8 the character stays in swimming state.~~
- Too big health reduction after diving time is over. Vanilla is like 10dmg/s.

### DIFF
--- a/game/camera.cpp
+++ b/game/camera.cpp
@@ -108,7 +108,7 @@ void Camera::setMode(Camera::Mode m) {
   if(camMod==m)
     return;
 
-  const bool reset = (m==Inventory || camMod==Inventory || camMod==Dialog);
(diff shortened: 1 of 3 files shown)


thokkat:
Ok, waterRay gives wrong values if the water surface is not free. Guard against it by checking for !isDive().

thokkat:
Actually swimming case works. Problematic cases are when you dive through the dam in the new camp in G1 or the paladin ship. There was also a bug depending on the orientation of the character.

I kicked it out and think about it again. I open a PR if i come up with something useful.

thokkat:
@Try Is there something that needs to be changed? I left marvin stuff out.

Try:
Hi, sorry I kinda forgot about this PR, thank for poking me.
Generally LGTM - will do testing today

Issue #296, opened 2022-09-06
Long wait time until dialogbox pops up


thokkat:
When you interact with Bones or Cord it takes quite a while until they talk to you. I thought this was classical behavior in NotR but in a quick test with Cord  he reacted almost instantly.

Try:
Tested locally - should 'just work'

thokkat:
Yep works, thanks.

PR #336, opened 2022-10-08
stop Npc animation if player wants to talk

thokkat:
For long duration animations like the sword training.

fixes https://github.com/Try/OpenGothic/issues/296

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -3510,10 +3510,13 @@ void Npc::setPerceptionDisable(PercType t) {
   }
 
 void Npc::startDialog(Npc& pl) {
-   if(pl.isDown() || pl.isInAir() || isPlayer())
+  if(pl.isDown() || pl.isInAir() || isPlayer())
     return;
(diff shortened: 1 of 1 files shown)


Try:
Hi, @thokkat !

This solution doesn't seems to be complete. The `1H_FREE` and other transition animations are problematic not only for dialogs:
изображение

I'm not sure what is correct behavior here. Maybe hookup on ai_clearqueue instead

thokkat:
Are there cases where an animation isn't allowed to be interrupted? Otherwise something like this:
```
    case AI_PlayAnim:{
      if(auto sq =playAnimByName(act.s0.c_str(),BS_NONE)) {
      // implAniWait(uint64_t(sq->totalTime()));
        } else {
```

Try:
> Are there cases where an animation isn't allowed to be interrupted?

Intimidation animations for monsters - to make sure work will not run forward, while barking

thokkat:
Close because proper solution for all cases needs to be figured out.

PR #338, opened 2022-10-13
update Readme, add Fedora compilation info, document -dx12

Nindaleth:
The result can be seen in practice here: https://github.com/Nindaleth/OpenGothic/tree/feature/readme-fedora

Summary of changes:

* improve individual sections, but don't change the section order or presence - most obvious refactors are in the Linux compilation and the command line argument description sections
* document dependencies for compiling on Fedora
* document the -dx12 flag
* make the headlines bit more headline-y


Try:
Looks good to me. Merged, thanks!

Issue #349, opened 2022-11-03
[MiniMod_Balance] Game reinitializes when coming back from Jharkendar first time


sirsteve1:
Hi there

I have a strange issue. When entering Jharkendar and coming back to the new world the first time, the game reinitializes the new world. The intro-videos are replayed, the portal is closed and everything world-related is set back to start (npcs, monsters, items). Quest log, npc dialog options and player stats are correct. Water mages are not present in NewWorld but in Jharkendar (seems Quest related), yet Greg (probably due to the error that he is not appearing next to Dexter) is still in NewWorld even though he is also in Jharkendar as I already visited the pirates camp.

When I now re-enter Jharkendar (by reactivating the portal) and coming back to NewWorld a second time, the portal stays open and no intro videos are loaded (I guess that now the initialized state is stored correctly).

Is there a way to get around the reinitiliazation? Or checking the chapter, as I am already in chapter 2?

This is really an odd problem, also because all merchant trading inventories now are empty in NewWorld (which is really bad).
Don't know if this is only for me.

I am playing though using the MiniMod Balance mod. So far everything worked quite well.

sirsteve1:
So I did find the issue regarding the reinitialization of the game.

I do play with MiniMod_Balance activated. So far all worked great. The problem arises during world changes back to newworld. The world name is handled differently, probably due to the installed mod. This might be a problem that arises also with other mods installed.

Without any mod, starting the game the world name is `newworld.zen`.
Using __MiniMod_Balance__, the world name is `NewWorld.zen`. 

As long as you are in the initial world, everything is fine. Switching to another world also is no problem. Switching back to the starting world for the first time tries to load `newworld.zen` and finds no stored WorldStateStorage, thus restarting the game with a full quest log and character stats. You start off at the closed (as initialzed) portal with a completely mixed game state.

I was able to solve the problem temporarily by modifiying the passed world name within `GameSession::implChangeWorld` in `gamesession.cpp`:
```
auto GameSession::implChangeWorld(std::unique_ptr<GameSession>&& game,
                                  const std::string& world, const std::string& wayPoint) -> std::unique_ptr<GameSession> {
  
  const char*   w   = world.compare("newworld\\newworld.zen") == 0 ? "newworld\\NewWorld.zen" : world.c_str();
  // const char*   w   = world.c_str();
  size_t        cut = world.rfind('\\');
```

[...]

PR #350, opened 2022-11-08
Respawn System

sirsteve1:
Dead @Try 

Though I used it initially for playing around with the code, my expiremt went so well that I created a PR.

The respawn system mostly implements the class RespawnObject which uses itself as a RespawnObject for monsters that will initialized and stored within a static respawn-list vector object. There are two main methods:

- `RespawnObject::registerObject` which is called when a npc dies. Here respawnable monsters will added to the vector list.
- `RespawnObject::processRespawnList` is the actual respawn routine which makes some initial tests and then walks through the list and respawns all monsters that reached a precalculated respawn day. The method is called within `GameScript::wld_settime` and `MainWindow::onWorldLoaded`. The ladder is for switching between worlds as wld_settime is called when the world is not yet there (nullptr) during world-change events. To avoid processing after loading a savegame I also added a change-world flag.

Regarding __what will respawn__ (see `RespawnObject::npcShouldRespawn`):
- Only monsters, no humans or orcs
- No dragons, shadowbeasts & alike, summoned monsters, stoneguardians

Regarding __when monsters will respawn__:
- There are two static uint values `RespawnObject::minDays` and `RespawnObject::maxDays`
- With minDays and maxDays monsters will get registered with _respawnDay = currDay + minDays + random( maxDays - minDays )_
- When the current day is equal or greater than the stored respawn day, the monster will respawn

The respawn system comes with some additional __functionality__:
- Global activation by cli option `-respawn` (also adapted Readme)
- Cache-processing for fast-skip if the current day in the current world has already been processed
- Added marvin commands `respawn [show|clear|process]`
- Own log and screen-print messages. Except for marvin commands screen-print is only processed when doFrate is active
- Serialization with support for savegames without respawn-list stored (backward-compatible savegame format) 

I tested everything, from simply respawn, monsters that should not respawn and switching between worlds. Respawn can only happen within the current loaded world, thus the current world is stored alongside when registering the respawn entries. When entering a world (world-change) the list is processed and all registered respawns for that world (with `currDay >= respawnDay`) are respawned immediately. Same when sleeping.

Greetings
Steve

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1767,7 +1767,7 @@ void Npc::takeDamage(Npc& other, const Bullet* b, const CollideMask bMask, int32
       else if(isDead()) {
         owner.sendPassivePerc(*this,other,*this,PERC_ASSESSMURDER);
         // Register monsters for respawn (method handles human / monster / orc evaluation)
-        RespawnObject::registerObject(size_t(instanceSymbol()), hnpc.wp.c_str(), guild());
+        RespawnObject::registerObject(size_t(instanceSymbol()), hnpc->wp.c_str(), guild());
         }
       else {
         if(owner.script().rand(2)==0)
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -10,6 +10,7 @@
 #include "game/gamescript.h"
 #include "world/objects/interactive.h"
 #include "world/objects/item.h"
+#include "world/respawnobject.h"
 #include "world/world.h"
 #include "utils/versioninfo.h"
 #include "utils/fileext.h"
@@ -1765,6 +1766,8 @@ void Npc::takeDamage(Npc& other, const Bullet* b, const CollideMask bMask, int32
         }
       else if(isDead()) {
         owner.sendPassivePerc(*this,other,*this,PERC_ASSESSMURDER);
+        // Register monsters for respawn (method handles human / monster / orc evaluation)
+        RespawnObject::registerObject(size_t(instanceSymbol()), hnpc.wp.c_str(), guild());
         }
       else {
         if(owner.script().rand(2)==0)
--- a/game/world/respawnobject.cpp
+++ b/game/world/respawnobject.cpp
@@ -24,12 +24,10 @@ void RespawnObject::save(Serialize &fout) const{
 }
 
 void RespawnObject::registerObject(size_t inst, std::string wp, uint32_t guild){
-    if (!RespawnObject::npcShouldRespawn(guild)) {
-        printMsg("Not a respawnable monster (Guild ID " + std::to_string(guild) + ")");
+    if (!RespawnObject::npcShouldRespawn(guild))
         return;
-    }
     World *world = Gothic::inst().world();
-    // Skip also if the waypoint is not findable
+    // Skip if waypoint is unknown
     if (!world->findPoint(wp))
         return;
     RespawnObject ro;
@@ -47,8 +45,11 @@ void RespawnObject::registerObject(size_t inst, std::string wp, uint32_t guild){
 }
 
 void RespawnObject::processRespawnList(){
+    World* world  = Gothic::inst().world();
+    // Ensure respawn list is not processed during world transition / loading sequences
+    if(world==nullptr || Gothic::inst().player()==nullptr)
+        return;
     printMsg("Process respawn list, currently " + std::to_string(respawnList.size()) + " monsters registered");
-    World *world = Gothic::inst().world();
     uint32_t currDay = (uint32_t)world->time().day();
     uint32_t cnt = 0;
     std::vector<RespawnObject>::iterator iter;
--- a/game/world/respawnobject.cpp
+++ b/game/world/respawnobject.cpp
@@ -1,0 +1,113 @@
+#include "respawnobject.h"
+
+#include "game/serialize.h"
+#include "gothic.h"
+#include <string>
+#include <vector>
+#include <sstream>
+#include <stdarg.h>
+
+using namespace Tempest;
+
+std::vector<RespawnObject> RespawnObject::respawnList = {};
+
+RespawnObject::RespawnObject(Serialize &fin){
+    fin.read(inst,wp,respawnDay);
+}
+
+void RespawnObject::save(Serialize &fout) const{
+    fout.write(inst,wp,respawnDay);
+}
+
+void RespawnObject::registerObject(size_t inst, std::string wp, uint32_t guild){
+    if (!RespawnObject::npcShouldRespawn(guild)) {
+        printMsg("Not a respawnable monster (Guild ID " + std::to_string(guild) + ")");
+        return;
+    }
+    RespawnObject ro;
+    ro.inst = inst;
+    ro.wp = wp;
+    uint32_t minDays = 3; // Should be at least 1
+    uint32_t maxDays = 10;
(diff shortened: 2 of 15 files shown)


game/world/objects/npc.cpp
```diff
@@ -1765,6 +1766,8 @@ void Npc::takeDamage(Npc& other, const Bullet* b, const CollideMask bMask, int32
         }
       else if(isDead()) {
         owner.sendPassivePerc(*this,other,*this,PERC_ASSESSMURDER);
+        // Register monsters for respawn (method handles human / monster / orc evaluation)
+        RespawnObject::registerObject(size_t(instanceSymbol()), hnpc->wp.c_str(), guild());
```
Try, line 1770: This might break in swampshark in Addon:
```
instance MIS_Addon_Swampshark_Lou(Mst_Default_Swampshark) <--- normal shark
{
	B_SetVisuals_Swampshark();
	Npc_SetToFistMode(self);
	CreateInvItems(self,ITKE_Addon_Tavern_01,1);  <--- Quest item!
};
```
Again, see `WorldObjects::resetPositionToTA`/`Npc::resetPositionToTA` - body cleanup handle there

game/world/respawnobject.cpp
```diff
@@ -0,0 +1,189 @@
...
+    World* world  = Gothic::inst().world();
+    // Ensure respawn list is not processed during world transition / loading sequences
+    if(world==nullptr || Gothic::inst().player()==nullptr) {
+        printMsg("Skip processing respawn list (world not fully loaded)");
+        return;
+    }
+    uint32_t currDay = getCurrDay(world);
+    // Check against cache value for early abort (no multi-processing on same day in same world)
+    std::string currDayWld = std::to_string(currDay) + std::string("_") + std::string(world->name());
+    if (processedDayWorld.compare(currDayWld) == 0) {
```
Try, line 61: Case sensitivity? As touched in https://github.com/Try/OpenGothic/issues/349 mod may have mixed case in world name and get away with it. Original game seems using regular windows-files to store per-world state

game/world/respawnobject.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+RespawnObject::RespawnObject(Serialize &fin){
+    fin.read(inst,wp,respawnDay,world);
+}
+
+void RespawnObject::save(Serialize &fout) const{
+    fout.write(inst,wp,respawnDay,world);
+}
+
+void RespawnObject::registerObject(size_t inst, std::string wp, uint32_t guild){
```
Try, line 27: Path seeming passes string by value a lot with little reason to do so.
And also performs copy, instead of move in this function.

Try (commented):
Will mod based solution be more applicable?
Script language, technically already has `Wld_InsertNpcAndRespawn` api-point and `c_nps::spawnPoint`+`c_nps::spawnDelay` - need only to implement logic behind.

Also having respawn settings in *.ini file allow you to use `onChgSetOption`, when working with game-ui on a script side

game/world/respawnobject.cpp
```diff
@@ -0,0 +1,189 @@
...
+std::string RespawnObject::processedDayWorld = "";
+uint32_t RespawnObject::minDays = 2;  // Should be at least 1
+uint32_t RespawnObject::maxDays = 10; // Must be bigger than minDays
+
+RespawnObject::RespawnObject(Serialize &fin){
+    fin.read(inst,wp,respawnDay,world);
+}
+
+void RespawnObject::save(Serialize &fout) const{
+    fout.write(inst,wp,respawnDay,world);
```
Try, line 24: `size_t inst;` - serializing variable-sized type will break 32bit vs 64bit save file compatibility

sirsteve1:
Thanks for your replies. I will check them out later and go through.

> Will mod based solution be more applicable? Script language, technically already has `Wld_InsertNpcAndRespawn` api-point and `c_nps::spawnPoint`+`c_nps::spawnDelay` - need only to implement logic behind.
> 
> Also having respawn settings in *.ini file allow you to use `onChgSetOption`, when working with game-ui on a script side

Regarding your initial question: Surely mod based solutions will also work, but for people like me playing with mods like _MiniMod_Balance_ you cannot simply add a respawn mod as well, as this might break everything. Also I feel like having this as opt-in even without any mods brings more value to OpenGothic. But this is my opinion, surely you might also simply say you only want to have vanilla Gothic without additional improvements.  

For `Wld_InsertNpcAndRespawn`, I checked the function but found out that it actually is not used anywhere. I guess its more useful within mods. Using the function requires also storing the dead npc's somewhere for respawn later, as the function  `Wld_InsertNpcAndRespawn` inserts them directly. This would lead to instant respawn for the first death (at least I understood the function like that). And later you would discern if the dead NPC is intially inserted or already respawned once to not respawn it twice... So I came up with my solution. But I can recheck the api points `c_nps::spawnPoint`+`c_nps::spawnDelay`, have not seen them up until now. Anyway, there still is the problem about spawning npcs in other worlds -- which brings me back to storing the respawnable npcs and waiting for the correct world to spawn them.

Try:
Hi @sirsteve1 !

For mod-based case my point is:
engine suppose to provide an api for scripting language to use, but avoid to implement game-related functionality. 
if there is already some-sort of solution, we should better align to it.

At this point we have `Wld_InsertNpcAndRespawn`, from mod kit. It's broken in original game, but we can fix it.
It would be a little odd, if we will introduce orthogonal way to have respawn, while ignoring legacy api.

Nindaleth:
Hi @sirsteve1 ! Could you introduce the intended functionality a bit? From your initial description I have a good idea of how it works, but I don't know _why_ I want to use the mod/parameter. 

Why would you want to enable it? Is the game more fun when the world is never completely empty? Is it intended to allow easier XP gains? Does this provide better atmosphere without a major impact as only non-rare monsters are continuously replenished? :-)

sirsteve1:
> Hi @sirsteve1 ! Could you introduce the intended functionality a bit? From your initial description I have a good idea of how it works, but I don't know _why_ I want to use the mod/parameter.
> 
> Why would you want to enable it? Is the game more fun when the world is never completely empty? Is it intended to allow easier XP gains? Does this provide better atmosphere without a major impact as only non-rare monsters are continuously replenished? :-)

Dear Nindaleth

Thanks for your question. I would like to explain how I came to the conclusion in doing a respawn system. I restarted playing the Gothic series during this year, haven't played it for a while. I also tried some mods, like Legend of Ahssun, which comes with an implemented timely respawn system likewise as I tried introducing it. As I play on Linux playing the game with Wine always gives some sort of headache, so I stumpled across OpenGothic. As I realized the strength of playing via OpenGothic yet realized all the open issues, I though of a simple feature to implement to start playing around with all the code. So I decided to implement a similar respawn system as it is with Legend of Ahssun. 

This is mostly why I created it. Like you said, it is quite fun to run around a world that is never empty, giving also some extra XP without fully overpowering the game. E.g. when I now use the teleporter stones in Jharkendar, sometimes I immediately find myself teleporting next to some desert rats or other monsters. That is surely no game changer in experience (probably some more levels at the end), yet makes the world still somehow more vivid. The fact that it changes the original game as it was intended was an important thing for me to create an opt-in respawn system.

For me the system works already good, as I wrote, it was more as a small side-project to get into the code to hopefully also work on some issues OpenGothic still has. I also see all the flaws @Try noted and will try to work soon on them to hopefully create a mature opt-in respawn system for OpenGothic.

Try:
closing as outdated

PR #363, opened 2022-11-21
Set diving state to false after marvinF8 moving to air

thokkat:
This time it should work. There are still some corner cases like the paladin ship. If you F8 up from the seabed you stay in diving mode until you reach the deck. But in the end you can escape dive/swimming state.

I guess It's better to have everything in `movealgo.cpp` but there we don't know if the movement is caused by marvin or normal play. 

Is there a better way to access the move states from the marvin routines  than introducing 2 new functions?

### DIFF
--- a/game/game/movealgo.cpp
+++ b/game/game/movealgo.cpp
@@ -308,11 +308,13 @@ bool MoveAlgo::tickRun(uint64_t dt, MvFlags moveFlg) {
   const auto  dp            = npcMoveSpeed(dt,moveFlg);
   const auto  pos           = npc.position();
   const float fallThreshold = stepHeight();
+  const auto  chest         = waterDepthChest();
 
   // moving NPC, by animation
   bool  valid   = false;
+  bool  validW  = false;
   auto  ground  = dropRay (pos+Tempest::Vec3(0,fallThreshold,0), valid);
-  auto  water   = waterRay(pos);
+  auto  water   = waterRay(pos,&validW);
   float dY      = pos.y-ground;
   bool  onGound = true;
 
@@ -371,6 +373,13 @@ bool MoveAlgo::tickRun(uint64_t dt, MvFlags moveFlg) {
       // start to fall of cliff
       auto dpCliff = (dp==Tempest::Vec3()) ? Tempest::Vec3(cache.n.x,0,cache.n.z)*float(dt) : dp;
       if(tryMove(dp.x,dp.y,dp.z)){
+        if(validW) {
+          setInWater(true);
+          setAsSwim (true);
+          if(pos.y+chest<water)
+            setAsDive(true);
+          return false;
+          }
         fallSpeed.x = 0.3f*dpCliff.x;
         fallSpeed.y = 0.f;
         fallSpeed.z = 0.3f*dpCliff.z;
@@ -734,6 +743,7 @@ bool MoveAlgo::startClimb(JumpStatus jump) {
 void MoveAlgo::startDive() {
   if(isSwim() && !isDive()) {
     if(npc.world().tickCount()-diveStart>1000) {
+      npc.setDirectionY(-40);
       setAsDive(true);
 
       auto  pos   = npc.position();
@@ -743,6 +753,16 @@ void MoveAlgo::startDive() {
       tryMove(0,water-chest-pY,0);
       }
     }
+  }
+
+void MoveAlgo::invalidatePhysics() {
+  setInWater (false);
+  setAsSwim  (false);
+  setAsDive  (false);
+  setAsJumpup(false);
+  setAsSlide (false);
+  setAsClimb (false);
+  setInAir   (false);
   }
 
 bool MoveAlgo::isFaling() const {
@@ -830,7 +850,6 @@ void MoveAlgo::setAsDive(bool f) {
   if(f==isDive())
     return;
   if(f) {
-    npc.setDirectionY(-40);
     diveStart = npc.world().tickCount();
     npc.setWalkMode(npc.walkMode() | WalkBit::WM_Dive);
     } else {
--- a/game/game/playercontrol.cpp
+++ b/game/game/playercontrol.cpp
@@ -431,6 +431,7 @@ void PlayerControl::marvinF8(uint64_t dt) {
   pl.quitIneraction();
   pl.setAnim(AnimationSolver::Idle);
   pl.clearAiQueue();
+  pl.invalidatePhysics();
 
   if(auto c = Gothic::inst().camera())
     c->reset();
@@ -441,18 +442,20 @@ void PlayerControl::marvinK(uint64_t dt) {
   if (w == nullptr || w->player() == nullptr)
     return;
 
-  auto& pl = *w->player();
-  auto  pos = pl.position();
-  float rot = pl.rotationRad();
-  float s = std::sin(rot), c = std::cos(rot);
+  auto& pl   = *w->player();
+  auto  pos  = pl.position();
+  float rot  = pl.rotationRad();
+  float rotY = pl.isDive() ? pl.rotationYRad() : 0;
+  float s    = std::sin(rot), c = std::cos(rot);
 
-  Tempest::Vec3 dp(s, 0.0f, -c);
+  Tempest::Vec3 dp(s,rotY,-c);
   pos += dp * 6000 * float(dt) / 1000.f;
 
   pl.clearState(false);
   pl.setPosition(pos);
   pl.clearSpeed();
   pl.quitIneraction();
+  pl.invalidatePhysics();
   // pl.setAnim(AnimationSolver::Idle); // Original G2 behaviour: K doesn't stop running
   }
 
--- a/game/game/movealgo.h
+++ b/game/game/movealgo.h
@@ -53,6 +53,7 @@ class MoveAlgo final {
 
     bool    startClimb(JumpStatus ani);
     void    startDive();
+    void    invalidatePhysics();
 
     bool    isFaling()  const;
     bool    isSlide()   const;
--- a/game/marvin.cpp
+++ b/game/marvin.cpp
@@ -284,6 +284,7 @@ bool Marvin::exec(std::string_view v) {
           return false;
         }
       player->setPosition(float(c[0]),float(c[1]),float(c[2]));
+      player->invalidatePhysics();
       player->updateTransform();
       Gothic::inst().camera()->reset(player);
       return true;
@@ -297,6 +298,7 @@ bool Marvin::exec(std::string_view v) {
       if(wpoint==nullptr)
         return false;
       player->setPosition(wpoint->x,wpoint->y,wpoint->z);
+      player->invalidatePhysics();
       player->updateTransform();
       Gothic::inst().camera()->reset(player);
       return true;
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1021,6 +1021,10 @@ bool Npc::isInAir() const {
   return mvAlgo.isInAir();
   }
 
+void Npc::invalidatePhysics() {
+  mvAlgo.invalidatePhysics();
+  }
+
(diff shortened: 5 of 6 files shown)


game/game/playercontrol.cpp
```diff
@@ -386,11 +386,21 @@ void PlayerControl::marvinF8(uint64_t dt) {
...
   pl.changeAttribute(ATR_HITPOINTS,pl.attribute(ATR_HITPOINTSMAX),false);
   pl.changeAttribute(ATR_MANA,     pl.attribute(ATR_MANAMAX),     false);
   pl.clearState(false);
+
+  if (pl.isSwim()) {
+    auto waterRay = pl.world().physic()->waterRay(pos);
+    auto res      = pl.world().physic()->ray(pos,{pos.x,20000,pos.z});
+    auto gl       = pl.guild();
+    float chest   = float(pl.world().script().guildVal().water_depth_chest[gl]);
+    if ((waterRay.hasCol && pos.y+chest>waterRay.wdepth) || (!res.hasCol && !waterRay.hasCol))
```
Try, line 400: What this checks for? Why straight overriding swim/dive flag is not enough? Why 20000?

game/game/playercontrol.cpp
```diff
@@ -405,15 +415,21 @@ void PlayerControl::marvinK(uint64_t dt) {
...
   float s = std::sin(rot), c = std::cos(rot);
-
   Tempest::Vec3 dp(s, 0.0f, -c);
-  pos += dp * 6000 * float(dt) / 1000.f;
+  dp *= 6000 * float(dt) / 1000.f;
 
   pl.clearState(false);
+  if (pl.isSwim()) {
+    auto waterRay = pl.world().physic()->waterRay(pos);
+    auto res      = pl.world().physic()->ray(pos,{pos.x,20000,pos.z});
```
Try, line 428: does marvin+K behaves the same in original?

game/game/playercontrol.cpp
```diff
@@ -405,15 +415,21 @@ void PlayerControl::marvinK(uint64_t dt) {
...
+    if (!res.hasCol && !waterRay.hasCol)
+      pl.marvinJump();
```
Try, line 430: Introducing something like `Npc::invalidatePhysics` and removing ray-related logic from PlayerControl feels more natural in such case

game/game/playercontrol.cpp
thokkat: Overriding ports you straight to the surface, while vanilla allows underwater use. 20000 is `DynamicWorld::worldHeight`. I made it a constant now as well.

game/game/playercontrol.cpp
thokkat: Vanilla marvinK in diving state considers `rotationYRad` so it's like a generalized marvinF8. Whenever there's an obstacle in the path the game tries to port the character upwards. This behavior is not consistent, sometimes nothing happens. I replicated basically the same behavior in the new commit, because sometimes no collision is detected.

thokkat:
Moved everything to `movealgo.cpp` as discussed and replicated marvinK behavior.

game/game/movealgo.cpp
```diff
@@ -709,27 +709,31 @@ void MoveAlgo::startDive() {
...
+  bool  validW        = false;
+  waterRay(pos,&validW);
+
+  if(pathRay.hasCol) {
+    if (skyRay.hasCol)
+      npc.setPosition(skyRay.v+Tempest::Vec3(0,5*fallThreshold,0));
     setAsSwim(false);
     setAsDive(false);
     setInAir (true);
     return;
```
thokkat, line 726: `if(pathRay.hasCol)` is there to copy vanilla behavior and move the character upwards. `5*fallThreshold` is an arbitrary constant added to prevent the character getting stuck.

game/game/movealgo.cpp
```diff
@@ -709,27 +709,31 @@ void MoveAlgo::startDive() {
...
-    if (sky.hasCol)
-      npc.setPosition(sky.v+Tempest::Vec3(0,5*fallThreshold,0));
+  if (!validW) {
+    auto groundRay = npc.world().physic()->ray(pos,Tempest::Vec3(pos.x,-worldHeight,pos.z));
+    if (skyRay.hasCol && groundRay.hasCol) {
+      bool validWG = false;
+      waterRay({groundRay.v.x,groundRay.v.y+1.f,groundRay.v.z},&validWG);
+      if (!validWG)
+        return;
+      }
```
thokkat, line 736: This is to work around the loopholes of water detection. We know `validW` is wrongly false only if there is `ground->water->ceiling`. If `validWG` is true we know the character is in air above water and it's save to leave swim state.

game/game/playercontrol.cpp
```diff
@@ -395,6 +396,8 @@ void PlayerControl::marvinF8(uint64_t dt) {
   pl.clearSpeed();
   pl.quitIneraction();
   pl.setAnim(AnimationSolver::Idle);
+  if (pl.isSwim())
+    pl.invalidatePhysics(dp);
```
Try, line 400: Why invalidation is needed only in `swim` case?

game/game/movealgo.cpp
```diff
@@ -708,6 +708,38 @@ void MoveAlgo::startDive() {
...
+    if (skyRay.hasCol)
+      npc.setPosition(skyRay.v+Tempest::Vec3(0,5*fallThreshold,0));
+    setAsSwim(false);
+    setAsDive(false);
+    setInAir (true);
+    return;
+    }
+
+  if (!validW) {
+    auto groundRay = npc.world().physic()->ray(pos,Tempest::Vec3(pos.x,-worldHeight,pos.z));
```
Try, line 730: Why not `landRay`? Also why sometimes you reaching to `physics()` directly and sometimes, thru cache, like with `waterRay`?

game/game/playercontrol.cpp
thokkat: Because if `isDive()` is true `isSwim()` is true as well.

game/game/movealgo.cpp
thokkat: This is for `K`. `F8` resets the character to the previous position if moving into a wall, but this doesn't feel natural.  `K` probably happens if non-water state is detected in vanilla. In this case swim conditions are invalidated.

thokkat:
> This look like a collision detection to me, for F8 and K. K has no collision in vanilla; F8 has, but it's bad for debugging.

I thought this upwards teleporting happens at collision time. At least it's not consistent with visible water.  So you can be in water and still being moved upwards. I rather would ignore this behavior and just invalidate. What's your opinion ? 

> I don't understand this arrow notation. What you kid of "loopholes" anyway - water work correct in game, isn't it?

For normal play yes where the only entry and exit point is the surface. But in marvin you can theoretically invalidate any time in dive state and like i said water detection fails in some cases. `ground->water->ceiling` means when you have ground, water above and then a ceiling like in an underwater cave, so there's no direct contact between air and water.  

Of course the optimal solution would  be to make `waterRay` perfect. If the `false` is changed to `true`

https://github.com/Try/OpenGothic/blob/220f046f83a5dab7f649fa4874b9a1f3df4f11c8/game/physics/dynamicworld.cpp#L511-L516

underwater cave detection works but then the problem is just moved somewhere else. I figured it's better to catch it  during invalidation  because it can't be handled in `tickSwim` anyway when using marvin commands.

game/game/playercontrol.cpp
Try: I mean what is the point of invalidating in not all cases? Is there any issues?

Try:
> underwater cave detection works but then the problem is just moved somewhere else. 

There is no way to have correct state-less detection of in-cave swim/no-swim cases. In Lurker cave(Jarkendar) there is no water layer at all for large cave section, but player has to swim thru. 

> i said water detection fails in some cases.

What cases? Water ray is pretty simple function: it shoot ray from player upwards, and returns `true` if water is first to hit, and `false` if landscape was first to hit or if there is no hits at all.

game/physics/collisionworld.cpp
```diff
@@ -336,6 +336,10 @@ void CollisionWorld::setBBox(const btVector3& min, const btVector3& max) {
   bbox[1] = max;
   }
 
+float CollisionWorld::worldHeight() const {
```
Try, line 339: this function doesn't do, what it claims. In case, if `bbox[1].y` is zero - all physics will break

Try:
Sorry, but I'm losing tracks of this PR. This one suppose to be quite trivial - just invalidate player entire state with one or two rays - that's should be all.

thokkat:
I made it a simple vanilla like water check.

game/game/movealgo.cpp
```diff
@@ -708,6 +708,20 @@ void MoveAlgo::startDive() {
     }
   }
 
+void MoveAlgo::invalidatePhysics() {
```
Try, line 711: This function doesn't do what it claims to do.

thokkat:
Corrected `Air state`

thokkat:
@Try  I changed it to only set air state to true. Now it can be handled in swim routine directly. This is also needed on level ground or water surface because the uplift of F8 is not high enough to trigger air state.

Try:
Hi, @thokkat !

Noted, PR - planning to test changes later at weekend

game/game/playercontrol.cpp
```diff
@@ -430,6 +430,7 @@ void PlayerControl::marvinF8(uint64_t dt) {
   pl.clearSpeed();
   pl.quitIneraction();
   pl.setAnim(AnimationSolver::Idle);
+  pl.applyAirPhysics();
```
Try, line 433: Looking into this change: I'm not sure about this design choose. So far all physics state bits were encapsulated in `MoveAlgo`, and been read-only for rest of code. Why `invalidatePhysics`-kind approach was discarded?

game/game/movealgo.cpp
```diff
@@ -289,6 +289,16 @@ void MoveAlgo::tickSwim(uint64_t dt) {
       }
     }
 
+  if(isInAir()) {
+    if(pos.y+chest>water) {
```
Try, line 293: This state can be reached only by calling `applyAirPhysics`?

game/game/movealgo.cpp
```diff
@@ -708,6 +718,10 @@ void MoveAlgo::startDive() {
     }
   }
 
+void MoveAlgo::applyAirPhysics() {
+  setInAir(true);
```
Try, line 722: Assuming that condition `isInAir == true && isSwim == true` can be caused only by this call. Is this an complicated way to call `setAsSwim (false);` ?

Try:
> The idea was if setInAir(true) was needed anyway than use it as a signal that the move was triggered by F8 so invalidation of swimming conditions is safe.

I'm sorry, but this is a very good idea - this is more workaround in it's nature, with no reason for that. Again I see no reason not to leverage physics engine.

thokkat:
I restored the previous state

game/game/movealgo.cpp
Try: Comment is still relevant:
if someone will attempt to use this function, `mvAlgo.invalidatePhysics()` expecting physics state invalidation:
1. `Flags::InWater` is recalculated at least.
2. `Flags::InAir` set to false, without care, if npc is actually in air or not
3. same for all other flags
4. cached, for ray-cast, state is still cached

thokkat:
>Flags::InWater is recalculated at least.

Problem was i had no other idea how to keep diving state if it's set to false but character is still diving. 

If not on ground after Invalidation `air` is set to true. Then `tickGravity` checks for water presence and a`tryMove` causes an uplift.  I added an earlier check for water to allow the usage of F8/K while diving.


> `Fags::InAir` set to false, without care, if npc is actually in air or not

Isn't this the purpose of the function? Reset to normal walking state and let movealgo figure out if something else is actually true.

Try:
Original issue is no longer reproducible after 910923eb

PR #379, opened 2022-12-28
Update phoenix and fix resulting compilation errors

lmichaelis:
Due to an issue in how phoenix' CMake script was set up, compilation flags were propagated across targets incorrectly. This was recently fixed and also caused some compilation errors in OpenGothic which this commit addresses.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -335,8 +335,8 @@ void GameScript::initCommon() {
     size_t tbSz=size_t(std::sqrt(tableSz->get_int()));
     for(size_t i=0;i<tbSz;++i)
       for(size_t r=0;r<tbSz;++r) {
-        gilAttitudes[i*gilCount+r]=guilds->get_int(i*tbSz+r);
-        gilAttitudes[r*gilCount+i]=guilds->get_int(r*tbSz+i);
+        gilAttitudes[i*gilCount+r]=guilds->get_int(uint8_t(i*tbSz+r));
+        gilAttitudes[r*gilCount+i]=guilds->get_int(uint8_t(r*tbSz+i));
         }
   }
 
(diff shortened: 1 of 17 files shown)


game/game/gamescript.cpp
```diff
@@ -335,8 +335,8 @@ void GameScript::initCommon() {
     size_t tbSz=size_t(std::sqrt(tableSz->get_int()));
     for(size_t i=0;i<tbSz;++i)
       for(size_t r=0;r<tbSz;++r) {
-        gilAttitudes[i*gilCount+r]=guilds->get_int(i*tbSz+r);
-        gilAttitudes[r*gilCount+i]=guilds->get_int(r*tbSz+i);
+        gilAttitudes[i*gilCount+r]=guilds->get_int(uint8_t(i*tbSz+r));
```
Try, line 338: Hm, can `get_*` functions take a uint32_t/size_t as a index instead? That would math original script close, here uint8_t is not a thing, also usage of 8-bit for array index doesn't fill right.

lmichaelis: Seems plausible. I was going to say that I just used the same type as the compiled script bit it uses 12 bits, not 8.

https://github.com/lmichaelis/phoenix/blob/84026ed39d29dfbc0920b158041c52c68cb89104/source/script.cc#L229-L234

I'll consider using `size_t` but to stick closer to the actual values used, I might choose a `uint16_t` instead. I'll let you know tomorrow :)

lmichaelis: Ah I remember now why I chose a `uint8_t`! It's because instructions only contain 8-bits for array indexing. I've decided to accept a `size_t` instead though since that generally seems like it's the right thing to do.

Patch coming soon.

Try (approved):
LGTM, except sending `size_t` to gpu - made inline fix, hopefully it work

PR #383, opened 2023-01-15
Implemented focus handling

### LINKED ISSUE #374: alt+tab traps mouse
alt+tab focuses another window and overlays it on top of OpenGothic, a normal behavior for Windows. But moving the mouse outside the focused window always moves the mouse to the center of the screen.
It seems that OpenGothic processes mouse movements even when it is not in focus. This makes it harder to do anything else while playing OpenGothic.

I would suggest ignoring mouse input while OpenGothic is out of focus.

Reveares:
Closes #374 and centers mouse at focus event to avoid unwanted mouse movement.


Try:
Look good to me, thanks!

PR #385, opened 2023-01-17
Marvin console improvements

thokkat:
A bit more convenience for marvin console. 

The cursor is now moveable, the last input is preserved while moving through history and the delete key is usable.

### DIFF
--- a/game/ui/consolewidget.cpp
+++ b/game/ui/consolewidget.cpp
@@ -82,7 +82,6 @@ void ConsoleWidget::close() {
 
   ov->takeWidget(this);
   delete ov;
-  log.back().clear();
   }
 
 int ConsoleWidget::exec() {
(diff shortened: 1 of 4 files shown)


game/ui/consolewidget.cpp
```diff
@@ -137,36 +137,64 @@ void ConsoleWidget::keyDownEvent(KeyEvent& e) {
...
-    if(log.back().size()>0)
-      log.back().pop_back();
+    if(log.back().size()>0 && cursPos>0) {
+      log.back().erase(--cursPos,1);
+      }
+    return;
+    }
+  if(e.key==Event::K_Delete) {
+    if(log.back().size()>cursPos) {
+        log.back().erase(cursPos,1);
```
Try, line 191: nitpick: tabulation

game/ui/consolewidget.cpp
```diff
@@ -137,36 +137,64 @@ void ConsoleWidget::keyDownEvent(KeyEvent& e) {
...
+    cursPos = 0;
+    currCmd = "";
+    histPos = size_t(-1);
     return;
     }
 
   if(e.key==Event::K_Back) {
-    if(log.back().size()>0)
-      log.back().pop_back();
+    if(log.back().size()>0 && cursPos>0) {
```
Try, line 184: Maybe something like 
`if(0<curssPos && curssPos<=log.back().size())`
would be more clean way to express this condition?

thokkat:
Addressed the changes.

Try:
Merged, thanks!

PR #397, opened 2023-02-01
Implemented AI_LOOKAT

CReimer:
This is my first try at actually contributing here. So please be gentle.

Some NPCs in Gothic 1 don't look in the direction they're supposed to. I figured AI_LOOKAT could be the problem.
So I implemented it as best as I could, and the NPCs still don't look in the right direction.

It turns out AI_LOOKAT is only used in the new camp (Neues Lager in German) to make an NPC look at the big pile of ore there. (Waypoint FP_OREPILE_CENTER). I haven't figured out which NPCs but if that information is needed, I might now know how to get it.

I made AI_LOOKAT mutually exclusive with AI_LOOKATNPC as I figured an NPC can't really look at two points at the same time. So whichever function is called last wins.

Your code style is a bit unusual. I hope I got it right.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -257,7 +257,7 @@ void Npc::load(Serialize &fin, size_t id) {
   loadAiState(fin);
 
   fin.read(currentInteract,currentOther,currentVictum);
-  if(fin.version()>42)
+  if(fin.version()>=42)
     fin.read(currentLookAt);
   fin.read(currentLookAtNpc,currentTarget,nearestEnemy);
 
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -257,7 +257,7 @@ void Npc::load(Serialize &fin, size_t id) {
   loadAiState(fin);
 
   fin.read(currentInteract,currentOther,currentVictum);
-  fin.read(currentLookAt,currentTarget,nearestEnemy);
+  fin.read(currentLookAtNpc,currentTarget,nearestEnemy);
 
   go2.load(fin);
   fin.read(currentFp,currentFpLock);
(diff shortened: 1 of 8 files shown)


game/world/objects/npc.cpp
```diff
@@ -257,7 +257,7 @@ void Npc::load(Serialize &fin, size_t id) {
   loadAiState(fin);
 
   fin.read(currentInteract,currentOther,currentVictum);
-  fin.read(currentLookAt,currentTarget,nearestEnemy);
+  fin.read(currentLookAtNpc,currentTarget,nearestEnemy);
```
Try, line 260: 1. `write` code wasn't updated to mirror read
2. Waypoint `currentLookAt` is not serialized 

For proper serialization, you need to bump save file version in `Serialize::Version` and add check for `fin.version()` before reading new data.

game/world/objects/npc.cpp
```diff
@@ -1988,7 +1995,8 @@ void Npc::tick(uint64_t dt) {
     }
 
   if(!isDown()) {
-    implLookAt(dt);
+      implLookAtNpc(dt);
```
Try, line 1998: Nit: tabs

Try (commented):
Hi, @CReimer !

Thanks for PR. Looks good, but need to fix serialization (see comments)

game/world/objects/npc.cpp
CReimer: I'm not exactly sure how the serialization works. But I think it's fixed now

game/world/objects/npc.cpp
CReimer: I hope this is now right, too.

To be fair. I just spend about an hour trying to set your code styling in CLion.
It's just impossible. To many edge-cases and inconsistencies.

CReimer:
If everything is okay now. I'd like to squash the first new commit in my original fist commit and the same for the second.

Try:
Hi, @CReimer !

There were few issues with serialization, so I've changed them inline, in merge view.
Squashed and merged, thanks!

PR #402, opened 2023-02-06
Make Mac Command keys selectable

### LINKED ISSUE #395: [macOS] keyboard issues
Hi, just discovered this project - awesome job!

Trying to play I discovered a couple of problems with the controls. I'm unable to bind most of the lowest row of my keyboard. Specifically on my MBP keyboard, these are control, left option (alt), left command, right command, right option (alt). Also neither Shift key works. The main menu just does nothing when I try to bind them. Only the space bar works. I'm assuming this is some kind of mapping issue, but I have no idea what to do about it. I'm using an US keyboard layout.

I also tried equipping a weapon (the heavy branch at the very beginning), but hitting action in the inventory drops the item instead of using it. This can be worked around by using the numbers to assign slots, but I imagine that won't work for consumables.

Finally, swinging the weapon left or right and then holding down either left or right when not focused on an enemy will cause the character to keep turning in that direction, same as if action was not held down. In "official" Gothic I believe the behavior is that as long as action is being held, the character won't turn.

sjavora:
Resolves #395.

### DIFF
--- a/game/utils/keycodec.cpp
+++ b/game/utils/keycodec.cpp
@@ -57,10 +57,12 @@ std::initializer_list<KeyCodec::K_Key> KeyCodec::keys = {
   {Tempest::Event::K_LControl, 0x1d00},
   {Tempest::Event::K_LShift,   0x2a00},
   {Tempest::Event::K_LAlt,     0x3800},
+  {Tempest::Event::K_LCommand, 0x4800},
   // Right
   {Tempest::Event::K_RControl, 0x9d00},
   {Tempest::Event::K_RShift,   0x3600},
   {Tempest::Event::K_RAlt,     0xb800},
+  {Tempest::Event::K_RCommand, 0x4801},
   };
 
 std::initializer_list<KeyCodec::M_Key> KeyCodec::mkeys = {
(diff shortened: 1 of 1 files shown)


game/utils/keycodec.cpp
```diff
@@ -57,10 +57,12 @@ std::initializer_list<KeyCodec::K_Key> KeyCodec::keys = {
   {Tempest::Event::K_LControl, 0x1d00},
   {Tempest::Event::K_LShift,   0x2a00},
   {Tempest::Event::K_LAlt,     0x3800},
+  {Tempest::Event::K_LCommand, 0x4800},
```
Try, line 60: How `0x4800` and `0x4801` were selected?
While I mention before, that can be anything really; curious if you can check what vanilla(Wineskin) writes in *.ini for those keys

sjavora: Wineskin has checkboxes for remapping these that I think I posted before:
Screenshot 2023-02-06 at 21 59 34

With the checkboxes **unchecked**:
- left command is interpreted as left alt: `3800`
- right command is interpreted as right alt: `b800`
- left ctrl is interpreted as left ctrl: `1d00`
- my keyboard doesn't have a right ctrl, but I'd expect it to work the same as with left control (so right ctrl - `9d00`)
- option/alt keys do nothing

With the checkboxes **checked**:
- left command is interpreted as left ctrl: `1d00`
- right command is interpreted as right ctrl: `9d00`
- left ctrl is interpreted as left ctrl: `1d00`
- left alt is interpreted as left alt: `3800`
- right alt is interpreted as right alt: `b800`

So with the checkboxes checked, ctrl and command do the same thing. I don't think there is a "cannonical" code for either command keys. So I picked the  values at random really - tried to find something that wasn't occupied, but also wanted the two to be close to each other. If you prefer some other codes, feel free to edit them.

Try:
Merged, thanks!

Issue #215, opened 2021-11-21
Finalize 'Marvin' mode


Try:
At this point only a few `marvin` commands are implemented - need to finalize it.

fhanrath:
I'm working on the marvin 'o' [command](https://github.com/fhanrath/OpenGothic/commit/0f9a10f407e28f3bbdd47425d8cea0a0bb6140a4) right now.
After I take control of an npc, the npc is still controlled by ai. Do you have a suggestion how I'd be able to pause the ai until the npc is no longer controlled by me?

Try:
We cannot disable AI completely, event for player (dialogs are AI driven). Probably the best solution is to clear `aiState` and `aiQueue`, `aiQueueOverlay`.
And after, once player switches to another `npc`, AI state should be restarted (see `AI_ContinueRoutine`)

fhanrath:
Thank you!

fhanrath:
Missing features:
  - [F2] open console (still needed/ do we have a console?)
  - [F3] toggle windowed mode
  - [F4] normal mode  
  - [F6] move the camera freely
  - [F7] skip ahead
  - [H]  reduce health points
  - [Z]  run around
  - [G]  display information about the own character
  - [T]  display information about the currently selected NPC
  - [Alt + P] Display current position

Try:
Hi, @fhanrath !

Thanks for highlighting the missing features.

About F2/F3 - are you sure? Console should be in game already (as well as some commands). Same with F3

fhanrath:
Hey @Try,

No, I am not sure about F2 and F3. I just had a look at a list of available commands and the commands implemented in [playercontrol.cpp](https://github.com/Try/OpenGothic/blob/master/game/game/playercontrol.cpp).

I'll mark them as done.

PR #404, opened 2023-02-08
More Marvin commands

thokkat:
This PR implements marvin keys `F4-F7` and the console command `toggle time` https://github.com/Try/OpenGothic/issues/215.

F4 - Reset camera to normal mode.
F5 - Freeze camera. Camera stays in place and can't be moved. Character  can move freely.
F6 - Free camera. Allows to fly the camera through the world. Character is frozen.
F7 - Like normal play but camera is shifted to a position reached by F6. Camera moves with the character.

F4-F6 are implemented  the same as vanilla. There `F7` aligns the camera to the players head-to-foot axis. I haven't looked into that and kept the usual y-axis, but can try if it's desirable. 

I encountered some hiccups when the camera changes to a different mode. I opted to not stop marvin cam in those cases for now, there's always `F4` to reset.

Comparison F7 PR/vanilla

https://user-images.githubusercontent.com/100468436/217689362-e630d643-764a-4393-a6a0-3b304f9e7dcd.mp4

https://user-images.githubusercontent.com/100468436/217688735-b3a97cdc-b140-4f9f-956a-4d3c3a5a3ec7.mp4


Try:
Hi, @thokkat , thanks for PR!

Can you please give more detail explanation on each command? Expected vanilla behavior and implemented one(if differs)

thokkat:
Edited start post.

game/camera.cpp
```diff
@@ -145,6 +150,45 @@ void Camera::setLookBack(bool lb) {
...
+      return;
+    auto offset    = dst.target - pl->cameraBone();
+    float k        = float(M_PI/180.0);
+    float s        = std::sin(dst.spin.y*k), c=std::cos(dst.spin.y*k);
+    cameraOffset.x = -offset.x * c - offset.z * s;
+    cameraOffset.z = offset.x * s - offset.z * c;
+    cameraOffset.y = offset.y;
+    }
+  else
+    dst.target = origin;
```
Try, line 184: nitpick: inconsistent brackets. Should be
```
if(...) {
  foo();
  } else {
  boo();
  }
```

Try (commented):
I've made a quick only look so far, will do testing, once I'm done with npc swimming code, that I'm currently working on.

game/camera.h
```diff
@@ -31,6 +31,13 @@ class Camera final {
       Dive
       };
 
+    enum MarvinMode {
+      M_Normal,
+      M_Freeze,
+      M_Free,
+      M_Fixed,
```
Try, line 38: M_Fixed
> Like normal play but camera is shifted to a position reached by F6. Camera moves with the character.

Name is misleading. Should it be, for example, "pinned" ?

game/camera.h
```diff
@@ -117,6 +127,7 @@ class Camera final {
     bool                  lbEnable      = false;
     bool                  inertiaTarget = true;
     Mode                  camMod        = Normal;
+    MarvinMode            camMarvinMod  = M_Normal;
```
Try, line 130: Are `Mode` and  `MarvinMode` complementary or mutably exclusive?
If exclusive - it make sense to have them into one enum.

game/world/worldobjects.cpp
```diff
@@ -161,9 +161,15 @@ void WorldObjects::tick(uint64_t dt, uint64_t dtPlayer) {
 
   for(size_t i=0; i<npcArr.size(); ++i) {
     auto& npc = *npcArr[i];
-    if(npc.isPlayer())
-      npc.tick(dtPlayer); else
+    if(npc.isPlayer()) {
+      if(Gothic::inst().camera() && Gothic::inst().camera()->isMarvinMode(Camera::M_Free)) {
+        return;
```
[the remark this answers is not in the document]
Try: Also why player is not receiving ticks, when in free-cam mode?

game/mainwindow.cpp
```diff
@@ -687,16 +704,32 @@ void MainWindow::onMarvinKey() {
...
       break;
     case Event::K_F5:
-      Gothic::inst().quickSave();
-      // note: camera lock
+      if(Gothic::inst().isMarvinEnabled()) {
+        if(auto camera = Gothic::inst().camera()) {
+          camera->setMarvinMode(Camera::M_Freeze);
+          }
+        } else {
+        Gothic::inst().quickSave();
```
Try, line 718: While this is correct, not having `F5` in debug play would be painfully annoying. Maybe remove `camera-freeze`, from debug build?

game/camera.cpp
```diff
@@ -122,6 +125,32 @@ void Camera::setMode(Camera::Mode m) {
...
+void Camera::setMarvinMode(Camera::MarvinMode m) {
+  if(camMarvinMod==m)
+    return;
+  auto pl = Gothic::inst().player();
+  if(pl==nullptr)
+    return;
+  if(m==M_Free)
+    pl->setAnim(AnimationSolver::NoAnim);
+  if(m==M_Fixed) {
+    auto offset    = dst.target - pl->cameraBone();
```
Try, line 137: This math odd. 
Is it another way to write `inverse(pl->cameraBone()) * dst.target` ?

game/camera.cpp
```diff
@@ -122,6 +125,32 @@ void Camera::setMode(Camera::Mode m) {
...
+    float k        = float(M_PI/180.0);
+    float s        = std::sin(dst.spin.y*k), c=std::cos(dst.spin.y*k);
+    cameraOffset.x = -offset.x * c - offset.z * s;
+    cameraOffset.z = offset.x * s - offset.z * c;
+    cameraOffset.y = offset.y;
+    }
+  else if(camMarvinMod==M_Fixed) {
```
Try, line 144: This line is misleading.
2 previous condition in a row were about `m`(new mode), suggesting switch-case like workflow, but here it's suddenly `camMarvinMod`.

game/camera.cpp
thokkat: This is supposed to calculate the vector from player to camera position. Is there a better way?

game/world/worldobjects.cpp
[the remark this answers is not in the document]
thokkat: Player can't move, doesn't drop if in air, receives no damage if diving bar is empty, `aigoto` only starts if `free cam` mode is left and current animation is frozen. Player can be damaged by monsters and npc though but death animation behaves like `aigoto`. Maybe it's best to just activate `godmode` here. 

So basically all what's done in player-tick is skipped/ignored.

game/camera.h
thokkat: This is a bit unclear to me. If a dialog is started in F5/F7 mode dialog cam takes over and afterwards F5/F7 takes over again. So there has to be some memory which cam was active before.
 
I moved the marvinMode enum to `mainwindow.cpp` and use `solveCameraMode` routine  to return the respective camera. So only one enum in `camera.cpp` with marvin cams included.

game/world/worldobjects.cpp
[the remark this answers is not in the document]
Try: Tested vanilla today:
In free cam move playable npc doesn't receive animation ticks for skeletal animation, an no physics ticks. It can receive damage and perceptions(magic damage at least).
Also main menu not working in free-camera O-o

Anyway `return;` is definitely wrong here.

game/camera.h
Try: > f a dialog is started in F5/F7 mode dialog cam takes over and afterwards F5/F7 takes over 

I didn't know that - resolved

game/camera.cpp
thokkat: Isn't it what I'm currently doing?  Code below would lead to same offset vector.
```
   Vec3 offset(0,0,1);
   Matrix4x4 m;
   m.identity();
   m.translate(pl->cameraBone(true));
   m.inverse();
   m.translate(dst.target);
   m.project(offset);  
   cameraOffset = offset;
```

Try: I mean it make sense to change it, and fetch `cameraBone` from npc as matrix, with tilt and rotation.

thokkat: Edit: Looks like `offsetAng` could do the tilt.

ezamelczyk:
I can't really pinpoint why it's happening but when I build this branch on m1 mac I have an issue. Menus are working but in the game everything's funky. I preformed a clean build and it did not help. Main branch does not have this issue. Screenshot attached.

Zrzut ekranu 2023-02-28 o 11 54 07

thokkat:
@ezamelczyk 
Everything this PR does is disabled by default. Once activated only the camera position should change. Maybe it's because this pr was some weeks behind master. But master is now merged in so you could give it another try.

@Try
 I addressed the requested changes. I don't know if there's a better way to prevent animation ticks. It seems the animation is reset  instead of continued if e.g `F6->F7`, which leads to problems with mobsi states. As a workaround i prevented the use of F6 in these cases.

Also didn't manage to get tilt in. Everything i did also changed the rotation axis. I'd like to leave that for a follow pr, except you know right away how to do it. Although i like it how it is, just a change of camera perspective.

ezamelczyk:
Yup. Seems to be working correctly with updates from main.

Try:
Checkout with github no longer working :(
Anyway, code look ok - will merge it as-is now, and fix nitpicks in post

PR #415, opened 2023-02-21
Porting To FreeBSD: Bash is not always in /bin

raphaelahrens:
FreeBSD and other BSDs store ported software under /usr/local This includes bash.

This change uses /usr/bin/env to find the location of bash

### DIFF
--- a/scripts/Gothic2Notr.sh
+++ b/scripts/Gothic2Notr.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 export LD_LIBRARY_PATH="$DIR:$LD_LIBRARY_PATH"
 export DYLD_LIBRARY_PATH="$DIR:$DYLD_LIBRARY_PATH"


Try:
Hi, @raphaelahrens, thanks for PR!
Merged.

aklv:
Perhaps it's better to switch to sh syntax instead? The changes are trivial and  /bin/sh exist in all Linux distro and other unix-like OS.
[patch.txt](https://github.com/Try/OpenGothic/files/10798401/patch.txt)

Try:
Hi, @aklv , interesting suggestion.

Generally I'm not Linux guy, so don't really know what is best for you. If no objection - can change script

aklv:
> There you can replace `${BASH_SOURCE[0]}` with `$0` to be POSIX compliant.

Yes, please see attached patch.

> But since the script is only useful for someone who compiles the game manual, I don't see the need since POSIX makes things complicated and everthing is an edge case.

Please don't equate the POSIX in general and POSIX sh syntax. Every major sh based shell have sh comaptibility mode. Every Linux distro has /bin/sh. In some it's a symlink from bash or zsh, some use standalone ash or sh. So there is no complication or edge cases.
 
> For the FreeBSD you would build a port which installs the files and then you will not need the script at all. Since the `*.so` libs will be in `/usr/local/lib`. A port has also the advantage that it can be commited into the the [ports tree](https://github.com/freebsd/freebsd-ports).

It's separate, but valid question. I think amdmi3@ works (or worked) on this.

Issue #409, opened 2023-02-12
[GI] small assortment of interesting bugs


dreimer1986:
Yes, I still know that GI is not officially supported (yet). I just throw these in here for a maybe fix in the future and hopefully to help some GII corner cases, too.

![image_2023-02-12_13-13-40](https://user-images.githubusercontent.com/29841368/218311172-30a9b22b-f40f-4cc0-918c-e7b490526118.png)

Looks like the "Mighty Alien Dwarf" hid his UFO in the old camp dungeon? Or did they find Uranium down there? No serious, this bright green light is not there normally.

![image_2023-02-12_13-19-18](https://user-images.githubusercontent.com/29841368/218311179-7511afab-1cac-4dc2-8bb4-993c570c300b.png)

You forgot to implement the needed mountaineering equipment for poor Milten trying to get my letter to Corristo. The new camp really overdid it this time and stole both stairs!

![image](https://user-images.githubusercontent.com/29841368/218311411-92739285-989f-4b7b-a696-12d2624ebbb3.png)

I really hoped that the new added and formerly missing AI_LOOKAT fixes these guys, but nope. They should look at each other and chat. Here they look at a wall in silence and vain.

Will be updated wehenever I find other interesting stuff.

Try:
> Looks like the "Mighty Alien Dwarf" hid his UFO in the old camp dungeon? Or did they find Uranium down there?

This is center of coordinate system. Possibly startup script is trying to create objects(lightsources) on invalid(absent) waypoints. In this case OpenGothic defaults to zero.

> You forgot to implement the needed mountaineering equipment for poor Milten

Fixed, thanks.

Rest need to look into, probably script-api mismatch

dreimer1986:
```
OpenGothic v1.0 dev
GPU = NVIDIA GeForce RTX 2080
Depth format = Depth32F Shadow format = Depth16
[phoenix] world: parsing object [MeshAndBsp % 0 0]
[phoenix] bsp_tree: parsing chunk c000
[phoenix] bsp_tree: parsing chunk c010
[phoenix] bsp_tree: parsing chunk c040
[phoenix] bsp_tree: parsing chunk c045
[phoenix] bsp_tree: parsing chunk c050
[phoenix] bsp_tree: parsing chunk c0ff
[phoenix] world: parsing object [VobTree % 0 0]
[phoenix] world: parsing object [WayNet % 0 0]
[phoenix] world: parsing object [EndMarker % 0 0]
[...]

PR #422, opened 2023-03-07
Fix for G1 chests

thokkat:
If trying to open a locked chest there's a check if at least one lockpick is in inventory. The game checks only for the G2 name `ItKE_lockpick`. Now also the G1 name `itkelockpick` is considered.
https://github.com/Try/OpenGothic/issues/409

Gothic 1 also allows lockpicking without having to learn the respective Talent. Lockpick loss probability depends on talent skill rather than dexterity.

### DIFF
--- a/game/game/playercontrol.cpp
+++ b/game/game/playercontrol.cpp
@@ -905,8 +905,8 @@ void PlayerControl::processPickLock(Npc& pl, Interactive& inter, KeyCodec::Actio
 
   if(pickLockProgress<cmp.size() && std::toupper(cmp[pickLockProgress])!=ch) {
     pickLockProgress = 0;
-    const int32_t dex = pl.attribute(ATR_DEXTERITY);
-    if(dex<int32_t(script.rand(100)))  {
+    const int32_t dex = Gothic::inst().version().game==2 ? pl.attribute(ATR_DEXTERITY) : 100 - pl.talentValue(TALENT_PICKLOCK);
+    if(dex<=int32_t(script.rand(100)))  {
       script.invokePickLock(pl,0,1);
       pl.delItem(ItKE_lockpick,1);
       if(pl.inventory().itemCount(ItKE_lockpick)==0) {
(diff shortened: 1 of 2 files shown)


game/game/playercontrol.cpp
```diff
@@ -905,8 +905,8 @@ void PlayerControl::processPickLock(Npc& pl, Interactive& inter, KeyCodec::Actio
 
   if(pickLockProgress<cmp.size() && std::toupper(cmp[pickLockProgress])!=ch) {
     pickLockProgress = 0;
-    const int32_t dex = pl.attribute(ATR_DEXTERITY);
-    if(dex<int32_t(script.rand(100)))  {
+    const int32_t dex = Gothic::inst().version().game==2 ? pl.attribute(ATR_DEXTERITY) : 100 - pl.talentValue(TALENT_PICKLOCK);
+    if(dex<=int32_t(script.rand(100)))  {
```
Try, line 909: Why `<=`? Rand(100) returns number in [0..99] range. That means, if dexterity equals 99 player will always be able to to open locks. Is that intended?

game/game/playercontrol.cpp
```diff
@@ -905,8 +905,8 @@ void PlayerControl::processPickLock(Npc& pl, Interactive& inter, KeyCodec::Actio
 
   if(pickLockProgress<cmp.size() && std::toupper(cmp[pickLockProgress])!=ch) {
     pickLockProgress = 0;
-    const int32_t dex = pl.attribute(ATR_DEXTERITY);
-    if(dex<int32_t(script.rand(100)))  {
+    const int32_t dex = Gothic::inst().version().game==2 ? pl.attribute(ATR_DEXTERITY) : 100 - pl.talentValue(TALENT_PICKLOCK);
```
Try, line 908: Nitpick: please add brackets, for false branch, to avoid confusion

game/game/playercontrol.cpp
thokkat: This happens with `<`.  `99<=99` enters the delete branch and 100 is required to never loose a lockpick.

Try:
Merged, thanks!

game/game/playercontrol.cpp
Try: Please note those kind of changes explicitly in future PR's - this will make review process more simple

PR #424, opened 2023-03-08
Prevent ranged damage from hitting twice

thokkat:
After a npc is hit by a bolt it's stopped. In move routine there's a check if npc is moving into the shooting line. If true this causes a double hit because it happens before the bolt is deleted from `bullets` list.  This can be prevented by checking if bolt has the `Stopped` flag.

### DIFF
--- a/game/world/bullet.cpp
+++ b/game/world/bullet.cpp
@@ -130,7 +130,7 @@ void Bullet::onCollide(phoenix::material_group matId) {
   }
 
 void Bullet::onCollide(Npc& npc) {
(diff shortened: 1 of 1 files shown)


Try:
Hm, nice!
Just curious, how to reproduce case, when damage happen twice?

thokkat:
Actually i haven't found a way to reproduce this reliably. When i load a save for testing it happens like every second time at least once

Try:
I've added check into other `onCollide`, for sake of consistency

PR #434, opened 2023-03-29
Rework damage calculation

thokkat:
The hit chance for physical ranged damage is defined at three distance points from target with a linear scaling in between:

- 100% hit chance at `0`
- Bow/Crossbow skill value (Dexterity for G1) hit chance at `RANGED_CHANCE_MINDIST`
- 0% hit chance at `RANGED_CHANCE_MAXDIST`

 where `RANGED_CHANCE_MINDIST=1500` (`2000` for G1 based on some testing) and `RANGED_CHANCE_MAXDIST=4500`. These constans are defined in `AI_Constants.d` script file. I found a good explanation [here](https://strafkolonie-online.net/forum/board/thread/1895-info-erkl%C3%A4rung-der-berechnung-der-trefferchance-im-fernkampf/) (german, but chart and table are self-explaining).
Since distance is only known at hit time the hit chance calculation was moved to the damage calculation code. Critical hit chance calculation was added and the dexterity bonus to damage removed for G1.  

I also adjusted G1 sword damage to consider `talentValue` instead of the not present `hitChance` attribute.  In crit chance calculation`<` were changed to `<=`  to keep 0% hit chance if the skill value is 0 instead of 1%.

G1 has no minimal damage so don't consider `MinDamage`. Plus some spelling and code style rework.

### DIFF
--- a/game/game/damagecalculator.cpp
+++ b/game/game/damagecalculator.cpp
@@ -21,7 +21,7 @@ DamageCalculator::Val DamageCalculator::damageValue(Npc& src, Npc& other, const 
     ret = swordDamage(src,other);
     }
 
-  if(ret.hasHit && !ret.invinsible)
+  if(ret.hasHit && !ret.invincible && Gothic::inst().version().game==2)
     ret.value = std::max<int32_t>(ret.value,MinDamage);
   if(other.isImmortal() || (other.isPlayer() && Gothic::inst().isGodMode()))
     ret.value = 0;
@@ -40,9 +40,9 @@ DamageCalculator::Val DamageCalculator::damageFall(Npc& npc, float speed) {
   int32_t prot        = npc.protection(::PROT_FALL);
 
   Val ret;
-  ret.invinsible = (prot<0 || npc.isImmortal() || (npc.isPlayer() && Gothic::inst().isGodMode()));
+  ret.invincible = (prot<0 || npc.isImmortal() || (npc.isPlayer() && Gothic::inst().isGodMode()));
   ret.value      = int32_t(dmgPerMeter*(height-h0)/100.f - float(prot));
-  if(ret.value<=0 || ret.invinsible) {
+  if(ret.value<=0 || ret.invincible) {
     ret.value = 0;
     return ret;
     }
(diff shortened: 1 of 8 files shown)


game/game/damagecalculator.cpp
```diff
@@ -51,17 +51,44 @@ DamageCalculator::Val DamageCalculator::damageFall(Npc& npc, float speed) {
...
+    float critRange = g2 ? CritBowRange : (CritBowRange + 500);
+    float skill     = b.hitChanceVal();
+
+    if(dist<critRange)
+      hitCh = int((skill - 100.f) / critRange * dist + 100.f + 0.5f); else
+      hitCh = int(skill / (critRange - float(MaxBowRange)) * (dist - float(MaxBowRange)) + 0.5f);
+
+    noHit = (dist>float(MaxBowRange) || hitCh<=hitChance);
+
+    auto w = nsrc.inventory().activeWeapon();
```
Try, line 73: This is not reliable. `activeWeapon` can be anything, by the time bullet hits.
If you need to access `talentValue` - you can embed it into bullet, similar to hit chance

game/game/fightalgo.cpp
```diff
@@ -298,9 +298,8 @@ float FightAlgo::weaponRange(GameScript &owner, const Npc &npc) {
     case WeaponState::Fist:
       return float(gv.fight_range_fist[gl]);
     case WeaponState::Bow:
-      return float((MaxBowRange*npc.hitChanse(TALENT_BOW))/100);
     case WeaponState::CBow:
-      return float((MaxBowRange*npc.hitChanse(TALENT_CROSSBOW))/100);
+      return float(MaxBowRange);
```
Try, line 302: Why? Also, this change to AI is not in PR description.

game/game/damagecalculator.cpp
```diff
@@ -51,17 +51,44 @@ DamageCalculator::Val DamageCalculator::damageFall(Npc& npc, float speed) {
...
+  bool  noHit      = dist>float(MaxMagRange);
+  bool  invincible = !checkDamageMask(nsrc,nother,&b);
+  auto  dmg        = b.damage();
+
+  if(!b.isSpell()) {
+    auto& script    = nsrc.world().script();
+    int   hitChance = int(script.rand(100));
+    int   hitCh     = 0;
+    bool  g2        = Gothic::inst().version().game==2;
+    float critRange = g2 ? CritBowRange : (CritBowRange + 500);
```
Try, line 64: `500` is magical. Can be `ReferenceBowRangeG1`/`ReferenceBowRangeG2`.

Also this is more `Reference` range, since at 15meters hit chance matches player stats exactly.

game/game/damagecalculator.cpp
```diff
@@ -51,17 +51,44 @@ DamageCalculator::Val DamageCalculator::damageFall(Npc& npc, float speed) {
...
+    float skill     = b.hitChanceVal();
+
+    if(dist<critRange)
+      hitCh = int((skill - 100.f) / critRange * dist + 100.f + 0.5f); else
```
Try, line 68: Waaaaait... this is `((critRange-dist)/critRange)*skill*100.f + 0.5f`?

Also measuring probability in 0..100 rage is awkward. It makes way more sense to use 0..1, as in probability theory, and cast it to integer scale at very end, than you'll get:
`((critRange-dist)/critRange)*skill`

game/world/objects/npc.cpp
```diff
@@ -3503,12 +3502,15 @@ bool Npc::shootBow(Interactive* focOverride) {
...
   b.setOrigin(this);
   b.setDamage(DamageCalculator::rangeDamageValue(*this));
-
-  auto rgn = currentRangeWeapon();
-  if(rgn!=nullptr && rgn->isCrossbow())
-    b.setHitChance(float(hnpc->hitchance[TALENT_CROSSBOW])/100.f); else
-    b.setHitChance(float(hnpc->hitchance[TALENT_BOW]     )/100.f);
-
+  if(Gothic::inst().version().game==1) {
+    b.setHitChanceVal(float(hnpc->attribute[ATR_DEXTERITY]));
```
Try, line 3506: As noted in previous comments: probability should be in `0..1`

game/game/fightalgo.cpp
thokkat: Sorry forgot to mention. Before PR changes hit chance was always 100% below this range an 0% above. Now it's variable so it's unclear what this value should be. Maybe `ReferenceRange` to assure a somewhat decent hit chance?

game/game/damagecalculator.cpp
thokkat: If `dist=critRange` both formulas give `0` instead of `skill`. I just used 2 point formula for linear functions. I thought it's better to let it as is and let compiler optimize.

I used 0..100 range because of the comparison with the random generated number which is in this range in `swordDamage`. But 0..1 range is also ok and avoids rounding.

```C++
    float hitChance = float(script.rand(100))/100.f;
    if(dist<refRange)
      hitCh = (skill - 1.f) / refRange * dist + 1.f; else
      hitCh = skill / (refRange - float(MaxBowRange) * (dist - float(MaxBowRange)));
    noHit = (dist>float(MaxBowRange) || hitCh<=hitChance);
```

game/game/fightalgo.cpp
Try: Yes, `ReferenceRange` is good choise

thokkat:
Addressed the requested changes. I didn't cast to int for hit chance calculation because the calculated value is a decimal number.

Try:
Merged, thanks!

PR #436, opened 2023-04-02
Set `bestAzimuth` to zero in ranged camera mode for G1

thokkat:
Only for ranged Cam `bestAzimuth` differs between G1 and G2. It's `0` for G2 but `30` for G1. This leads to an unwanted camera perspective, so set it to `0` instead.

### DIFF
--- a/game/camera.cpp
+++ b/game/camera.cpp
@@ -594,6 +594,9 @@ void Camera::calcControlPoints(float dtF) {
     rotBest      = Vec3();
     //spin.y += def.bestAzimuth;
     }
+  else if(Gothic::inst().version().game==1 && camMod==Ranged) {
+    rotBest = Vec3();
+    }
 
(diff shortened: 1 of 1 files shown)


Try:
>  This leads to an unwanted camera perspective, so set it to 0 instead.

Do you know the root cause, why it's set to `30`, instead of zero in G1? In other work is ignoring of `bestAzimuth` aligned with what G1 engine expected to do, or just workaround ?

thokkat:
Was meant just as a workaround. I now checked marvin `toggle camdebug` output. Camera is in normal mode when a ranged weapon is drawn and while swimming, melee mode for magic. There's also a `CAMMODRUN` for moving if no weapon is drawn. 

So correct solution would be to set normal cam for ranged weapons for G1?

Try:
Yes, aligning with now known G1 behavior regards camera modes make a lot more sense

thokkat:
I open a new PR.

PR #437, opened 2023-04-03
Adjust camera selection for G1

thokkat:
If in certain weapon or move state G1 does not always use the corresponding camera. If a ranged weapon is drawn `CAMMODNORMAL` is used because `CAMMODRANGED` has a unwanted 30 degree rotation offset. 

There are other cams in vanilla like `CAMMODNORMAL` while swimming or `CAMMODRUN` for non combat moving. I haven't touched any of those because there are no issues with current camera usage.


Try:
LGTM - merged, thanks!

PR #440, opened 2023-04-06
Bump phoenix to v1.1.1 and update remote URL

lmichaelis:
Hi @Try, I'm moving phoenix into a new [GitHub organization](https://github.com/GothicKit) so the Git submodule remote should be updated with this patch. I also noticed that you're on an old version of phoenix, so I have bumped it up to [`v1.1.1`](https://github.com/GothicKit/phoenix/releases/tag/v1.1.1) :>


Try:
> moving phoenix into a new [GitHub organization](https://github.com/GothicKit)

Whoa, nice :)  Merged!

PR #441, opened 2023-04-10
fix: input problem when releasing one of two mapped keys

PoetaKodu:
## Problem

When you press a primary key action and then press the secondary one, if you then release one of the keys the action stops. Practical example:
- Press primary forward key (<kbd>W</kbd>)
- Press secondary forward key (<kbd>Up</kbd>)
- Release <kbd>W</kbd>
- **Problem:** your character stops moving forward.

## Fix

I fixed this problem by introducing tracking ([`KeyStatus`](https://github.com/Try/OpenGothic/blob/e55bb06514844881de4c239954414bf586bce01a/game/game/playercontrol.h#L78)) of both mappings: `Primary` and `Secondary`. Action is considered "on" if at least one mapped button is pressed.

### DIFF
--- a/game/game/playercontrol.cpp
+++ b/game/game/playercontrol.cpp
@@ -518,16 +518,16 @@ bool PlayerControl::tickMove(uint64_t dt) {
       return true;
       }
 
-    auto turningModifier = movement.turnRightLeft.value();
-    if(turningModifier > 0.f)
+    auto turningVal = movement.turnRightLeft.value();
+    if(turningVal > 0.f)
       camera->rotateRight(dt);
-    else if(turningModifier < 0.f)
+    else if(turningVal < 0.f)
       camera->rotateLeft(dt);
 
-    auto forwardModifier = movement.forwardBackward.value();
-    if(forwardModifier > 0.f)
+    auto forwardVal = movement.forwardBackward.value();
+    if(forwardVal > 0.f)
       camera->moveForward(dt);
-    else if(forwardModifier < 0.f)
+    else if(forwardVal < 0.f)
       camera->moveBack(dt);
     return true;
     }
--- a/game/game/playercontrol.cpp
+++ b/game/game/playercontrol.cpp
@@ -32,7 +32,7 @@ void PlayerControl::setTarget(Npc *other) {
   const auto ws    = pl->weaponState();
   const bool melle = (ws==WeaponState::Fist || ws==WeaponState::W1H || ws==WeaponState::W2H);
   if(other==nullptr) {
-    if(!(melle && ctrl[Action::ActionGeneric])) {
+    if(!(melle && ctrl[Action::ActionGeneric].isEnabled())) {
       // dont lose focus in melee combat
       pl->setTarget(nullptr);
       }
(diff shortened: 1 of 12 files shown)


Try:
Hm, looking at problem statement, I'm not sure about whole this.
* problem is not relevant to single-shoot, only for repeatable ones
* repeatable are.. movement only, right?

If problem exists exclusively for `WASD`, than redoing whole control scheme seem to be overkill to me.  We should probably look for alternatives, that are more local, preferably fully localized to PlayerControl

PoetaKodu:
We can use an approach similar to how Unreal Engine handles input. There are two types of inputs:
- Input Action - press/release events are relevant
- Input Axis - being pressed is relevant

Similar to this we can divide inputs to two groups and handle them separately. What do you think about this?

PoetaKodu:
Hi again. I finished the PR. I've implemented it in a very non-invasive way. The changes affected only the following movement input:

- fwd/bwd movement
- strafe left/right movement
- turning left/right

The way other input work was not affected.

Try:
Seems this PR can be improved and make even less invasive:
`DocumentMenu`, `InventoryMenu`, `MenuRoot` not suppose to care about mappings, only `PlayerControl::onKeyPressed/onKeyReleased` need to know about mapping.

My propose here is to keep
```
Action KeyCodec::tr(Tempest::KeyEvent& e) const;
Action KeyCodec::tr(Tempest::MouseEvent& e) const;
```
and add
```
Mapping KeyCodec::mapping(Tempest::KeyEvent& e) const;
```
and change `onKeyPressed` to take mapping  as one more argument.

PoetaKodu:
sure, I'll do it

PoetaKodu:
Done :p

game/game/playercontrol.cpp
```diff
@@ -820,7 +846,7 @@ void PlayerControl::implMove(uint64_t dt) {
     pl.setAnim(ani);
     }
 
-  setAnimRotate(pl, rot, ani==Npc::Anim::Idle ? rotation : 0, ctrl[KeyCodec::RotateL] || ctrl[KeyCodec::RotateR], dt);
+  setAnimRotate(pl, rot, ani==Npc::Anim::Idle ? rotation : 0, movement.turnRight.any(), dt);
```
Try, line 849: Naming `turnRight` is somewhat confusing, makes reader thing, that there is `turnLeft` somewhere. Maybe simply `turn`?

game/game/playercontrol.h
```diff
@@ -75,6 +75,68 @@ class PlayerControl final {
 
     using Action=KeyCodec::Action;
 
+    // Forward-Backward, Right-Left
+    static constexpr size_t NumAxes = 2;
+
+    struct AxisStatus { 
+
+        /// Main direction (e.g. W or Up arrow)
+        std::array<bool, 2> main;
```
Try, line 84: Should it be `NumAxes`, instead of 2? Or maybe something like `KeyCodec::NumMappings`?

game/game/playercontrol.h
```diff
@@ -75,6 +75,68 @@ class PlayerControl final {
...
+        
+        /// Reverse direction (e.g. S or Down arrow)
+        std::array<bool, 2> reverse;
+
+        /// Returns the current axis value (scale from -1 to 1)
+        auto modifier() const -> float {
```
Try, line 90: code: `auto modifier()`
comment: `current axis value`

Maybe `auto value() const` is a better name? Also no need to do so many comment and empty lines - code is quite readable as-is

Try (changes requested):

PoetaKodu:
I've applied the requested changes. The only thing I did different is with `turn` and other names. By changing `turnRight` to `turn` I'd have to also change `strafeRight` to `strafe` which would be misleading.

Try:
I've made a few more code-style changed, and merged. Thanks!

PR #443, opened 2023-04-10
Fallback for invalid start point

thokkat:
When entering `freemine.zen` the waypoint `ENTRANCE_FREEMINE_FREEMINECAMP` is requested in `insertPlayer` routine. But the real name is `ENTRANCE_OLDMINE_SURFACE` and player insertion fails. Vanilla uses `(0,0,0)` as fallback which works because it's close to that point in this particular world.

Instead of failing just fall back to the waypoint from `startPoint` routine which uses `(0,0,0)` as well as last resort, but usually returns a better point created from `zCVobStartpoint`. 

Start points created from `zCVobStartpoint` have an empty name except for `world.zen` in G1. Since empty names aren't considered in `findPoint` routine, I set it to `START` for these cases so zen loading with `-w` start parameter inserts the character at the correct position.

### DIFF
--- a/game/game/gamesession.cpp
+++ b/game/game/gamesession.cpp
@@ -39,12 +39,15 @@ void GameSession::HeroStorage::putToWorld(World& owner, std::string_view wayPoin
 
   if(auto pl = owner.player()) {
     pl->load(sr,0);
-    if(auto pos = owner.findPoint(wayPoint)) {
-      pl->setPosition  (pos->x,pos->y,pos->z);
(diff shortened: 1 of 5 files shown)


Try:
Hi, @thokkat !

> Since empty names aren't considered in findPoint routine, I set it to START for these cases so zen loading with -w start parameter inserts the character at the correct position.

Please, don't - waypoints/freepoints are observable from script. Changing names can cause unexpected behavior changes in game/mods

thokkat:
I made player creation use `insertPlayer` routine instead of `addNpc` to have the `startPoint `fallback as well.

PR #449, opened 2023-05-04
Ladder for pathfinding

thokkat:
Marvin `G` in vanilla shows that `USEMOB` events are triggered when the waypoint next to ladder is reached. The idea is to precompute ladder points by considering all edges of the waynet that intersect a ladder bounding box. The edge whose midpoint y-value is within a small fixed distance of the bbox's midpoint y-value is taken. The corresponding ladder is referenced in the waypoints of the edge. 
There's one edge which slips through between the bbox's of a double ladder, which is interestingly the only spot where no `USEMOB` event is sent in vanilla. To make it work the bbox's volume used is slightly increased.

Then in `implGoTo` it's checked if the waypoint and successor are ladder points and interaction is forced that case. Because the starting waypoint can be arbitrarily far away from interaction point it was necessary to save ladder in `GoTo` structure. Some small adjustments were made so `aigoto` marvin command works aswell.

Sometimes  e.g. when npc has reached top of ladder they fall down if there's a monster. In case the next waypoint has already been popped from `wayPath` a recalculation is required to make npc use ladder again because y-value isn't considered in pathfinding. There were some waypoints triggering this recalculation because they were sent down by `adjustWaypoints` routine if below the map. I added a check to ignore those.

I shifted the display label to the correct position for tilted ladders by using the formula from `tickMobsi` routine. For consistency I changed it there to look like in `interactive` constructor.

There was also a bug that interaction wasn't correctly ended when `quick` parameter is true in `deattach` routine. I also changed the used  routine for interaction ending in marvin `F8` to make sure `user` is `nullptr` afterwards.

For testing `goto waypoint NC_TAVERN_BACKROOM` and `set time 21 59` or `set time 2 59`. Then a npc should use ladder. There's also marvin `insert` or marvin `O` to watch them walk to their spots.

### DIFF
--- a/game/world/objects/interactive.cpp
+++ b/game/world/objects/interactive.cpp
@@ -96,6 +96,9 @@ Interactive::Interactive(Vob* parent, World &world, const phoenix::vobs::mob& vo
       --at;
     stepsCount = std::atoi(mdlVisual.c_str()+at);
     stateNum   = stepsCount;
+
+    transform().project(displayOffset);
+    displayOffset -= position();
     }
 
   world.addInteractive(this);
@@ -242,7 +245,7 @@ void Interactive::tick(uint64_t dt) {
   if(p->user==nullptr && (state==stateNum && p->attachMode))
     return;
 
-  if(isLadder() && p->started==Started)
+  if(isLadder() && p->started==Started && p->user!=nullptr && p->user->isAiQueueEmpty())
     return;
   implTick(*p);
   }
@@ -754,7 +757,7 @@ bool Interactive::dettach(Npc &npc, bool quick) {
     if(i.user==&npc && i.attachMode) {
       if(canQuitAtState(*i.user,state)) {
         auto sq = npc.setAnimAngGet(Npc::Anim::InteractToStand);
-        if(sq==nullptr)
+        if(sq==nullptr && !quick)
           return false;
         i.user       = nullptr;
         i.attachMode = false;
--- a/game/world/objects/interactive.cpp
+++ b/game/world/objects/interactive.cpp
@@ -416,6 +416,10 @@ std::string_view Interactive::displayName() const {
   return s->get_string();
   }
 
+const Tempest::Vec3* Interactive::bBox() const {
+  return bbox;
+  }
+
 bool Interactive::setMobState(std::string_view scheme, int32_t st) {
   const bool ret = Vob::setMobState(scheme,st);
   if(state==st)
@@ -755,16 +759,12 @@ bool Interactive::attach(Npc &npc) {
 bool Interactive::dettach(Npc &npc, bool quick) {
   for(auto& i:attPos) {
     if(i.user==&npc && i.attachMode) {
-      if(canQuitAtState(*i.user,state)) {
-        auto sq = npc.setAnimAngGet(Npc::Anim::InteractToStand);
-        if(sq==nullptr && !quick)
-          return false;
-        i.user       = nullptr;
-        i.attachMode = false;
-        npc.quitIneraction();
-        return true;
-        }
-      else if(quick) {
+      if(quick || canQuitAtState(*i.user,state)) {
+        if(!quick) {
+          auto sq = npc.setAnimAngGet(Npc::Anim::InteractToStand);
+          if(sq==nullptr)
+            return false;
+          }
         i.user       = nullptr;
         i.attachMode = false;
         npc.quitIneraction();
--- a/game/world/objects/interactive.h
+++ b/game/world/objects/interactive.h
@@ -46,7 +46,7 @@ class Interactive : public Vob {
     Tempest::Vec3       displayPosition() const;
     std::string_view    displayName() const;
 
-    auto                bBox() const { return &bbox; }
+    auto                bBox() const -> const Tempest::Vec3*;
 
     int32_t             stateId() const { return state; }
     int32_t             stateCount() const { return stateNum; }
--- a/game/world/objects/interactive.h
+++ b/game/world/objects/interactive.h
@@ -46,6 +46,8 @@ class Interactive : public Vob {
     Tempest::Vec3       displayPosition() const;
     std::string_view    displayName() const;
 
+    auto                bBox() const { return &bbox; }
+
     int32_t             stateId() const { return state; }
     int32_t             stateCount() const { return stateNum; }
     bool                setMobState(std::string_view scheme,int32_t st) override;
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1361,7 +1361,7 @@ bool Npc::implGoTo(uint64_t dt, float destDist) {
       go2.wp = go2.wp->hasLadderConn(wayPath.first()) ? wayPath.first() : wayPath.pop();
       if(go2.wp!=nullptr) {
         attachToPoint(go2.wp);
-        if(useLadder()) {
+        if(setGoToLadder()) {
           mvAlgo.tick(dt);
           return true;
           }
@@ -1371,7 +1371,11 @@ bool Npc::implGoTo(uint64_t dt, float destDist) {
     if(finished)
       clearGoTo();
     } else {
-    if(useLadder() || (mvAlgo.checkLastBounce() && implTurnTo(dpos.x,dpos.z,false,dt))) {
+    if(setGoToLadder()) {
+      mvAlgo.tick(dt);
+      return true;
+      }
+    if(mvAlgo.checkLastBounce() && implTurnTo(dpos.x,dpos.z,false,dt)) {
       mvAlgo.tick(dt);
       return true;
       }
@@ -1695,7 +1699,7 @@ bool Npc::implAiFlee(uint64_t dt) {
   return true;
   }
 
-bool Npc::useLadder() {
+bool Npc::setGoToLadder() {
   if(go2.wp==nullptr || go2.wp!=wayPath.first())
     return false;
   auto inter = go2.wp->ladder;
@@ -2120,7 +2124,7 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
       break;
       }
     case AI_GoToPoint: {
-      if(!setInteraction(nullptr)) {
+      if(isInAir() || !setInteraction(nullptr)) {
         queue.pushFront(std::move(act));
         break;
         }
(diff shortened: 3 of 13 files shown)


game/world/objects/npc.cpp
```diff
@@ -1345,29 +1348,48 @@ bool Npc::implGoTo(uint64_t dt, float destDist) {
...
+      mvAlgo.tick(dt);
+      return true;
+      }
     bool finished = true;
     if(go2.flag==GT_Way) {
+      auto wpoint = go2.wp;
       go2.wp = wayPath.pop();
       if(go2.wp!=nullptr) {
+        if(wpoint!=nullptr && wpoint->ladder==go2.wp->ladder)
+          go2.ladder = go2.wp->ladder;
```
Try, line 1373: Can you please clarify how this works:
`wpoint->ladder` and `go2.ladder` are generally thing, so what is reason to keep `go2.ladder` around?

game/world/objects/npc.cpp
```diff
@@ -2175,7 +2197,7 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
     case AI_StandUpQuick:
       // NOTE: B_ASSESSTALK calls AI_StandUp, to make npc stand, if it's not on a chair or something
       if(interactive()!=nullptr) {
-        if(!setInteraction(nullptr,false)) {
+        if((interactive()->isLadder() && !isPlayer()) || !setInteraction(nullptr,false)) {
```
Try, line 2200: Why player is special case here?

game/world/waymatrix.cpp
```diff
@@ -178,11 +181,51 @@ void WayMatrix::marchPoints(DbgPainter &p) const {
...
+    auto ray = world.physic()->landRay(w.position());
+    if(ray.hasCol)
+      w.y = ray.v.y;
     indexPoints.push_back(&w);
     }
   }
 
+void WayMatrix::calculateLadderPoints() {
+  float dist   = 100.f;
+  Vec3  offset = {3.f,0,3.f};
```
Try, line 193: Hm, padding for 3 centimeters? Is it really that impactful? 

also `static const`

game/world/objects/npc.cpp
```diff
@@ -3962,6 +3984,25 @@ void Npc::stopWalking() {
...
+    return false;
+  wayPath.clear();
+  go2.clear();
+  attachToPoint(nullptr);
+  wayPath       = owner.wayTo(*this,*target);
+  auto wpoint   = wayPath.pop();
+  if(wpoint!=nullptr) {
+    go2.set(wpoint);
+    attachToPoint(wpoint);
+  } else {
```
Try, line 3999: nitpcik: codestyle

game/world/objects/npc.cpp
```diff
@@ -1345,29 +1348,48 @@ bool Npc::implGoTo(uint64_t dt, float destDist) {
...
 
-  auto dpos = go2.target()-position();
+  auto  dpos      = go2.target()-position();
+  float destDistY = 200.f;
 
   if(go2.flag==GT_Flee) {
     // nop
     }
   else if(mvAlgo.isClose(go2.target(),destDist)) {
+    if(std::abs(dpos.y)>destDistY && recalculateWayPath()) {
```
Try, line 1363: Is it check for waypath pointing vertically to top, so ladder can be assumed?
Also `200` is somewhat arbitrary, in  theory it can be a small hill with `dpos = {1000, 200,0}`

Try:

So far this one is way more promising than a previous, I've posted some questions about implementation in comments.

Also sorry, for late response - been busy debugging water shader :|

game/world/objects/interactive.h
```diff
@@ -46,6 +46,8 @@ class Interactive : public Vob {
     Tempest::Vec3       displayPosition() const;
     std::string_view    displayName() const;
 
+    auto                bBox() const { return &bbox; }
```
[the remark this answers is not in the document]
thokkat: Something didn't work with accessing  `bbox[1]` before if returning `Vec*, but it does work now. Don't know what i did wrong. Now it's 

`auto bBox() const -> const Tempest::Vec3*;`
and
`const Tempest::Vec3* Interactive::bBox() const {  return bbox;  }`

game/world/waymatrix.cpp
```diff
@@ -178,11 +181,51 @@ void WayMatrix::marchPoints(DbgPainter &p) const {
...
-    w.y = world.physic()->landRay(w.position()).v.y;
+    auto ray = world.physic()->landRay(w.position());
+    if(ray.hasCol)
+      w.y = ray.v.y;
     indexPoints.push_back(&w);
     }
   }
 
+void WayMatrix::calculateLadderPoints() {
+  float dist   = 100.f;
```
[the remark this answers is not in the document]
thokkat: The biggest value occurring is around 75 so 100 is somewhat arbitrary. Value is unrelated to npc distance maybe name is a bit misleading?

game/world/objects/interactive.cpp
```diff
@@ -754,7 +757,7 @@ bool Interactive::dettach(Npc &npc, bool quick) {
     if(i.user==&npc && i.attachMode) {
       if(canQuitAtState(*i.user,state)) {
         auto sq = npc.setAnimAngGet(Npc::Anim::InteractToStand);
-        if(sq==nullptr)
+        if(sq==nullptr && !quick)
```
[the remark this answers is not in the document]
thokkat: Something like this and remove quick branch?
```c++
      if(canQuitAtState(*i.user,state) || quick) {
        if(!quick) {
          auto sq = npc.setAnimAngGet(Npc::Anim::InteractToStand);
          if(sq==nullptr)
            return false;
          }
        i.user       = nullptr;
        i.attachMode = false;
        npc.quitIneraction();
        return true;
        }
```

game/world/objects/npc.cpp
thokkat: If two consecutive waypoints in `Waypath` have the same non `nullptr` interactive we know a ladder has to be used. Ladder can have any distance to the waypoint and if a new waypoint is popped  the information about the previous one is lost. The idea is to save ladder in `go2` and try `setInteraction` in every iteration. This works  because the edge intersects with ladder bbox so npc gets close eventually.

game/world/objects/npc.cpp
thokkat: If player uses ladder while this event is triggered automatic climbing occurs. 

Probably the check for empty aiqueue in `Interactive::tick` to find out if marvin `aigoto` is used is bad. Maybe check if go2 flag is not `GT_No` is better?

game/world/waymatrix.cpp
thokkat: There's one double ladder which fails otherwise. But i don't think any npc uses that ladder and it's not working in vanilla. So could remove it.

game/world/objects/npc.cpp
thokkat: Ladder points are completely precomputed while loading. Every ladder is correctly associated with a waypoint or not except the one we talked above. 

There can be a problem if npc climbs ladder, pops waypoint and falls down because of a monster or bad terrain. Then walking is continued on the lower floor because y-value is not used in waypoint distance checking. Maybe there is a better way to achieve this?

thokkat:
> Also sorry, for late response - been busy debugging water shader :|

No problem we all want shiny graphics :)

game/world/objects/npc.cpp
Try: Do I understand correctly:
in sequence of waypoints `w[0], w[1], w[2], w[3]` , if `w[1]` and `w[2]` point to same ladder npc need to use it, when approaching `w[1]` ?

would it be easier to implement `wayPath.peek()` to peek into next point without dequeuing it from list, so point can be poped, only after `setInteraction` succeeded?

game/world/objects/npc.cpp
thokkat: > Hm, maybe than `AI_StandUpQuick` should have no effect on ladders?

You mean ignore or push to queue again?  I swapped empty aiqueue with interactive check in `implMove` in playercontrol for marvin `aigoto`. If player is out of dialog range this ends in a deadlock if interaction is not ended.

> In script(G2), `AI_StandUpQuick` is called from dialog related code, and a few spells: `ZS_MagicSleep`, `ZS_Zapped`, `ZS_Pyro` and some else.

Seems to be called every time before a monster is attacked.

> Can you check please what happens in G1, when some npc climbing ladder and sleep spell is used on him?

Npc climbs, walks to target and then falls asleep.

game/world/objects/npc.cpp
```diff
@@ -1691,6 +1695,21 @@ bool Npc::implAiFlee(uint64_t dt) {
   return true;
   }
 
+bool Npc::useLadder() {
```
Try, line 1698: Name `useLadder` suggests general use, but this function can really work only with go2-pathing

game/world/objects/npc.cpp
```diff
@@ -1358,16 +1358,20 @@ bool Npc::implGoTo(uint64_t dt, float destDist) {
...
+          return true;
+          }
         finished = false;
         }
       }
     if(finished)
       clearGoTo();
     } else {
-    if(mvAlgo.checkLastBounce() && implTurnTo(dpos.x,dpos.z,false,dt)){
+    if(useLadder() || (mvAlgo.checkLastBounce() && implTurnTo(dpos.x,dpos.z,false,dt))) {
```
Try, line 1374: Expression is too complex with multiple side-effects inside if.
Can you please split it into 2?
```
if(ladder()) {
  ...
  }
if(checkLastBounce && implTurnTo()) {
  ...
  }
```

game/world/objects/npc.cpp
Try: > Npc climbs, walks to target and then falls asleep.

In that case enqueue again `AI_StandUpQuick` make most sense:
this will block all other commands until climb animation is finished

thokkat:
In original if player receives `AI_StandUpQuick` he's detached from ladder. The same happens if npc uses sleep spell on player, though player is stuck after hitting the ground. Sleep generally only seems to start to have an effect if a walking npc has reached it's target.
I'd like to let it as is, enqueue for npc if on ladder so monster at top doesn't cause a climbing loop, but detach player for a triggered dialog with npc.

I also added the `isInAir` check for `AI_GoToPoint` to make the waypath calculation viable in that case.

Try:
Merged, thanks!

PR #454, opened 2023-05-13
Svm's for G1

thokkat:
In G1 output related script functions sometimes have `NULL` as target instead of `other`. This is corrected by assigning `other` in gamescript file in those cases. This is motivated by e.g. a line like this  
`if (Wld_DetectNpc(self,	-1, ZS_Smalltalk, -1) && (Npc_GetDistToNpc	(self, other)<HAI_DIST_SMALLTALK))`
where `other` is assigned and distance checked just to call `B_Say( self, NULL, "$SMALLTALK01");` without a target.

When a daily routine ends at midnight `hours` returned from `endTime` in `tickRoutine` are zero. In that case `day` is incremented by one now as compensation. This should fix the npc's looking at a wall https://github.com/Try/OpenGothic/issues/409. This behavior is caused by the lines	

```
else if (Npc_GetStateTime(self) >= 5)
	{
	    PrintDebugNpc		(PD_TA_CHECK,	"... kein Gesprächspartner gefunden!");
	    
	    AI_StartState	(self, ZS_Stand, 1, "");
	};

```
If no talking partner is found npc just stands. This can be observed in vanilla when talking to an npc. `ZS_Smalltalk` state is left and other talking partner npc turns away. This is generally a problem when smalltalk routine starts and one npc still has to walk to that location, for example `Baal Taran` has to walk to location mentioned in the issue at midnight and conversation fails. In vanilla this is alleviated by offloading npc's out of range and reloading at right location.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1277,6 +1277,14 @@ uint32_t GameScript::rand(uint32_t max) {
   return uint32_t(randGen())%max;
   }
 
+Npc* GameScript::globalOther() {
+  if(vm.global_other()->is_instance_of<phoenix::c_npc>()) {
+    auto cNpc = reinterpret_cast<phoenix::c_npc*>(vm.global_other()->get_instance().get());
+    return findNpc(cNpc);
+    }
+  return nullptr;
+  }
+
 Npc* GameScript::findNpc(phoenix::c_npc *handle) {
   if(handle==nullptr)
     return nullptr;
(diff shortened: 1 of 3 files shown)


game/game/gamescript.cpp
```diff
@@ -1277,6 +1277,14 @@ uint32_t GameScript::rand(uint32_t max) {
   return uint32_t(randGen())%max;
   }
 
+Npc* GameScript::globalOther() {
```
Try, line 1280: Having `globalOther` and have no `globalSelf`, `globalVictum`, etc doesn't fill quite right.

Maybe you can create `Npc* GameScript::findNpc(const symbol*)` instead?

game/game/gamescript.cpp
```diff
@@ -2590,7 +2597,7 @@ void GameScript::ai_turntonpc(std::shared_ptr<phoenix::c_npc> selfRef, std::shar
   }
 
 void GameScript::ai_outputsvm(std::shared_ptr<phoenix::c_npc> selfRef, std::shared_ptr<phoenix::c_npc> targetRef, std::string_view name) {
-  auto target = findNpc(targetRef);
+  auto target = targetRef==nullptr ? globalOther() : findNpc(targetRef);
```
Try, line 2600: You mention that `NULL` case is specific to G1, so `VersionInfo` need to be used, similar to how `hasZSStateLoop` is implemented

Try (commented):

Generally LGTM, added few comments

thokkat:
Addressed the requested changes.

PR #457, opened 2023-05-22
`npc_isdead` : return false if npc is null

thokkat:
In `ZS_Attack.d` there's this line that checks in G1 if plunder animation should be played:

`if (!C_NpcIsBoss(self) && C_NpcIsHuman(other) && (Npc_IsInState(other, ZS_Unconscious) || Npc_IsDead(other) || Npc_IsInState(other, ZS_MagicSleep)) ) `

`other` can be `null` (triggered when running away) and then animation is played if `true` is returned but npc is still up.

G2 has only four occurrences of `Npc_IsDead(other)` and always connected with other script functions so nothing is executed if npc is `null`.


Try:
Hi, @thokkat , and thanks for PR! Merged.

PR #471, opened 2023-06-23
marvin 'toggle desktop'

Nindaleth:
- hides the framerate, clock, HP bar, mana bar, oxygen bar from UI

This is something I've noticed in the screenshots - AFAIK all of them contain the HP bar and mana bar and it turns out the code is nicely prepared to hide them when needed. I'm not sure if it's something that should be available through a shortcut, so I've added a new marvin command instead.

### DIFF
--- a/game/gothic.h
+++ b/game/gothic.h
@@ -88,6 +88,9 @@ class Gothic final {
     bool         isGodMode() const { return godMode; }
     void         setGodMode(bool g) { godMode = g; }
 
+    void         toggleDesktop() { desktop = !desktop; }
+    bool         isDesktop() { return desktop; }
+
     bool         doHideFocus () const { return hideFocus; }
     bool         doFrate() const { return showFpsCounter; }
     void         setFRate(bool f) { showFpsCounter = f; }
@@ -166,6 +169,7 @@ class Gothic final {
     uint16_t                                pauseSum=0;
     bool                                    isMarvin       = false;
     bool                                    godMode        = false;
+    bool                                    desktop        = false;
     bool                                    showFpsCounter = false;
     bool                                    showTime       = false;
     bool                                    hideFocus      = false;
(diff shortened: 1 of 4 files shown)


Try:
Hi, @Nindaleth , and thanks for PR!

There is same command in vanilla: `TOGGLE DESKTOP`, no need for new name.
And there are also config options in SystemPack.ini:
```
[INTERFACE]

ShowManaBar=1
; ... mana bar settings: is always invisible (0), is only visible when the hero is holding a spell in his hand (1), is always visible (2), default in the game 1.

ShowSwimBar=1
; ... remaining air bar (for swimming underwater) settings: is always invisible (0), is only visible when underwater (1), is always visible (2), default in the game 1.

HideHealthBar=0
; ... life bar settings: is always visible (0), is always invisible (1), default is 0.
```

Nindaleth:
I've done the easy thing and renamed `photo mode` to `desktop`, keeping it part of the Camera class. But I think it would fit more into the Gothic class, is that correct?
I also intend to do the ini option handling in this PR, with precedence default < INI set differently < Marvin toggled to hide.

Try:
> But I think it would fit more into the Gothic class, is that correct?

Yes

Try:
@Nindaleth 
Have already drafted release 1.0.2182, to be released early next week. Do you want to to hold it, so we can finish this PR, or merge it later, after release, is fine?

Nindaleth:
Oh a new release, that's great news! I wouldn't want to block release for people who can't compile themselves.

How about I finish the `toggle desktop` part as soon as I can so that it's in this release? And the INI part can be done in a next PR.

Try:
> How about I finish the toggle desktop part as soon as I can so that it's in this release?

Yep, sounds good. `toggle desktop` looks finished - should we proceed with merge?

Nindaleth:
Yes please, let's merge :+1:

PR #477, opened 2023-07-02
parse Game and OutputUnits from mod ini if set

### LINKED ISSUE #473: G2 classic mod - language selection doesn't work
G2 classic mod allows playing NOTR as if it was the original non-expansion G2. v1.1 download link: https://www.worldofgothic.de/dl/download_230.htm

Running via `./Gothic2Notr.sh -g ~/Games/gothic-ii-gold-edition/drive_c/GOG\ Games/Gothic\ 2\ Gold/ -game:G2_Classic.ini` the in-game texts are in German despite the ini containing `_EN` suffixes by default. In the original the default results in English, removing the `_EN` suffixes does the switch to German.

Other than being in German, the mod appeared to work during a very short test (different main menu background, Cavalorn and Greg not present on their NOTR start points).

Nindaleth:
fixes #473

Tested that running with the G2Classic mod in `-game` switches language depending on whether I've chosen the `_EN` or suffix-less values. The standard experience without mods also works. Didn't test with G1 but the related defaults of `GOTHIC.DAT` and `OU.*` are supposed to be the same.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -60,7 +60,7 @@ bool GameScript::GlobalOutput::isFinished() {
   }
 
 GameScript::GameScript(GameSession &owner)
-  :owner(owner), vm(Gothic::inst().loadPhoenixScriptCode(Gothic::inst().getGameDatName()), phoenix::execution_flag::vm_allow_null_instance_access) {
+  :owner(owner), vm(Gothic::inst().loadPhoenixScriptCode(Gothic::inst().defaultGameDatFile()), phoenix::execution_flag::vm_allow_null_instance_access) {
 
   if (vm.global_self() == nullptr || vm.global_other() == nullptr || vm.global_item() == nullptr ||
       vm.global_victim() == nullptr || vm.global_hero() == nullptr)
@@ -404,15 +404,8 @@ void GameScript::initDialogs() {
 
 void GameScript::loadDialogOU() {
   auto gCutscene = Gothic::inst().nestedPath({u"_work",u"Data",u"Scripts",u"content",u"CUTSCENE"},Dir::FT_Dir);
-  std::string prefix = std::string(Gothic::inst().settingsGetS("FILES","OutputUnits"));
-  std::vector<std::string> names;
-  if (prefix == "") {
-    names.emplace_back("OU.DAT");
-    names.emplace_back("OU.BIN");
-    } else {
-    names.emplace_back(prefix + ".DAT");
-    names.emplace_back(prefix + ".BIN");
-    }
+  std::string prefix = std::string(Gothic::inst().defaultOutputUnits());
+  std::vector<std::string> names = {prefix + ".DAT", prefix + ".BIN"};
 
   for(auto OU:names) {
     if(Resources::hasFile(OU)) {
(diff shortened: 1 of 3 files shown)


game/game/gamescript.cpp
```diff
@@ -404,10 +404,15 @@ void GameScript::initDialogs() {
 
 void GameScript::loadDialogOU() {
   auto gCutscene = Gothic::inst().nestedPath({u"_work",u"Data",u"Scripts",u"content",u"CUTSCENE"},Dir::FT_Dir);
-  static const char* names[] = {
-    "OU.DAT",
-    "OU.BIN",
-    };
+  std::string prefix = std::string(Gothic::inst().settingsGetS("FILES","OutputUnits"));
+  std::vector<std::string> names;
```
Nindaleth, line 408: I suppose this is suboptimal compared to the previous state, please feel free to change to your taste.

game/gothic.h
```diff
@@ -48,6 +48,8 @@ class Gothic final {
     std::string_view defaultPlayer() const;
     std::string_view defaultSave()   const;
 
+    std::string  getGameDatName();
```
Try, line 51: see functions above:
`std::string_view defaultGameDatFile()  const;`

game/gothic.cpp
```diff
@@ -660,6 +678,8 @@ void Gothic::settingsSetS(std::string_view sec, std::string_view name, std::stri
 float Gothic::settingsGetF(std::string_view sec, std::string_view name) {
   if(name.empty())
     return 0;
+  if(instance->modFile != nullptr && instance->modFile->has(sec,name))
```
Try, line 681: modFile not suppose to have any settings, therefore is not part of init-chain. See how `wrldDef` is implemented.

game/game/gamescript.cpp
Try: This part is OK - one more allocation wont hurt load-time

game/gothic.cpp
```diff
@@ -575,6 +578,15 @@ std::string_view Gothic::defaultSave() const {
   return CommandLine::inst().defaultSave();
   }
 
+std::string Gothic::getGameDatName() {
+  if (settingsGetS("FILES","Game") != "") {
```
Try, line 582: Why if-else? `gameDef` appears to be constant (`=="GOTHIC.DAT"`) for most of it's life-time.
This should be implemented same as `plDef`/`wrldDef`

game/gothic.cpp
Nindaleth: I wouldn't have a way to get the OutputUnits value then, should I have an `ouDef` too?

Try: same as with world?
`std::string_view defaultOutputUnits() const`

Nindaleth: Ok, that's a yes, then. WIll do.

game/gothic.h
Nindaleth: Done

game/gothic.cpp
Nindaleth: Done

game/gothic.cpp
Nindaleth: ok, this should be now in line with `plDef` & `wrldDef`

Nindaleth:
Let me know if this way it's alright, I'll squash the commits before merging.

Try:
LGTM, also I'm always do "squash and merge", in github anyway :)

Issue #322, opened 2022-09-16
NPCs can stand on water surface


Dehela:
A bit of Jesus role-playing. Here this guard is angry at me so whenever I try to stand on solid ground, he would attempt to chase me and end up swimming; and he will stand on water if I get to water myself or distance from him.

![1](https://user-images.githubusercontent.com/111627003/190633596-97db0f9b-d91b-4ed2-bce1-9ee2b5f54b76.jpg)

Try:
Closing as duplicate of #320  - there is no difference between humans and other living creatures on engine side

PR #482, opened 2023-07-06
Draw weapon if attacking

thokkat:
Edit: As pointed out vanilla engine issues a draw weapon command if npc is in attack mode and no weapon has been drawn. The only npc's relying on this are orcs in G1 orc graveyard because script doesn't take care in that case.

Orcs in G1 orc-graveyard are in `ZS_MM_ATTACK` state when attacking. G2 script has this check for weapon draw:

`if(C_NpcIsMonsterMage(self) || (self.guild == GIL_SKELETON) || (self.guild == GIL_SUMMONED_SKELETON) || (self.guild > GIL_SEPERATOR_ORC))`

G1 script only checks for MonsterMage. Since skeletons have their weapon already drawn only the orc case has to be handled.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1430,9 +1430,10 @@ bool Npc::implAttack(uint64_t dt) {
 
   auto ws = weaponState();
   // vanilla behavior, required for orcs in G1 orcgraveyard
-  if(ws==WeaponState::NoWeapon)
-    drawWeaponMelee();
-  ws = weaponState();
+  if(ws==WeaponState::NoWeapon) {
+    if(drawWeaponMelee())
+      return true;
+    }
 
   FightAlgo::Action act = fghAlgo.nextFromQueue(*this,*currentTarget,owner.script());
 
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1414,8 +1414,7 @@ bool Npc::implAttack(uint64_t dt) {
     return false;
     }
 
-  auto ws = weaponState();
-  if(ws==WeaponState::NoWeapon)
+  if(!fghAlgo.hasInstructions())
     return false;
 
   if(bodyStateMasked()==BS_STUMBLE) {
@@ -1429,9 +1428,7 @@ bool Npc::implAttack(uint64_t dt) {
     return true;
     }
 
-  if(!fghAlgo.hasInstructions()) {
-    return false;
-    }
+  auto ws = weaponState();
 
   FightAlgo::Action act = fghAlgo.nextFromQueue(*this,*currentTarget,owner.script());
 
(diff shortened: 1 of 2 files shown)


game/world/objects/npc.cpp
```diff
@@ -2357,7 +2357,13 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
       break;
       }
     case AI_Attack:
-      if(currentTarget!=nullptr && weaponState()!=WeaponState::NoWeapon){
+      if(weaponState()==WeaponState::NoWeapon) {
+        // WA for orcgraveyard orcs
+        if(owner.version().game==1 && guild()>GIL_SEPERATOR_ORC)
```
Try, line 2362: would it make more sense to not to check for the orc here?
logic seem to be: if G1 and we are going to attack -> draw some weapon. Also consider `drawWeaponFist` as a case, when no melee is equipped

thokkat: > logic seem to be: if G1 and we are going to attack -> draw some weapon.

I suspected something like this but couldn't prove it but actually there's a simple way. 

Tested vanilla and disabled all script functions that draw a weapon or spell. Command seems to be `drawWeaponMelee`. Every npc without a melee weapon attacks in fist mode. Also active in G2.

game/world/objects/npc.cpp
```diff
@@ -2357,7 +2357,11 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
       break;
       }
     case AI_Attack:
-      if(currentTarget!=nullptr && weaponState()!=WeaponState::NoWeapon){
+      if(weaponState()==WeaponState::NoWeapon) {
+        if(!drawWeaponMelee())
+          break;
```
Try, line 2362: Not 100% about this `break`

Basically you say, if weapon can't be drawn (`npc` is in water for example), then command should be skipped.
Skipping command mean `npc` wont attack(fine), but also won't chaise and move resulting in soft-lock in water.

Maybe it's better to move `drawWeaponMelee` part into `implAttack` ?

Try:
> There's still some buggy behavior if npc is spawned in water.

you mean standing, like this: https://github.com/Try/OpenGothic/issues/322 ? Also how npc been spawned in water ? Summon spell?

game/world/objects/npc.cpp
```diff
@@ -1429,9 +1428,11 @@ bool Npc::implAttack(uint64_t dt) {
     return true;
     }
 
-  if(!fghAlgo.hasInstructions()) {
-    return false;
-    }
+  auto ws = weaponState();
+  // vanilla behavior, required for orcs in G1 orcgraveyard
+  if(ws==WeaponState::NoWeapon)
+    drawWeaponMelee();
```
Try, line 1434: probably:
```
if(ws==WeaponState::NoWeapon) {
  drawWeaponMelee();
  return true;
  }
```
otherwise npc may do 2 things in same tick

thokkat:
>  you mean standing, like this

Yes like in the issue. I spawned npc with marvin insert command. For example `insert pc_mage_nw` in khorinis harbor. If taken over with `o` swimming animation is played and after going back to player standing pose is triggered again.
Other npc like `insert Pal_202_Albrecht` have swimming animation.

Try:
Merged, thanks!

> For example insert pc_mage_nw in khorinis harbor.

This to be add up to swimming issues

PR #491, opened 2023-07-15
Some script functions for G1

thokkat:
`wld_exchangeguildattitudes`
Replaces the guild table. Needed to change guild attitudes after attack of new mine.

`npc_hasspell`
Checks if a spell is present in inventory. Needed to let npc use spells.

`npc_hasreadiedweapon` and  `npc_hasreadiedrangedweapon`
Checks if npc has drawn any weapon/ranged weapon. Used to react if player wields a weapon.

`npc_hasrangedweaponwithammo`
Checks if npc has ranged weapon and corresponding ammo. Needed for npc's to draw ranged weapons.

`ai_unreadyspell`
Opposite of `ai_readyspell`. Puts spell away.

I removed `spellById` because it's not used anywhere and intoduced `hasSpell` instead.


Try:
Merged, thanks!

PR #496, opened 2023-07-26
Revert `npc_isdead : return false if npc is null`

thokkat:
The change was wrong. Testing shows that original engine indeed returns true in that case. I played some mod and two quests couldn't be completed due to this. I also can't reproduce the G1 behavior that was the reason for the change.


Try:
Merged, thanks)

Issue #497, opened 2023-07-27
Weapon stats can be applied multiple times


thokkat:
In `AI_DrawWeaponMelee` `drawWeaponMelee` eventually calls `switchActiveWeapon` that updates the damage stats. Then function returns true and `drawWeaponMelee` is not repushed to queue anymore. Problem is that until weaponstate is updated other `drawWeaponMelee` commands can be called and increase stats again. I found two occurences.

First is the always draw a weapon before attacking command https://github.com/Try/OpenGothic/commit/772c779f620e37487cfc7782df5cf299751b8181. When `Ai_attack` is sent weapon state is still `NoWeapon` and `drawWeaponMelee `is called a few times in `implAttack`. Test case can be Cavalorn fighting Goblins at beginning.

Second is script itself that sends a second draw weapon command, probably because a check for readied weapon returned false. Test case is again Cavalorn though it does not always trigger. Lure the bandit to him and a second `AI_DrawWeaponRange` is sent after first is completed but before weapon state could be updated.

As i understand weaponstate is defined by animation, so some kind of transition gap? Couldn't `hnpc->weapon` be used that has the same values but assigned at the right time? Quick solution i tested would be to only update stats if `next` is different than `active`.

Try:
Hi @thokkat, thanks for reporting!

> Quick solution i tested would be to only update stats if next is different than active.

seen to be a correct one: `active` is in inner engine state, unreachable by script - feel free to use.
Also, I'm currently occupied with ikarus support, are you planning to do PR?

thokkat:
Sure, actually I already had prepared a different one but then found the bow case and wanted to hear your thoughts first.

PR #498, opened 2023-07-28
Fix abnormal weapon damage

thokkat:
I made additional change that `hnpc->weapon` in `drawWeaponBow` is set depending on weapon state `Bow` instead of `W1H`

fixes https://github.com/Try/OpenGothic/issues/497


Try:
Thanks for PR, merged!

PR #499, opened 2023-07-30
Blend out rotation anim

kuzy000:
Maybe this can be applied to other animations as well by replacing `stopAnim` with `stopAnimBlended` but I'm not sure how to test it properly


Try:
Hi, @kuzy000 !

I've look into you solution, probably it best not to merge it. Like you said, it only works for rotation anim. We already do have one partial solution with `Layer::blendOut` - no need to multiply them.
At this point looking for proper solution, and only minor changes to `blendOut` can be considered. Probably bleed it into base-layer is good enough.

Try:
Implemented bleed-based solution instead in #d7d2e633

PR #502, opened 2023-08-05
Take previous npc routine if none is set for a time interval

thokkat:
I found a buggy daily npc routine where some time intervals have two identical routines and one interval where no routine at all is given. Identical routines aren't problematic only second case.

If no routine is given for a time interval npc continues with previous routine but when routine is reset npc becomes stuck. In this case a default routine is used.

In vanilla npc keeps using previous routine after a reset like conversion or if loaded out of world. Only `set time` command uses routine from next time interval. I opted for previous because this is used during normal play.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -461,32 +461,8 @@ bool Npc::resetPositionToTA() {
 
   auto& rot = currentRoutine();
   auto  at  = rot.point;
-
-  if(at==nullptr) {
-    const auto time  = owner.time().timeInDay();
-    const auto day   = gtime(24,0).toInt();
-    int64_t    delta = std::numeric_limits<int64_t>::max();
-
-    // closest time-point
-    for(auto& i:routines) {
-      int64_t d=0;
-      if(i.start<i.end) {
-        d = time.toInt()-i.start.toInt();
-        } else {
-        d = time.toInt()-i.end.toInt();
-        }
-      if(d<=0)
-        d+=day;
-
-      if(i.point && d<delta) {
-        at    = i.point;
-        delta = d;
-        }
-      }
-
-    if(at==nullptr)
-      return false;
-    }
+  if(at==nullptr)
+    return false;
 
   if(at->isLocked() && !isDead){
     auto p = owner.findNextPoint(*at);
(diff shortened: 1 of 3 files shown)


game/world/objects/npc.cpp
```diff
@@ -2830,6 +2830,24 @@ const Npc::Routine& Npc::currentRoutine() const {
       return i;
     }
 
+  // take previous routine if none is set
+  const auto     day   = gtime(24,0).toInt();
```
Try, line 2834: This code probably duplicated one from `resetPositionToTA`
```
  if(at==nullptr) {
    const auto time  = owner.time().timeInDay();
    const auto day   = gtime(24,0).toInt();
    int64_t    delta = std::numeric_limits<int64_t>::max();

    // closest time-point
    for(auto& i:routines) {
   ...
```
Need to generalize, and add handling of null-waypooint here

Try:

Need to generalize `currentRoutine` a bit

thokkat:
I removed the duplicated code from `resetPositionToTA` so `currentRoutine` now always returns a valid routine except if there's none. Added some error logging if none is found.
There was a conflict because `endTime` checked if current time is within current routine's time window. Since `endTime` is only used with the current routine as argument and `currentRoutine` already checks the time window I removed the check here.

game/world/objects/npc.cpp
```diff
@@ -2823,13 +2799,40 @@ void Npc::commitSpell() {
...
       return i;
     if(i.start<=time && time<i.end)
       return i;
     }
 
+  static std::set<int32_t> s;
+  if(routines.size()>0 && s.find(hnpc->id)==s.end()) {
+    s.insert(hnpc->id);
+    auto sym = owner.script().findSymbol(*hnpc);
+    Log::e("no valid routine found [",sym->name(),", time: ",time.hour(),":",time.minute(),"]");
```
Try, line 2815: Any practical reason of `Log::e` here? Invalid routines are happening in normal play, and logging it as error will confuse someone who reads the log.
Probably can skip logging at all

thokkat: Removed the logging part.
>Any practical reason of Log::e here?

 Npc becomes stuck if invalid routine is queried and no fallback is present. I'd call accessing invalid routine an error. If error is the wrong word could call it warning like phoenix has. Was merely meant to see where buggy routines are because you don't notice them ingame now.

Try:
Merged, thanks!

PR #507, opened 2023-08-12
Correctly end interaction with `set time` command

thokkat:
If an npc has an interact out animation `attachMode` is already false but `currentInteract` is not yet null. If `set time` is used interaction can't be ended because `state` is now -1 instead of 0, leading to stuck npc. 
This is fixed by including -1 in the quit check. There's still a delay befor npc is moving again caused by `waitAnim` but good enough for  practical purposes.

### DIFF
--- a/game/world/objects/interactive.cpp
+++ b/game/world/objects/interactive.cpp
@@ -330,7 +330,7 @@ void Interactive::implTick(Pos& p) {
       }
     }
 
-  if((state==0 && !attach) || (isLadder() && (attach && state>=stepsCount-1))) {
+  if((state<=0 && !attach) || (isLadder() && (attach && state>=stepsCount-1))) {
     implQuitInteract(p);
     return;
     }


Try:
Hi, @thokkat !

Commit itself looks good, yet there is more issues with `set time`, related to MOBSI:
if npc was reset at item-interaction (stove cooking, as example), then dummy item (pan with meat) wont be removed from inventory.
I can look into that eventually, or if you have time - be my guest :)

thokkat:
I tried this in `Npc::resetPositionToTA` :
```c++
- invent.clearSlot(*this,"",false);
  if(routines.size()==0)
    return true;

  attachToPoint(nullptr);
  if(!isPlayer())
-   setInteraction(nullptr,true);
+   setInteraction(nullptr);
  clearAiQueue();
```
Pan appears for a short moment in hand but is then correctly removed. `clearSlot` could be executed only if interactive is null.

Try:
> Pan appears for a short moment in hand but is then correctly removed. clearSlot could be executed only if interactive is null.

OK, fixed that in 3c8bf7b1, so PR can be merged as-is. Thanks, for PR! :)

PR #526, opened 2023-10-22
feat: deploy in extra job that runs when all builds are finished

fhanrath:
I read through the [AppVeyor documentation](https://www.appveyor.com/docs), specifically https://www.appveyor.com/docs/job-workflows/, and I think this could be a solution how to provide binaries for every platform on release.

I added the Build Jobs to their own group and only let those Jobs run the build scripts. There is a new artifact for the Win VS2022 configuration, so that is doesn't overwrite the VS2019 one.

Only Mac, Linux and VS2019 will be deployed by a new deploy job, which runs after the build jobs terminate (and only on tagged commits like before)


fhanrath:
Squashing and reopening...

PR #527, opened 2023-10-22
feat: deploy in extra job which runs when all builds are finished

fhanrath:
I read through the [AppVeyor documentation](https://www.appveyor.com/docs), specifically https://www.appveyor.com/docs/job-workflows/, and I think this could be a solution how to provide binaries for every platform on release.

I added the Build Jobs to their own group and only let those Jobs run the build scripts. There is a new artifact for the Win VS2022 configuration, so that is doesn't overwrite the VS2019 one.

Only Mac, Linux and VS2019 will be deployed by a new deploy job, which runs after the build jobs terminate (and only on tagged commits like before)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -153,7 +153,7 @@ for:
       - path: build/opengothic/opengothic_win.zip
         name: archive_win
       - path: build/opengothic/opengothic_linux.zip
-        name: archive_linux
+        name: _linux
       - path: build/opengothic/opengothic_osx.zip
         name: archive_osx
   
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ environment:
       job_group: Builds
       WIN_RELEASE: true
 
-    - APPVEYOR_BUILD_WORKER_IMAGE: macos-bigsur
+    - APPVEYOR_BUILD_WORKER_IMAGE: macos-monterey
       GENERATOR: "Ninja"
       job_group: Builds
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,147 +7,175 @@ environment:
       GENERATOR: "Ninja"
       CC:         C:/msys64/mingw64/bin/gcc.exe
       CXX:        C:/msys64/mingw64/bin/g++.exe
+      job_group: Builds
+      WIN_RELEASE: true
 
     - APPVEYOR_BUILD_WORKER_IMAGE: macos-bigsur
       GENERATOR: "Ninja"
+      job_group: Builds
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
       GENERATOR: "Ninja"
       VCVARSALL: "C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Auxiliary/Build/vcvarsall.bat"
       PLATFORM:  x64
+      job_group: Builds
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2004
       GENERATOR: ""
+      job_group: Builds
 
-install:
-- ps: >-
-    $VK_SDK = '1.3.231.1'
+    - job_name: Deploy
+      job_depends_on: Builds
 
-    if($IsLinux) {
-      sudo apt-get update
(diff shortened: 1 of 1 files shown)


Nindaleth:
Hi @fhanrath, I noticed you've created and then closed several PRs. Unless there's a specific reason for it, you can just update the existing one (e.g. force push), there's no need to close an imperfect PR if you intend to keep working on it.

Try:
Hi @fhanrath and thanks for PR!

I need to test, if artifact that we produce for some build are actually viable (not sure about macos in particular to do signing), will merge it this or next week

Try:
Tested macos binary: it has no digital signature and OS wont allow to open it, so no reason to publish it.
No sure about Linux: can someone confirm, if Linux binary can be downloaded and used, or it's same as on mac?

Nindaleth:
[This run](https://ci.appveyor.com/project/Try/opengothic/builds/48338876/job/k4kbn2mmq2jrotxe/artifacts), right?
I'm getting an occasional SIGKILL when starting the artifact binary; not always but sometimes, I'm not yet sure why.

Try:
>  I'm not yet sure why.

Better to check latest most run - there was out-of-bounds bug, that been fixed. Also crash-log is welcome.
If it's just some random bug, not related to how we build linux binary - than we can publish linux builds, just fine

fhanrath:
I tested the [MacOS Binary](https://ci.appveyor.com/project/Try/opengothic/build/job/dhdhca9xiulji2dr/artifacts) on my Mac (M1, MacOS 13.6) as well.
At first, it complains about `Gothic2Notr.sh` and `libTempest.dylib` not being signed.
But when you go to system settings - Privacy & Security, you're able to allow both.

After I've done that OpenGothic starts with no issues.

In my mind providing unsigned binaries is better than providing no binaries for MacOS at all. 

I don't have a paid Apple Developer Account yet, but I'd be willing to change that and provide you with signed binaries if you're interested.
My suggestion would be to do that in another PR, especially since I'll have to find out how the signing of binaries works and how I'm able to integrate this into the pipeline first.

Nindaleth:
> Better to check latest most run - there was out-of-bounds bug, that been fixed. Also crash-log is welcome.

The issue's still present on a very recent build, it happens more often than it doesn't. I get SIGKILL almost immediately after running, the game doesn't even reach main menu, there's just a little bit of log and for some reason even when running under gdb, there's no backtrace to see, nothing in the crashlog either. My own local builds never have this happen.

```
$ ./Gothic2Notr.sh -g ~/Games/gothic-ii-gold-edition/drive_c/GOG\ Games/Gothic\ 2\ Gold/
OpenGothic v
GPU = AMD Radeon RX 6700 XT (RADV NAVI22)
Depth format = Depth32F Shadow format = Depth16
Killed
```

```
$ DEBUGGER=gdb ./Gothic2Notr.sh -g ~/Games/gothic-ii-gold-edition/drive_c/GOG\ Games/Gothic\ 2\ Gold/ -v
(...)
(gdb) r
Starting program: /home/black_fox/Downloads/opengothic_linux/Gothic2Notr -g /home/black_fox/Games/gothic-ii-gold-edition/drive_c/GOG\ Games/Gothic\ 2\ Gold/ -v
(...)
Program terminated with signal SIGKILL, Killed.
The program no longer exists.
(...)
(gdb) bt
No stack.
(gdb)
```

I can't seem to get anything from `auditctl` (so that I could look at what does the killing), strace reports `futex(0x7ffed6fbc9f0, FUTEX_WAIT_PRIVATE, 2, NULL) = ?` as the very last thing before getting the SIGKILL.

Try:
@fhanrath 

> I don't have a paid Apple Developer Account yet, but I'd be willing to change that and provide you with signed binaries if you're interested.

Thanks for suggestion, but I can have account - should be an issue :)
Just need to find time to figure how-to setup CI for MacOS-bundle builds

Try:
Merged, thanks!

Try:
Unfortunately, `Deploy` proposed in this PR didn't work as expected. After much of back and forward has to return to almost pre-merge state of CI.

To publish linux/mac artifact make all 3 jobs as releasable: first one creates github-release draft and others do append artifact on top:
изображение

PR #556, opened 2023-12-30
Adjust fall camera

thokkat:
Vanilla testing showed that `fallCam`is only activated once the deep fall animation is used. This is useful because otherwise the camera was not reset to the normal behind the player perspective when F8 flying around.

The second change makes sure bird perspective is left on landing.


Try:

LGTM, merged.

PR #571, opened 2024-01-24
fixed camera zooming out on repeated 0 value events

ezamelczyk:
On macOS using trackpad system sends a lot of events with deltaY=0 and made it impossible to zoom back in after zooming out.


Try:
Hi, @ezamelczyk and thanks for PR!

I don't think it makes any sense to emit event for `deltaY==0` in first place - this fix should be moved to `Tempest::EventDispatcher`

Try:
fixed on Tempest level

PR #572, opened 2024-01-25
FXAA integration

KirillAlekseeenko:
Integrated FXAA as a relatively simple spatial antialiasing algorithm at the end of the frame just before UI rendering

There are multiple presets:
0 -- no FXAA
from 1 to 5 -- the higher this number the more expensive filtering is used

To set FXAA preset add "-fxaa preset_num" parameter to the command line arguments


Try:
Hi, @KirillAlekseeenko and thanks for PR!

Currently busy with gpu-driven, - will look closely into PR afterward.

Some notes in a meantime:
* can you please provide a screen-shoots showcasing visual changes?
* how much performance FXAA takes?
* please provide a link to source of `Fxaa3_11.h` - for sake of bookkeeping. 

And one more thing to note: there was a debate in last release comments about better AA technique, so it best next time to raise ticket and ask. On my end I was looking toward programmable-resolve based AA (can do on Apple, but not yet on Vulkan).
It is OK with you, if after 6-12 months this FXAA will be replaced by else algorithm?

KirillAlekseeenko:
Hi!
Accroding to nvidia nsight frame capture FXAA takes from 0.06 to 0.08 ms (depends on quality preset) on nvidia 3070 card with full hd resolution.
Source for Fxaa3_11.h can be found here: https://gist.github.com/jwoolston/c96c22473110c9924d831cebb5f2cdb4
Originally it's a part of NVIDIA samples:
https://docs.nvidia.com/gameworks/content/gameworkslibrary/graphicssamples/opengl_samples/fxaa.htm
https://docs.nvidia.com/gameworks/content/gameworkslibrary/graphicssamples/d3d_samples/fxaa311sample.htm

It's ok if FXAA will be removed in future releases, it's just a starting AA solution and not good comparing to alternatives, but it's cheap and simple though. What did you mean by programmable-resolve based AA?

There are shots before and after applying FXAA
5_afterFXAA
5_beforeFXAA
6_afterFXAA
6_beforeFXAA
1_afterFXAA
1_beforeFXAA
2_afterFXAA
2_beforeFXAA
3_afterFXAA
3_beforeFXAA
4_afterFXAA
4_beforeFXAA

Try:
> What did you mean by programmable-resolve based AA?

This is though direction, rather a complete technique at this point. Basic idea here is to use (or abuse :D ) MSAA rendering with some extensions to get full-res gBuffer + 4x-coverage buffer.  This method originally was intended only to deal with sky-to-object aliasing (trees/ship) - coverage buffer is perfect to compose sky on top of scene.

Naturally there are quite a few challenges to make it work, and one of then is to be able to resolve MSAA image into single-sampled image from shader, rather then from fixed-function hardware.

game/graphics/renderer.cpp
```diff
@@ -81,6 +81,7 @@ void Renderer::resetSwapchain() {
   smpN.setClamping(ClampMode::ClampToEdge);
 
   sceneLinear    = device.attachment(TextureFormat::R11G11B10UF,w,h);
+  sceneTonemapped = device.attachment(TextureFormat::RGBA8, w, h);
```
Try, line 84: RenderTargets tend to take lot of gpu-memory - need to avoid allocation, if FXAA is not enabled

shader/antialiasing/fxaa.frag
```diff
@@ -0,0 +1,87 @@
...
+
+layout(push_constant, std140) uniform PushConstantsFxaa {
+  float fxaaInverseSharpnessCoeff;
+  float fxaaQualitySubpix;
+  float fxaaQualityEdgeThreshold;
+  float fxaaQualityEdgeThresholdMin;
+  float fxaaConsoleEdgeSharpness;
+  float fxaaConsoleEdgeThreshold;
+  float fxaaConsoleEdgeThresholdMin;
+  } pushConstantsFxaa;
```
Try, line 34: nitpick:
It's not required to give a name to push-block. Can be nameless and have cleaner shader code

shader/antialiasing/fxaa.frag
```diff
@@ -0,0 +1,87 @@
...
+
+layout(binding  = 0) uniform sampler2D aliasedInput;
+
+layout(location = 0) in  vec2 uv;
```
Try, line 38: Should be a `gl_FragCoord.xy / screenSize`. Asking, as there is an intent to get rid of `-DHAS_UV` option in post-process and make it more compatible with compute

game/graphics/renderer.cpp
```diff
@@ -168,6 +169,10 @@ void Renderer::resetSwapchain() {
   tonemapping.pso     = &Shaders::inst().tonemapping;
   tonemapping.uboTone = device.descriptors(*tonemapping.pso);
 
+  uint32_t availableFxaaPreset = std::clamp(Gothic::options().fxaaPreset, 0u, static_cast<uint32_t>(Shaders::FxaaPreset::LAST) - 1);
```
Try, line 172: `options().fxaaPreset` should be sanitized when parsing command-line, not at runtime.

game/graphics/shaders.h
```diff
@@ -50,6 +50,17 @@ class Shaders {
 
     Tempest::RenderPipeline  tonemapping;
 
+    enum class FxaaPreset : int {
```
Try, line 53: Is there a need if `class`? In other part of code, preset seem to be used almost always as integer, with static_cast

game/graphics/renderer.cpp
```diff
@@ -550,10 +562,23 @@ void Renderer::draw(Tempest::Attachment& result, Encoder<CommandBuffer>& cmd, ui
...
+  if (isFxaaOn) {
+  	tonemappingRt = &sceneTonemapped;
+    }
+
+  cmd.setFramebuffer({{*tonemappingRt, Tempest::Discard, Tempest::Preserve}});
   cmd.setDebugMarker("Tonemapping");
   drawTonemapping(cmd);
 
+  if (isFxaaOn) {
+  	cmd.setFramebuffer({ {result, Tempest::Discard, Tempest::Preserve} });
```
Try, line 577: code-style: indentation in project is 2-whitespaces

game/graphics/renderer.h
```diff
@@ -89,6 +90,8 @@ class Renderer final {
     Tempest::ZBuffer          zbuffer, shadowMap[Resources::ShadowLayers];
     Tempest::ZBuffer          zbufferUi;
 
+    Tempest::Attachment       sceneTonemapped;
```
Try, line 93: can be moved into `fxaa` block, since at this point only FXAA uses this attachment

game/commandline.h
```diff
@@ -23,7 +23,7 @@ class CommandLine {
       Vulkan,
       DirectX12
       };
-    auto                graphicsApi() const -> GraphicBackend;
+    auto                graphicsApi() const -> GraphicBackend;
```
Try, line 26: nitpick: whitespace

game/commandline.h
```diff
@@ -43,6 +43,7 @@ class CommandLine {
     bool                doForceG1()        const { return forceG1;      }
     bool                doForceG2()        const { return forceG2;      }
     bool                doForceG2NR()      const { return forceG2NR;    }
+    uint32_t            getFxaaPreset()    const { return fxaaPreset;   }
```
Try, line 46: Project doesn't use `get*` notation (except really old code, that hasn't been cleaned yet) - should be just `fxaaPreset()`
Names for private variable are not strict, can be `uint32_t fxaaPresetId = 0`

game/graphics/renderer.cpp
KirillAlekseeenko: added optional creation

game/graphics/shaders.h
KirillAlekseeenko: Used enum for readability, seems better than plain integers in the code section where fxaa's pso array is filled. 
Removed redundant prefix "FXAA_"

Try:
There some code-style issues - I'll fix it afterward.
Merged, thanks!

PR #574, opened 2024-01-31
Interactive focus: consider `focOver` attribute only if a weapon is drawn

thokkat:
`fovOver` is only used in combat mode otherwise normal interaction is always possible. I tested at Irdorath bridge switches. 

Reason for pr is a mod where a chest had this set to true, likely by accident.


Try:
LGTM, will merge after GDR stuff

Try (approved):

Try:
merged, thanks!

Issue #182, opened 2021-07-29
Collision model


Try:
Gothic 2 collision are pretty ticky - need to investigate.

#### Collision `npc`(human) vs `landscape`
1. Frontal
![изображение](https://user-images.githubusercontent.com/122894/127567049-d1f1a7c0-bf14-478c-ad1f-f4f56ccd60dd.png)
2. Side
![изображение](https://user-images.githubusercontent.com/122894/127566295-5f507794-2c3c-4113-a6a2-b871f5b3b9d5.png)
3. Rotate after side collision
![изображение](https://user-images.githubusercontent.com/122894/127566988-7220aeba-4aff-4a49-8892-f0ff1b2829bf.png)

#### Collision `npc`(snapper) vs `landscape`
1. Frontal
![изображение](https://user-images.githubusercontent.com/122894/127567154-b95698ea-7f76-4ba8-a508-3c139a2b49ce.png)
2. Side
![изображение](https://user-images.githubusercontent.com/122894/127567218-61167d2b-001b-4020-9020-855f61e2722c.png)

Try:
From https://github.com/Try/OpenGothic/pull/575:

Game had some issues with npc-origin point and it interaction with collision. Probably it's time to get rid of `translateY` and related hacks and make OpenGothic more like vanilla on this aspect.

Try:
Some testing of collisions:

<img width="591" height="707" alt="Image" src="https://github.com/user-attachments/assets/a0b21379-fbb9-4c92-86df-63e11e5eac09" /> 

<img width="591" height="707" alt="Image" src="https://github.com/user-attachments/assets/957a76ac-a0b4-4f35-93a4-1281cb1c6b4b" />

<img width="591" height="707" alt="Image" src="https://github.com/user-attachments/assets/c3c51da2-531d-4a41-befd-91e030674a4c" />

Running asymmetrical npc, wolf in that case, into a wall resulting at one type of collision.
Side-walk and then rotation - allows to dip into a wall.

After dip into a wall: no way to move forward, but side-move will push npc back from the wall.
Makes me think, that this can be not a 'real' collider proxy, but just a set of six rays to estimate distances.

thokkat:
Tested Karibik mod again and lava is not killing the player (save1). From https://github.com/Try/OpenGothic/pull/575#issuecomment-1977418490 

> mod:
> https://www.worldofgothic.de/dl/download_404.htm
> I used 1.1 patch. Second link are voice lines (not needed).
[...]

PR #575, opened 2024-01-31
Add firedelay to triggers

thokkat:
Trigger attribute `fireDelay` determines how much time should pass between trigger call and start. Currently only the time between calls is checked leading to erroneous behavior, e.g. the first call never waits.

### DIFF
--- a/game/world/triggers/abstracttrigger.cpp
+++ b/game/world/triggers/abstracttrigger.cpp
@@ -32,7 +32,8 @@ AbstractTrigger::AbstractTrigger(Vob* parent, World &world, const zenkit::Virtua
      data.type == VirtualObjectType::oCTriggerScript      || data.type == VirtualObjectType::zCMover ||
      data.type == VirtualObjectType::oCTriggerChangeLevel || data.type == VirtualObjectType::oCCSTrigger) {
     auto& trigger = reinterpret_cast<const zenkit::VTrigger&>(data);
-    fireDelaySec = trigger.fire_delay_sec;
+    fireDelaySec = uint64_t(trigger.fire_delay_sec*1000.f);
+    reTriggerDelaySec = uint64_t(trigger.retrigger_delay_sec*1000.f);
     maxActivationCount = uint32_t(trigger.max_activation_count);
     filterFlags = trigger.filter_flags;
     triggerFlags = trigger.flags;
@@ -52,13 +53,26 @@ bool AbstractTrigger::isEnabled() const {
   return !disabled;
   }
 
+bool AbstractTrigger::isDeferred() const {
+  return evtDeferred.timeBarrier>world.tickCount();
+  }
+
 void AbstractTrigger::processEvent(const TriggerEvent& evt) {
-  if(emitTimeLast>0 && world.tickCount()<emitTimeLast+uint64_t(fireDelaySec*1000.f)) {
-    world.triggerEvent(evt);
+  if(isDeferred())
+    return;
+  if(emitTimeLast>0 && world.tickCount()<emitTimeLast+reTriggerDelaySec)
+    return;
+  if(fireDelaySec>0) {
+    evtDeferred = evt;
+    evtDeferred.timeBarrier += fireDelaySec;
+    triggerDeferred();
     return;
     }
-  emitTimeLast = world.tickCount();
+  execEvent(evt);
+}
 
+void AbstractTrigger::execEvent(const TriggerEvent& evt) {
+  emitTimeLast = world.tickCount();
   switch(evt.type) {
     case TriggerEvent::T_Startup:
     case TriggerEvent::T_StartupFirstTime:
@@ -138,12 +152,20 @@ void AbstractTrigger::onIntersect(Npc& n) {
 
   if(boxNpc.intersections().size()==1) {
     // enableTicks();
-    TriggerEvent e("","",TriggerEvent::T_Activate);
+    TriggerEvent e(target,"",world.tickCount(),TriggerEvent::T_Activate);
     processEvent(e);
     }
   }
 
 void AbstractTrigger::tick(uint64_t) {
+}
+
+void AbstractTrigger::tickDeferred(uint64_t) {
+  if(isDeferred()) {
+    triggerDeferred();
+    return;
+    }
+  execEvent(evtDeferred);
   }
 
 bool AbstractTrigger::hasVolume() const {
@@ -163,6 +185,7 @@ void AbstractTrigger::save(Serialize& fout) const {
   boxNpc.save(fout);
   fout.write(emitCount,disabled);
   fout.write(emitTimeLast);
+  evtDeferred.save(fout);
   }
 
 void AbstractTrigger::load(Serialize& fin) {
@@ -170,6 +193,11 @@ void AbstractTrigger::load(Serialize& fin) {
   boxNpc.load(fin);
   fin.read(emitCount,disabled);
   fin.read(emitTimeLast);
+  if(fin.version()>46) {
+    evtDeferred.load(fin);
+    if(evtDeferred.timeBarrier>world.tickCount())
+      triggerDeferred();
+    }
   }
 
 void AbstractTrigger::enableTicks() {
@@ -178,6 +206,10 @@ void AbstractTrigger::enableTicks() {
 
 void AbstractTrigger::disableTicks() {
   world.disableTicks(*this);
+  }
+
+void AbstractTrigger::triggerDeferred() {
+  world.triggerDeferred(*this);
   }
 
 const std::vector<Npc*>& AbstractTrigger::intersections() const {
--- a/game/world/triggers/abstracttrigger.h
+++ b/game/world/triggers/abstracttrigger.h
@@ -50,10 +50,12 @@ class AbstractTrigger : public Vob {
 
     std::string_view             name() const;
     bool                         isEnabled() const;
-
+    bool                         isDeferred() const;
     void                         processEvent(const TriggerEvent& evt);
+    void                         execEvent(const TriggerEvent& evt);
     virtual void                 onIntersect(Npc& n);
     virtual void                 tick(uint64_t dt);
+    void                         tickDeferred(uint64_t dt);
 
     virtual bool                 hasVolume() const;
     bool                         checkPos(const Tempest::Vec3& pos) const;
@@ -89,16 +91,20 @@ class AbstractTrigger : public Vob {
     void                         disableTicks();
     const std::vector<Npc*>&     intersections() const;
 
+    void                         triggerDeferred();
+
   private:
     Cb                           callback;
     DynamicWorld::BBoxBody       box;
     CollisionZone                boxNpc;
     Tempest::Vec3                bboxSize, bboxOrigin;
 
-    float                        fireDelaySec = 0;
+    uint64_t                     fireDelaySec = 0;
+    uint64_t                     reTriggerDelaySec = 0;
     uint32_t                     maxActivationCount = 0;
     uint32_t                     triggerFlags = 0;
     uint32_t                     filterFlags = 0;
+    TriggerEvent                 evtDeferred = {};
 
     uint32_t                     emitCount = 0;
     bool                         disabled  = false;
--- a/game/world/worldobjects.cpp
+++ b/game/world/worldobjects.cpp
@@ -186,6 +186,11 @@ void WorldObjects::tick(uint64_t dt, uint64_t dtPlayer) {
   for(auto i:triggersTk)
     i->tick(dt);
 
+  auto triggers = std::move(triggersDeferredTk);
+  triggersDeferredTk.clear();
+  for(auto i:triggers)
+    i->tickDeferred(dt);
+
   bullets.remove_if([](Bullet& b){
     return b.isFinished();
     });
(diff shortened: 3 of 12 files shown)


Try:
Hi, @thokkat and thanks for finding!

> Currently only the time between calls is checked leading to erroneous behavior, e.g. the first call never waits

Fist call has wait in current implementation, since time start at `0` and `emitTimeLast` is initialized to zero. In what case trigger doesn't work for you?
I'm assuming that it break when player enters a new non-starting location, for first time, so it should be enough to set 
`emitTimeLast = world.tickCount();` in constructor.

thokkat:
Currently the last successful activation time is saved and next activation is delayed. So if time distance is large enough trigger would always activate instantly. Instead I want to change it so it always delays before activation regardless when last activation was. Took it from spacer help [file](https://wiki.worldofgothic.de/doku.php?id=spacer:hilfedatei):

```
[TOPIC MEMBER 'zCTrigger::fireDelaySec']
FLOAT                sec
Wieviel Zeit nach dem Aktivieren des Triggers muss vergehen, bevor er wirklich feuert. 
\\ How much time must pass after activating the trigger before it actually fires.
```


> In what case trigger doesn't work for you?

[Here](https://www.youtube.com/watch?v=WcSx3bwEhso&list=PLQYAznaaoajYpWOzTE9jyykzKPbvoYlbn&t=1132s)  in vanilla cannons pop up over time but in OG all are set up at the beginning of cutscene.

script:

```c
    wld_sendtrigger("KAMERA_KANONENEVENT_KANONEN_TRIGGER");
    ai_setwalkmode(self, npc_run);
    ai_gotowp(self, "PINTA_DSCHUNGEL_40");
    ai_playani(self, "T_PLUNDER");
    wld_sendtrigger("PINTA_KANONE_01_TRIGGER");
    ai_gotowp(self, "PINTA_DSCHUNGEL_41");
    ai_playani(self, "T_PLUNDER");
    wld_sendtrigger("PINTA_KANONE_02_TRIGGER");
    ai_gotowp(self, "PINTA_DSCHUNGEL_42");
    ai_playani(self, "T_PLUNDER");
    wld_sendtrigger("PINTA_KANONE_03_TRIGGER");
    ...
```

Cannon triggers have `fireDelaySec` values set to align with plunder animation.

thokkat:
Ok actually `emitTImeLast` refers to `retriggerWaitSec` instead of `fireDelaySec`. Adjusted accordingly.

https://zk.gothickit.dev/engine/objects/zCTrigger/

> * retriggerWaitSec
>     The number of seconds that have to elapse after processing an event before the trigger will process additional events. All events received by the trigger during that time are ignored.
> * fireDelaySec
>   The number of seconds to wait before emitting the OnTrigger event after processing.

game/world/triggers/abstracttrigger.h
```diff
@@ -96,13 +96,15 @@ class AbstractTrigger : public Vob {
     Tempest::Vec3                bboxSize, bboxOrigin;
 
     float                        fireDelaySec = 0;
+    float                        reTriggerDelaySec = 0;
```
Try, line 99: should `reTriggerDelaySec` be stored in uint64_t, in milliseconds?

game/world/triggers/abstracttrigger.cpp
```diff
@@ -53,10 +54,16 @@ bool AbstractTrigger::isEnabled() const {
   }
 
 void AbstractTrigger::processEvent(const TriggerEvent& evt) {
-  if(emitTimeLast>0 && world.tickCount()<emitTimeLast+uint64_t(fireDelaySec*1000.f)) {
+  if(emitTimeLast>0 && world.tickCount()<emitTimeLast+uint64_t(reTriggerDelaySec*1000.f))
+    return;
+  if(fireWaitTime==0)
```
Try, line 59: how it should work, if for example `retriggerWaitSec==1`, and `fireDelaySec==10` and triger been activated multiple times, before firing? Should we have multiple events in flight to cover that case?

`TriggerEvent::timeBarrier` seem to be a good candidate to implement fire-delay mechanics

thokkat: > TriggerEvent::timeBarrier seem to be a good candidate 

`if(t.name()==e.target) { // NOTE: trigger name is not unique - more then one trigger can be activated`

I played around a bit but this gives me a headache. To prevent infinite trigger cycles and assign wait time `triggerevent` has to reference trigger directly instead of name. This would require rewrite of constructor and adjust fucntion calls that pass trigger events. Just asking if you see a simpler answer or is this the way to go?

thokkat:
Added a pointer to reference trigger for Triggerevents to assure uniqueness. Pointer is not serialized, so in case of a loaded game Triggerevents only have the string reference like before and pick up pointer before first testing for `firedelay`. But this should be extremely rare and thus ok I guess.

game/world/triggers/abstracttrigger.h
```diff
@@ -41,6 +42,7 @@ class TriggerEvent final {
       zenkit::MoverMessageType msg = zenkit::MoverMessageType::FIXED_DIRECT;
       int32_t                  key = 0;
       } move;
+    mutable AbstractTrigger* trigger = nullptr;
```
Try, line 45: OK, I see what you meant by "has to reference trigger directly instead of name".
`mutable` is definitely not a keyword we would like to use in game code.

One solution I've considered, is to track deferred trigger-events inside `AbstractTrigger`, and enable ticks as long as that list is non-empty

game/world/worldobjects.cpp
```diff
@@ -408,10 +408,15 @@ void WorldObjects::triggerEvent(const TriggerEvent &e) {
...
 bool WorldObjects::execTriggerEvent(const TriggerEvent& e) {
+  if(e.trigger!=nullptr) {
+    e.trigger->processEvent(e);
+    return true;
+    }
   bool emitted=false;
   for(auto& i:triggers) {
     auto& t = *i;
     if(t.name()==e.target) { // NOTE: trigger name is not unique - more then one trigger can be activated
+      e.trigger = i;
```
Try, line 419: This doesn't seem to be correct: when `t.name()` matches multiple receivers, all of them should receive event.
After this change they will but only one of them will populate `e.trigger`.

thokkat:
Deferred triggers are now tracked in AbstractTrigger class as you suggested. I moved events with a delay applied in triggerlist there as well to be able to assure uniqueness. In case of user save game action events are pushed back to `triggerEvents` vector which keeps current save game behavior.

game/world/triggers/abstracttrigger.h
```diff
@@ -90,12 +93,20 @@ class AbstractTrigger : public Vob {
     const std::vector<Npc*>&     intersections() const;
 
   private:
+    struct DeferredTrigger {
+      TriggerEvent     evt     = {};
+      AbstractTrigger* trigger = nullptr;
+      };
+
+    static std::vector<DeferredTrigger> triggersDeferred;
```
Try, line 101: `static` makes no sense for intended purpose

game/world/triggers/abstracttrigger.cpp
```diff
@@ -138,12 +153,36 @@ void AbstractTrigger::onIntersect(Npc& n) {
...
   }
 
 void AbstractTrigger::tick(uint64_t) {
+}
+
+void AbstractTrigger::tickWait(uint64_t time) {
+  if(triggersDeferred.empty())
+    return;
+
+  auto it = std::stable_partition(triggersDeferred.begin(),triggersDeferred.end(),[time](const DeferredTrigger& d) {
```
Try, line 168: > std::stable_partition
> 1) Exactly N applications of the predicate and O(N) swaps if there is enough extra memory. 

It worse than brute force

game/world/triggers/abstracttrigger.cpp
```diff
@@ -165,6 +204,11 @@ void AbstractTrigger::save(Serialize& fout) const {
   fout.write(emitTimeLast);
   }
 
+void AbstractTrigger::saveDeferred(World& world) {
+  for(auto& i:triggersDeferred)
+   world.triggerEvent(i.evt);
```
Try, line 209: method `saveDeferred` doesn't save anything

thokkat:
Since `delay` from triggerlist is per string target but `fireDelay` is per trigger target I think it's best to keep it as is and push delayed triggers back in `World::execTriggerEventis`. Only a check for target uniqueness has been added. Triggers with a `fireDelay` are then pushed to `triggersDeferred` and wait until execution time has come.

Try:
@thokkat 
After reading change multiple time is still look over-engineered and incomplete. Can you please send me a link to mod+save that you using to test this feature?

thokkat:
Short explanation how I think it should be:
`wait triggerlist delay -> check if trigger is already allowed (retriggerWaitSec), if not discard -> wait fireDelay -> fire`
If trigger is on any delay don't allow retrigger. But tested only fireDelay here.

mod:
https://www.worldofgothic.de/dl/download_404.htm
I used 1.1 patch. Second link are voice lines (not needed).

save files:
[save_slot_.zip](https://github.com/Try/OpenGothic/files/14486773/save_slot_.zip)

In quick save talk to npc and use dialog option 1.  

There are more bugs in case you wanna look into them as well. 
- After CsCamera in quicksave another CsCamera is started. Cannons should fire but don't. 

- Save1 has lava which should kill the player. Problem is
https://github.com/Try/OpenGothic/blob/757ee90abbdd02494e7ec68c32b9c552e0db6061/game/world/worldobjects.cpp#L391
y value of added vector needs to be smaller than `i->translateY()*0.2`.

- Save2 has a cannon as item which can't be focused. Cause is a [change](https://github.com/Try/OpenGothic/commit/323a6e5adb286c5c04c8fc1b77bdc93c4310a50f#diff-16c5d86ed9e9c2bfb96675176468b87752ae6a26b8a277d51fc3d435bd650585) in recent graphics update that changes item mid position.

Try:
Save files are not compatible apparently:
```
unable to open Zen-file: "insel3.zen"
loading error: Cannot init RTN_START_415: not an instance
```
I'm assuming that problem is me installing mod on top of English game. You have steam+German as a base, right?

thokkat:
Old Gothic II Gold cd version which has only german language available. When I shared non mod save some time ago we had same problem, hoped it wouldn't occur again.

For non delay bugs I can give some marvin goto commands, but delay requires some quests/dialogs before so npc is at right position and has trigger dialog available.

thokkat:
This is how you can get to cutscene by "normal" playing:

1.  In `Karibik.ini` change start world to `World=Karibik\INSEL3.ZEN`
2. We want to use storyhelper (has talk perc but doesn't talk) and have access to all dialog options:

    In `GameScript::npc_knowsinfo` return always true and in `Npc::startDialog` add

```
  if(hnpc->id==3100) {
    setOther(&pl);
    aiPush(AiQueue::aiProcessInfo(pl));
    return;
    }
```

3. Start game. End first dialog and go to jungle in the center of island. Clean everything e.g. with firerain (`itsc_firerain`) so npc later don't get distracted.
4. Set breakpoint in `Marvin::printVariable` , use marvin `p pinta_pflanzen_onetime` and overwrite variable value to 1.
5. Use marvin `insert karibik_storyhelper` and select last dialog option `Ins 4.Kapitel (Vor Kanonenevent)`. Augustus  and other npc  should now be teleported close to your location. Use marvin `set time` to force npc to their waypoints.
6. Talk to Augustus and start dialog `Es kann losgehen!`.

For other bugs: 
- No focus canon : same start world, `goto pos -1500 200 2900`.
- No trigger:  Use start world `World=Karibik\INSEL_2.ZEN` and `goto pos 180 -400 1900`

Try:
> unable to open Zen-file: "insel3.zen"
> When I shared non mod save some time ago we had same problem, hoped it wouldn't occur again.

Finally found time to look into. Also found the problem with saves: command line been missing `-game:Karibik.ini`

I've pushed my solution to deferred triggers to deferred-triggers branch, commit d3196b65

Try:
> No focus canon : same start world, goto pos -1500 200 2900

fixed it on the master branch

> Save1 has lava which should kill the player. Problem is

I would say the read problem here is misalignment on how BBox is aligned to npc-model here. In vanilla, apparently, origin point of npc is body-center. In OpenGothic origin is in between the foots. While OpenGothic solution is better on handling ground-collision and spawns, it leads to numerious hacks and issues, like `translateY`.
This can be addressed as part of https://github.com/Try/OpenGothic/issues/182

thokkat:
This is how processing should look like for a triggerlist after some more testing:

0. Triggerlist call
1. Discard if one of it's targets is still waiting for delay to pass (tested for `LP_ALL`), othewise go next
2. execute target after it's delay has passed
3. process triggers that match target name
4. discard if trigger is not ready (`retriggerDelay`), otherwise go next
5. discard if trigger already has a delayed event, otherwise go next
6. set delayed event and wait for `fireDelay`
7. release delayed event, set last trigger time and fire

> deferred-triggers branch, commit https://github.com/Try/OpenGothic/commit/d3196b652819e3585e2a33fb437502dce70af4ee

Some comments based on the above:

-  Why vector for `delayedEvents`? Trigger can only have one delayed event.

- `hasDelayerEvents()` typo

-  Events are discarded if trigger is on retriggerDelay.
```
  if(emitTimeLast>0 && world.tickCount()<emitTimeLast+retriggerDelay) {
    world.triggerEvent(evt);
    return;
    }
```
- `triggersDef` is not repopulated on load.

Try:
> Some comments based on the above

Thanks for checking!
* hasDelayerEvents - fixed
* discard - fixed
* triggersDef is repopulated. This was a followup commit, see 8ed48c13

> Why vector for delayedEvents? Trigger can only have one delayed event.

If `retriggerDelay` is smaller than `fireDelay`, then there can be multiple deferred event in-flight

thokkat:
> If retriggerDelay is smaller than fireDelay, then there can be multiple deferred event in-flight

Just tested again to be sure with a firedelay of 5s. If trigger is on firedelay and trigger is retriggered no second/third/... trigger execution happens. Only after these 5s a retrigger is possible.

Try:
Thanks for double checking! Fixed in b1cb4274

thokkat:
One last remark, `emitTimeLast = world.tickCount()` needs to be in  `implProcessEvent`. Test was 5s retriggerDelay, 5s fireDelay. Constant poking gave first execution after 5s, but all following took 10s.

I think with the above it's good for merge. Haven't done ingame testing though because after driver update `aquireNextImage` fails at game start with khr out of date.

Try:
> One last remark

Fixed; This PR is closed by: https://github.com/Try/OpenGothic/pull/592

Issue #420, opened 2023-03-06
Can't return treasure to Greg in The Man with the Eyepatch quest


DragonSWDev:
I have an issue with The Man with the Eyepatch quest. I have fresh build of OpenGothic (commit b0be11140140d90247a4330e25b50d6328aa052a) and Polish version of the Gothic 2 Night of the Raven. After meeting him in front of tavern on the road to Onar farm I went with him to the cave to get first treasure. After killing goblins and digging up treasure I've got moneybag but I'm unable to return it to Greg. He just stands in front of cave and there are no dialog options so I'm unable to progress with this quest any further.

DragonSWDev:
After further investigation I've discovered that I can progress quest by going to sleep. After waking up and going to Greg I can talk to him and give him back moneybag and progress with quest. I leave issue as open because this should appear without sleeping.

Try:
Hi, @DragonSWDev, thanks for report !

Can you share a *.sav file to help with reproduction?

DragonSWDev:
Sure thing I have save that I made after I dug up moneybag. If you require save before that please let me know, I'll do quest again and do save before that.

Aside from that I've found another issue. Still related to Greg so I think I can mention that here as well (if I need to create another issue for that let me know). When I go talk to Dexter, ask him about missing people etc. and kill him Greg should come to help me killing Dexter bandits. But he doesn't. I kill Dexter and bandits don't care and Greg is missing. He is still standing on the way to Onar farm without any dialogue options. 

Here you have both saves for that. Save for first issue is in the slot 2, save for second issue is in the slot 1.
[save_slot.zip](https://github.com/Try/OpenGothic/files/10903699/save_slot.zip)

Try:
> save2
He just stands in front of cave and there are no dialog

In script, there is a check, that player has to be close to cave `(Npc_GetDistToWP(self,"NW_BIGFARM_LAKE_CAVE_01") < 1000)` and apparently in you play he moved too far after fighting with goblins.
Sleeping fixes this, by practically resetting npc positions.

> save1
 When I go talk to Dexter after finishing Greg quest, ask him about missing people etc. and kill him Greg should come to help me killing Dexter bandits. 

Greg comes, just slooooow:
![изображение](https://user-images.githubusercontent.com/122894/223547001-2d67adb3-7760-4725-99ac-20928a1a55f4.png)

Both cases seem to be not an engine bug, but issue with original game scripts.

DragonSWDev:
[...]

PR #582, opened 2024-02-19
Workaround for npc teleport

thokkat:
In vanilla `Npc_ExchangeRoutine` teleports npc's to new routine waypoint if they're offloaded from world. Examples are Greg-Dexter event and ring of water meeting. The idea is to imitate this by resetting far npc's if their routine is exchanged.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -3919,6 +3919,10 @@ void Npc::excRoutine(size_t callback) {
   routines.clear();
   owner.script().invokeState(this,currentOther,currentVictum,callback);
   // aiState.eTime = gtime();
+
+  // imitate B_StartOtherRoutine's npc teleport behavior
+  if(aiPolicy>ProcessPolicy::AiNormal)
+    resetPositionToTA();
   }
 
 void Npc::multSpeed(float s) {


Try:
Hm, thanks for raising this issue, bug unfortunately this solution is too big of a hack, and we should not merge this PR as-is.

You checking for `ProcessPolicy::AiNormal` - that's find, but destination waypoint might be observable by player. If you check for wayPoint and for npc, it probably wont be a 'fix'.

I was considering to teleport npc forward along his way-path, while whole segment of way-path is not visible. Yet it's hard to reason about path visibility, specially from CPU side

thokkat:
> but destination waypoint might be observable by player. 

Yeah I was aware of the concerns you raised in https://github.com/Try/OpenGothic/issues/420. But I mean if this ever happens it's surely out of vanilla render distance and I'd say an ok tradeoff for getting several events working in return.

> I was considering to teleport npc forward along his way-path, while whole segment of way-path is not visible.

This sounds even more hacky :D. But as you said above adding a ray between player and waypoint makes workaround only bigger so feel free to close.
But I wonder what a solution could be. Only thing I could imagine is script patch where every call is checked and `AI_Teleport` is added if not visible by player.

Try:
starting a new issue, to discuss on how implementation may look like: https://github.com/Try/OpenGothic/issues/585

PR #589, opened 2024-03-03
Rework `canSense` and `canSee` functions

thokkat:
I reworked some functions based on testing:

- active perceptions are only perceived if npc has `SENSE_SMELL` or `SENSE_SEE` and `canSee` returns true, `SENSE_HEAR` does nothing
- passive perceptions only care for perc range and ignore senses and `canSee`
-  `canSee` only cares for senses range
- `npc_getnexttarget` doesn't need line of sight. 

~This fixes Lares not stopping if guiding the player and shattered golem not attacking~ was already fixed by https://github.com/Try/OpenGothic/commit/b4ee93d764738431bf77994b559450122a38398a.
The check for empty aiQueue causes problems with G1 percs because npcs have often some `AI_Wait` call in script loop. This has been introduced once to fix some sneaking issues. To compensate walkbit is now used instead of bodystate for sneak mode testing.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -2288,7 +2288,7 @@ bool GameScript::npc_getnexttarget(std::shared_ptr<zenkit::INpc> npcRef) {
     dist*=dist;
 
     world().detectNpc(npc->position(),float(npc->handle().senses_range),[&,npc](Npc& oth){
-      if(&oth!=npc && !oth.isDown() && oth.isEnemy(*npc) && npc->canSeeNpc(oth,true)){
+      if(&oth!=npc && !oth.isDown() && oth.isEnemy(*npc)) {
         float qd = oth.qDistTo(*npc);
         if(qd<dist){
           dist=qd;
--- a/game/world/objects/interactive.cpp
+++ b/game/world/objects/interactive.cpp
@@ -517,7 +517,7 @@ uint32_t Interactive::stateMask() const {
 bool Interactive::canSeeNpc(const Npc& npc, bool freeLos) const {
   for(auto& i:attPos){
     auto pos = nodePosition(npc,i);
-    if(npc.canSeeNpc(pos.x,pos.y,pos.z,freeLos))
+    if(npc.canSeePos(pos.x,pos.y,pos.z,freeLos))
       return true;
     }
 
@@ -528,7 +528,7 @@ bool Interactive::canSeeNpc(const Npc& npc, bool freeLos) const {
     float x = pos.x;
     float y = pos.y;
     float z = pos.z;
-    if(npc.canSeeNpc(x,y,z,freeLos))
+    if(npc.canSeePos(x,y,z,freeLos))
       return true;
     }
   return false;
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1703,7 +1703,7 @@ bool Npc::implAiFlee(uint64_t dt) {
       return false;
     if(p.underWater)
       return false;
-    if(!canSeeNpc(p.x,p.y+10,p.z,true))
+    if(!canSeePos(p.x,p.y+10,p.z,true))
       return false;
     if(wp==nullptr || oth.qDistTo(&p)>oth.qDistTo(wp))
       wp = &p;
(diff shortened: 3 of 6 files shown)


thokkat:
@Try  I edited the start post. Basically this cleans up/aligns more with vanilla and prepares introduction for percs used only in G1.

Try:
Checking the mater of PR (mostly on sleeping golem):

* golem has only a sense of smell, what indeed demands big refactoring in perception system.
* unclear intended behavior of `npc_canseenpc`/`npc_canseenpcfreelos` - golem can't see, so those functions return false making it idle. If 'see' in this context is 'sense', than golem works fine.
* Modkit documentation still defines `npc_canseenpc*` in terminology of ray-castings, even specifies angles
* OpenGothic code tends to use 'canSee' as raycast

With 'npc_canseenpc' there are 2 options:
* check not only visibility on `npc_canseenpc`, but also all other senses 
* bypass `hnpc->senses` mask

I've asked on discord about this - then will do rewrite based on it

thokkat:
Close to starting position where sheep is, if npc is spawned via insert command he wants to attack the monsters in the cave below and gets stuck. Doesn't happen before rework.

Try:
Checked it - it seems npc can 'hear' another one. Current idea on how-to solve this case is to restrict hearing sense.
In vanilla npc wont 'hear' player, unless player moves. Checking for BS_STAND seem to work

thokkat:
>In vanilla npc wont 'hear' player, unless player moves

Hmm when I gave npc's only hear sense AssessPlayer function was never triggered which made me think this sense is unused.

Try:
> Hmm when I gave npc's only hear sense AssessPlayer function was never triggered which made me think this sense is unused

AFAIR, hear is used when you entering someone else's house

thokkat:
I tested with all senses disabled and npc reactions still work when entering a house. Only disabling EnterRoom and QuietSound passive percs would make npcs ignore player.

Try:
> I tested with all senses disabled and npc reactions still work when entering a house. 

I'm de confirming: when testing on Harad's house.
Sneak-in works, as long as player in 'sneak-mode'. Going from sneak to regular walk, alarms Brian and Harad, even if player is inside inner area (bedroom?)

thokkat:
> I'm de confirming: 

Yes what I said was for normal walking, same as you describe when going from sneak to walk. But this recognition works with no senses enabled at all. That would mean passive perc activation is not connected to senses and it's still unclear if Hear sense is ever needed.

Try:
> That would mean passive perc activation is not connected to senses and it's still unclear if Hear sense is ever needed.

It was hard to test in vanilla for me; So far:
* `senses = 0` - completely disables reaction for entering the house.
* `senses = 1` - allows for Harad to notice player, but after he run into building, I've double check his `senses` mask and it was `=3`

Here I've been using console commands to edit senses, and this may affect methodology. And case with `senses = 1` - mean that at some point senses were reset to defaults by the engine

thokkat:
> Here I've been using console commands to edit senses, and this may affect methodology.

I directly edited `Perception.d` to prevent overwriting and running game with reparsed scripts.

> senses = 0 - completely disables reaction for entering the house

My test house was behind Zuris, but Harald is same. Set senses to zero in script or left it default and used marvin, reaction still works.

thokkat:
video:
https://github.com/Try/OpenGothic/assets/100468436/d82b5635-3647-44b2-a5d7-a73a0da333ed

thokkat:
video how it looks for me:

https://github.com/Try/OpenGothic/assets/100468436/2d1c819b-b7b0-4162-b3ff-2080be5a8623

Try:
> video how it looks for me:

My testing case was almost same, except I've entered building while sneaking. Once asserted that there is not early alarm - cancel sneak, and made a noise.
How this case works for you? Are they reacting anyway?

thokkat:
> How this case works for you? Are they reacting anyway?

Sometimes they react on first step, sometimes it takes more but then they react. Behavior is same  with default senses and senses=0.

In marvin g there is `P.detect` output. From testing it looks like  this shows yes if a perc is triggered with player as source. Walking only one step at a time seems to trigger AssessQuietSound only every second step which would explain the above.

Try:
Started new issue to track sound-senses: https://github.com/Try/OpenGothic/issues/604

PR #593, opened 2024-03-14
Add attributes for triggers

thokkat:
Of the two triggers that open the gate to get to the 4 rooms with switch riddles in Irdorath only one is enabled. Both are activated at the same time and their enable status is toggled afterwards. 

Setting both to enabled on first load lead to the bug that both gate up and down animations were played on first activation but none on second. This change fixes those cases. https://github.com/Try/OpenGothic/commit/20af3701522cf0141f37e8f6d1a707d0b06dc48e had a fix for first case but second one was still open.

Also added a missing clear command reported by static analyzer for deferred triggers.

### DIFF
--- a/game/world/triggers/abstracttrigger.cpp
+++ b/game/world/triggers/abstracttrigger.cpp
@@ -32,7 +32,7 @@ AbstractTrigger::AbstractTrigger(Vob* parent, World &world, const zenkit::Virtua
     auto& trigger = reinterpret_cast<const zenkit::VTrigger&>(data);
     fireDelay          = uint64_t(trigger.fire_delay_sec*1000.f);
     retriggerDelay     = uint64_t(trigger.retrigger_delay_sec*1000.f);
-    maxActivationCount = uint32_t(trigger.max_activation_count);
+    maxActivationCount = (data.type==VirtualObjectType::zCMover && trigger.max_activation_count!=0) ? uint32_t(-1) : uint32_t(trigger.max_activation_count);
     target             = trigger.target;
     disabled           = !trigger.start_enabled;
     sendUntrigger      = trigger.send_untrigger;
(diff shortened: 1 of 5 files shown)


game/world/triggers/abstracttrigger.cpp
```diff
@@ -12,8 +12,6 @@ using namespace Tempest;
 
 AbstractTrigger::AbstractTrigger(Vob* parent, World &world, const zenkit::VirtualObject& data, Flags flags)
   : Vob(parent,world,data,flags & (~Flags::Static)), callback(this), vobName(data.vob_name) {
-  if(!hasFlag(StartEnabled))
```
Try, line 15: `StartEnabled` seen to be obsolete (and wrongly added in first place) - can be removed

Try (approved):
Generally LGTM, just asking: are you interested in looking into `sendUntrigger` and `ReactFlg`? 
If no - it's also fine to merge PR as-is

thokkat:
Looked into other attributes. I removed `T_Activate` special handling by defaulting `maxActivationCount` to infinite and now check this property for all  `onTrigger` events.

Try:
LGTM - I need to test(might not have time today/tommorow) it, then will be merged

Try:
Tested today. There is a regressing with pillars in Halls of Idorath:
изображение

thokkat:
Ok problem found. Apparently movers can be triggered indefinitely if `maxActivationCount` is non-zero. While at it I added `autoUntrigger` attribute for cscamera. Otherwise there is a non-cscamera gap between door circle spinning and pillar up movement. I excluded timedemo though for convenience.

Irdorath cscamera overlapping should now be resolved because one of them is always disabled and workaround is no longer needed. Gate up cscamera is triggered a few ticks earlier than spinning gear csamera is untriggered which isn't a problem in this case. But for general case it's worth to keep current logic and end old cscamera if a new one is triggered. Vanilla testing showed same behavior and timedemo became buggy or crashed the game.

Try:
Merged, thanks!

Issue #596, opened 2024-03-20
Installing dependencies (and then make) fails on Linux Mint


Xenoshell:
Installing dependencies:

an error when using sudo apt update:
```
W: GPG error: http://repository.spotify.com stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6224F9941A8AA6D1
E: The repository 'http://repository.spotify.com stable InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
W: https://packages.lunarg.com/vulkan/dists/focal/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```
and when trying to install vulkan-sdk:

```
The following packages have unmet dependencies:
 lunarg-via : Depends: libjsoncpp1 (>= 1.7.4) but it is not installable
 vulkan-profiles : Depends: libjsoncpp1 (>= 1.7.4) but it is not installable
 vulkancapsviewer : Depends: qt5-default but it is not installable
E: Unable to correct problems, you have held broken packages.
```

This is the make error:
```
In file included from /home/../Documents/OpenGothic/lib/Tempest/Engine/gapi/vulkan/vaccelerationstructure.h:5,
                 from /home/../Documents/OpenGothic/lib/Tempest/Engine/gapi/vulkan/vaccelerationstructure.cpp:3:
/home/../Documents/OpenGothic/lib/Tempest/Engine/gapi/vulkan/vulkan_sdk.h:4:10: fatal error: vulkan/vulkan.h: No such file or directory
    4 | #include <vulkan/vulkan.h>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [lib/Tempest/Engine/CMakeFiles/Tempest.dir/build.make:858: lib/Tempest/Engine/CMakeFiles/Tempest.dir/gapi/vulkan/vaccelerationstructure.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/home/../Documents/OpenGothic/build'
make[1]: *** [CMakeFiles/Makefile2:711: lib/Tempest/Engine/CMakeFiles/Tempest.dir/all] Error 2
make[1]: Leaving directory '/home/../Documents/OpenGothic/build'
make: *** [Makefile:136: all] Error 2
[...]

PR #597, opened 2024-03-23
update README for Ubuntu 22.04

Nindaleth:
Fix for #596

In addition to mentioning 22.04:
* LunarG repo also provides packages for Ubuntu 16.04 and 18.04 (but I don't know if it's possible to compile and run OPG there so I didn't mention it to stay safe) 
* in around two months' time the repo will also provide packages for Ubuntu 24.04, but there's some period of time between Ubuntu LTS releasing and LunarG adding support for it in the repo, so again to stay on the safe side I'm not mentioning it explicitly yet.

I've manually verified the /etc/os-release and its VERSION_CODENAME to be available on all of the LTS versions mentioned above (16.04-22.04).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,12 @@ There are ongoing efforts to support parts of it to make at least some popular m
 ## Build Instructions
 ### Linux
 Install dependencies:
-* Ubuntu 20.04
+* Ubuntu 20.04/22.04 and their derived distros
 ```bash
+source <(cat /etc/os-release | grep UBUNTU_CODENAME)
 # latest Vulkan SDK provided externally as Ubuntu packages are usually older
 wget -qO - http://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
-sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-focal.list http://packages.lunarg.com/vulkan/lunarg-vulkan-focal.list
+sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-${UBUNTU_CODENAME}.list http://packages.lunarg.com/vulkan/lunarg-vulkan-${UBUNTU_CODENAME}.list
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,11 +89,13 @@ There are ongoing efforts to support parts of it to make at least some popular m
 ## Build Instructions
 ### Linux
 Install dependencies:
-* Ubuntu 20.04
+* Ubuntu 20.04/22.04
 ```bash
+# obtain the VERSION_CODENAME
+source /etc/os-release
```
Nindaleth, line 95: I could avoid the comment by making the code self-explanatory while also avoiding sourcing of the unnecessary variables, if you'd prefer this approach:
`source <(cat /etc/os-release | grep VERSION_CODENAME)`

Try: Hi, @Nindaleth and thanks for looking into this!

Yes, ideally build instruction should be copy-paste friendly. Sad that it getting more and more complex with `VERSION_CODENAME`

Nindaleth: OK, change is done.

With yum/dnf you can create a single repo using special variables such as `baseurl=http://linux.dropbox.com/fedora/$releasever/` that will always pick the appropate packages, not sure if that's possible with apt but LunarG would make this easier if they used it. But that's not a thing we can change from here :-)

Nindaleth:
Actually, I'm not sure what Linux Mint sets there. I'll need to check this, please don't merge just yet.

Nindaleth:
Turns out that's what the UBUNTU_CODENAME variable is for - each Ubuntu derivative carries its own individual VERSION_CODENAME, but the (base) UBUNTU_CODENAME stays the same. I've verified that the code will also work on Linux Mint and Elementary OS which are some of the notable Ubuntu derivatives.

Try:
Thanks for checking it! Merged

PR #599, opened 2024-04-02
Fix monster damage

thokkat:
Fix a regression introduced in https://github.com/Try/OpenGothic/commit/b1f1b83b158d681771bfcb2bb1d8882bb4ebc33b. Change from `<` to `<=` in `if(src.hitchance[tal]<=critChance)` accidently made monster damage non critical. 

I took this opportunity to check monsters with a weapon and it seems they only do critical hits as well. If `Npc_SetToFightMode` is used in instance script like for goblins weapon damage is not considered. But skeletons have `EquipItem` instead which has makes damage include weapon damage.


Try:
Merged, thanks!

Issue #601, opened 2024-04-03
The teleport stones doesn't works


RaHimsky:
https://github.com/Try/OpenGothic/assets/165494433/6e1aa667-f6cb-47a2-a188-b454d100e0ac

It happens after build 2636.

thokkat:
Thanks for finding, https://github.com/Try/OpenGothic/pull/602 should fix it.

PR #602, opened 2024-04-03
Fix touch triggers

thokkat:
Don't check triggers activated by collision or intersection for `reactToOnTrigger`.

Fixes https://github.com/Try/OpenGothic/issues/601


Try:
FYI: small after-refactoring: b9e7e0c0

PR #606, opened 2024-04-17
Add missing failureTarget trigger for codemaster

thokkat:
If `ordered=true` and `firstFalseIsFailure=false` then `failureTarget` is triggered if number of codemaster ontrigger events matches the number of slaves and either not all slaves have sent trigger or order is wrong. 

Testcases are both upper rooms at switch riddle location. In the right room e.g. any wrong sequence of three activated switches causes a new spawn of shadow skeleton warriors.

### DIFF
--- a/game/world/triggers/codemaster.cpp
+++ b/game/world/triggers/codemaster.cpp
@@ -18,26 +18,25 @@ CodeMaster::CodeMaster(Vob* parent, World &world, const zenkit::VCodeMaster& cm,
   }
 
 void CodeMaster::onTrigger(const TriggerEvent &evt) {
-  for(size_t i=0;i<keys.size();++i) {
-    if(slaves[i]==evt.emitter) {
-      if(!ordered || count==i)
-        keys[i] = true;
-      else if(firstFalseIsFailure) {
-        onFailure();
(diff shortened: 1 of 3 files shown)


game/world/triggers/codemaster.cpp
```diff
@@ -18,53 +18,58 @@ CodeMaster::CodeMaster(Vob* parent, World &world, const zenkit::VCodeMaster& cm,
...
-  zeroState();
-  TriggerEvent e(target,vobName,TriggerEvent::T_Trigger);
-  world.triggerEvent(e);
+  if(count<keys.size())
+    return;
+  if(std::find(keys.begin(),keys.end(),false)!=keys.end()) {
+    if(ordered)
+      onFailure();
+    } else {
+      onSuccess();
```
Try, line 39: code style: extra tab

also would be nice to have less nested 'if''s (and possibly more early-outs instead), for readability

Try (approved):

thokkat:
> also would be nice to have less nested 'if''s (and possibly more early-outs instead), for readability

Done.

Try:
Merged, thanks!

PR #615, opened 2024-05-05
Fix npcs not standing up on player interaction

thokkat:
Praying/Sitting npcs who can't see the player when interacted with are made stand up by their script state end function. In `Npc::startState` aiQueue is cleared after calling the end function so stand up never happens. This is resolved by swapping `clearAiQueue` and `clearState`.


Try:
Merged, thanks @thokkat !

PR #617, opened 2024-05-06
Slow save games - Solution proposal

MadShadow-:
I found a few possibilities to speed up the save game process significantly.

In the serialize class, I changed the compression level from `Best Compression` to `Best Speed`. Files sizes are roughly the same, but without the attempts to perfect compression, the saving is a lot faster.

I also reduced the number of files created by collecting all npc and mob data in a respective single file. See npc and vob cpp.

### !! This breaks load/save game compatibility to previous versions of course !!

To me my changes look innocent, but I might have introduced bugs. 🐞

### Further thoughts
- interactive objects still end up as single files.. but they are way less, so don't have much impact in this world. In theory they could need the same improvement.
- The preview image is still a 4MB full screen image, which is unnecessary. (I've seen new commits regarding save game image size?)

### DIFF
--- a/game/game/serialize.cpp
+++ b/game/game/serialize.cpp
@@ -96,7 +96,7 @@ void Serialize::closeEntry() {
   if(entryBuf.empty())
     return;
 
-  mz_uint level  = entryBuf.size()>256 ? MZ_BEST_COMPRESSION : MZ_NO_COMPRESSION;
+  mz_uint level  = entryBuf.size()>256 ? MZ_BEST_SPEED : MZ_NO_COMPRESSION;
   mz_bool status = mz_zip_writer_add_mem(&impl, entryName.c_str(), entryBuf.data(), entryBuf.size(), level);
   entryBuf .clear();
   entryName.clear();
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -179,7 +179,6 @@ Npc::~Npc(){
   }
 
 void Npc::save(Serialize &fout, size_t id) {
-  fout.setEntry("worlds/",fout.worldName(),"/npc/",id,"/data");
   fout.write(*hnpc);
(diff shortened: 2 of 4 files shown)


Try:
Hi, @MadShadow- !

Unfortunately, I have to reject this PR, most of it at least. We can't drop save-game compatibility support - this is strong most reason. While not really specified, but at least 1year old save should be compatible.

> I changed the compression level from Best Compression to Best Speed

Is it really faster, or it's side effect of larger files? Can you provide numbers, to illustrate the performance gain?

MadShadow-:
> Is it really faster, or it's side effect of larger files? Can you provide numbers, to illustrate the performance gain?

image

I saved 3 games on each run. Left is `MZ_BEST_COMPRESSION` and right is `MZ_BEST_SPEED`. 3 seconds faster on my machine.

_gamesession.cpp_
```c++
void GameSession::save(Serialize &fout, std::string_view name, const Pixmap& screen) {
  SaveGameHeader hdr;
  hdr.version   = Serialize::Version::Current;
  hdr.name      = name;
  hdr.world     = wrld->name();

  using std::chrono::high_resolution_clock;
  using std::chrono::duration_cast;
  using std::chrono::duration;
  using std::chrono::milliseconds;

  auto t1 = high_resolution_clock::now();


  {
  time_t now = std::time(nullptr);
  tm*    tp  = std::localtime(&now);
  hdr.pcTime = *tp;
  }
  hdr.wrldTime  = wrldTime;
  hdr.playTime  = ticks;
  hdr.isGothic2 = Gothic::inst().version().game;

  // [savegame code]

  auto t2 = high_resolution_clock::now();

  /* Getting number of milliseconds as an integer. */
  auto ms_int = duration_cast<milliseconds>(t2 - t1);

  /* Getting number of milliseconds as a double. */
  duration<double, std::milli> ms_double = t2 - t1;

  std::cout << ms_int.count() << "ms\n";
  std::cout << ms_double.count() << "ms\n";
  }
```

Issue #604, opened 2024-04-15
Perceptions framework refinements


Try:
Followup to discussing here: https://github.com/Try/OpenGothic/pull/589

Relevant videos:
https://github.com/Try/OpenGothic/assets/100468436/2d1c819b-b7b0-4162-b3ff-2080be5a8623

This behavior observed by @thokkat, with `.senses = 0`, overridden in Perception.d. 

> In marvin g there is P.detect output. From testing it looks like this shows yes if a perc is triggered with player as source. Walking only one step at a time seems to trigger AssessQuietSound only every second step which would explain the above.

In OpenGothic, entering building already emits `PERC_ASSESSENTERROOM` and `PERC_ASSESSQUIETSOUND` is wired to ground-sounds, and then feed into senses-framework at `WorldObjects::tick`. Naturally those perc's are emitted only by player.

Try:
Current idea is to:
* bypass senses framework, for `PERC_ASSESSENTERROOM`/`PERC_ASSESSQUIETSOUND` and instead loop overs npc's in nearby.
* remove `SENSE_HEAR` process from `canSense`

Try:
In G1 npc also react inaccurate (or rather do not react) on same `PERC_*`

Try:
Both cases should be fixed now, by [42770ae](https://github.com/Try/OpenGothic/commit/42770ae343203c2d90e3c1ca13dcd57500830098)


One other case I'm not sure about is G1 script:
```
func void B_AssessQuietSound ()
{
	if (Npc_GetDistToItem(self, item) > PERC_DIST_DIALOG)   // what item?
	{
		return;
	};
	....
```

If player takes item, `item` wont exist anymore. Maybe bug, maybe this is case of game-logic bleed into UI

thokkat:
> bypass senses framework, for PERC_ASSESSENTERROOM/PERC_ASSESSQUIETSOUND

With senses=0 a lot of passive percs if not all could be triggered. In `Percption.d` there is a comment

`Passive perceptions are sense-independent (they always go to all NPCs, regardless of whether they are seen/heard or not)`

I know comments are sometimes wrong/outdated but here it would support the testing results.

> in opengothic npc react only if aiQueue is empty and there is no pending waits (aiWait)

[...]

PR #618, opened 2024-05-06
Improve guild attitudes script functions

thokkat:
In `wld_getguildattitude` the order of guilds is wrong. Attitude from second guild to first is returned instead of first to second.

In `wld_setguildattitude` two attitudes are set but in vanilla it's only the first.

I fixed the above, expanded bounds checking and unified names. In `wld_exchangeguildattitude` every attitude is now only set once instead twice.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1570,32 +1570,25 @@ int GameScript::wld_getformerplayerportalguild() {
   }
 
 void GameScript::wld_setguildattitude(int gil1, int att, int gil2) {
-  if(gilCount==0 || gil1>=int(gilCount) || gil2>=int(gilCount))
+  if(gil1<0 || gil2<0 || gil1>=int(gilCount) || gil2>=int(gilCount))
     return;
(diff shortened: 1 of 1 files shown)


Try:

Can you provide a test case or maybe a *.sav file?

thokkat:
When entering house behind Zuris  while sneaking in vanilla the militia guard is alerted. If AssessEnterRoom was disabled guard wouldn't react. In OG guard was not reacting even if perc was sent. 
Turned out check  `Wld_GetGuildAttitude(slf.guild, portalguild) == ATT_FRIENDLY)`  in `C_NpcIsBotheredByPlayerRoomGuild` fails. Instead of MIL to VLK (friendly), VLK  to MIL (neutral) is checked.

Was testing percs and looked for a case where not sending AssessEnterRoom while sneaking makes a difference, see https://github.com/Try/OpenGothic/issues/604#issuecomment-2096599054.

For `Wld_SetGuildAttitude` I just compared output of `Wld_GetGuildAttitude` before and after assigning new attitude by inserting print commands to script.

Try:
Merged, thanks!

Issue #628, opened 2024-05-14
[Gothic 1] Old camps gates do not close after chapter 4 started


Abendlied:
**Game:**
Gothic I

**Short Description/Observation:**
After Chapter 4 started, both old camp's gates remained open. Diego's and Milten's dialogs work normal.

**Expectation: **
Northern and southern gate should close when chapter 4 starts.

**Steps to reproduce:**
After giving 5 foki to Saturas, chapter 4 starts- Then go to old camp.

**Additional Info:**
-

Issue #585, opened 2024-02-24
NPC walk-teleport mislignment


Try:
In vanilla npc's do have 2 type of execution:
* close to player
* outside of player vision

When outside of player vision, `npc` doesn't walk thru way-path, but teleports from point-to-point instead.

At this point OpenGothic doesn't do 'outside' part, evaluation whole world as 'closeup' case. Evaluation whole world as 'closeup' has a correctness downside, if game-script relies(often times non-intentionally) on teleport mechanics.

Need to: design solution on how OpenGothic can incorporate same behavior as in vanilla, either equivalent one.
One issue here is unlimited draw-range - teleport might be observable to player
Visibility culling is done in CS, and engine has no information on what is visible on CPU-side.

hwnde:
My 2 cents: implement a logic closer to how vanilla works: do a teleport if the npc is outside a certain radius of the player, regardless of direct visibility. The source or destination theoretically might be observable by the player but I'm not sure it will be actually visible in practice. Would need a playtest to confirm. 

Bonus: do a teleport animation. In case of mages it would even make sense :)

hellozyemlya:
Played gothic long time ago, and what to ask what is the results of this behavior? Quests non-completable?

Try:
@hellozyemlya 
> Quests non-completable?

As mention in description: at least 2 quests are not same. In ring of water meeting, the ring members should appear in tavern shortly after quest starts. In OG they wont appear fast, instead they will literally walk to place.

CReimer:
Couldn't an extremely simple solution be used here? Is there anything against a cone shape in the viewing direction + FOV? Everything inside this cone shape is “inside” everything outside is “outside”. If both the source and the target of the NPC are “outside”, teleportation takes place. Otherwise the NPC goes from waypoint to waypoint. You could then repeat the check at each waypoint or perhaps even in between.

It's certainly more complicated to implement than I'm imagining, but sometimes a simple look at things helps to find a good solution.

Translated with DeepL.com (free version)

Abendlied:
An active cone/radius/waypoint implementation is different to the implemented base principle, full map life simulation. I think in the long run we have to think about it as there are multiple things that rely on teleporting.

What I noticed is that sleeping allows teleportation. In the Gothic 1 Chapter 4 scenario, Diego, Milten and Skorpio are at their destination points after sleeping. Dirty dirty fix could be an automated _sleep until next day_ after certain chapter changes, e.g. G1ch4. 
[...]

PR #629, opened 2024-05-16
Fix non-closing gates in Gothic 1

thokkat:
At start of chapter 4 gate guards change daily routine to `TA_GuardWheelClosed`. Then in function `ZS_GuardWheelClosed()` the state of the next mob with matching scheme name is checked.

```
	if (Wld_GetMobState	(self, 	"VWHEEL") == 0)						// Tor offen?
	{
		PrintDebugNpc 	(PD_TA_CHECK,	"...Tor offen!");	
		AI_UseMob		(self, 	"VWHEEL", 1);						// ...dann wieder zumachen!
		AI_UseMob		(self, 	"VWHEEL", -1);                      //und vom Mobsi abmelden
		AI_AlignToWP	(self);
	};
```
`Wld_GetMobState` internally checks only for mobs the npc has collided with while moving which makes the check fail. This is replaced by `World::findInteractive`, a search for all mobsis in close range. The default range has been increased slightly, otherwise only one gate could be opened.

Script comment for  `Wld_GetMobState` says it's looking for the next mob that matches the scheme name. This would be in line with the change.

Script also has a command to close the inner gate but this fails in vanilla. In OG with this PR it fails because the mob can't be found. A Problem could be that the responsible npc `GRD_280_Gardist` has wrong `id=230` but `GRD_230_Gardist` is the npc closing the main gate.

For testing you can use `insert sh` and use the first chapter 4 option.

Fixes https://github.com/Try/OpenGothic/issues/628

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1652,8 +1652,8 @@ int GameScript::wld_getmobstate(std::shared_ptr<zenkit::INpc> npcRef, std::strin
     return -1;
     }
 
-  auto mob = npc->detectedMob();
-  if(mob==nullptr || mob->schemeName()!=scheme) {
+  auto mob = world().availableMob(*npc,scheme);
+  if(mob==nullptr) {
     return -1;
     }
 


Try:
> Wld_GetMobState internally checks only for mobs the npc has collided with while moving which makes the check fail.

There is a reason to implement it like that, for G2:
```c++
// PERC_MOVEMOB handler
func void B_MoveMob()
{
	var string door;
	door = Npc_GetDetectedMob(self);
	if(Hlp_StrCmp(door,"DOOR"))
	{
		if(Wld_GetMobState(self,door) == 0)
		{
			Npc_ClearAIQueue(self);
			AI_UseMob(self,door,1);
			AI_UseMob(self,door,-1);
		};
	}
	else
	{
		return;
	};
	AI_ContinueRoutine(self);
};
``` 
This script has only one use-case, that I'm aware of: when Gorn leaves the jail he can open the door by himself

thokkat:
Ah didn't see that script but Gorn opening the gate works. I changed the search method to be like `availableMob` to return only a mob with correct scheme name.

For G1 gate guards this search is needed but I could a add a previous check for `detectedMob` if you think that makes sense.

Closing gates now don't allow npcs like Milten to leave old camp which is another problem caused by https://github.com/Try/OpenGothic/issues/585

game/game/gamescript.cpp
```diff
@@ -1652,8 +1652,8 @@ int GameScript::wld_getmobstate(std::shared_ptr<zenkit::INpc> npcRef, std::strin
     return -1;
     }
 
-  auto mob = npc->detectedMob();
-  if(mob==nullptr || mob->schemeName()!=scheme) {
+  auto mob = world().findMob(*npc,scheme);
```
Try, line 1655: Should this line be a part of  `detectedMob`?

Idea behind `detectedMob` is to return:
* actively used modsi
* mobsi that cause collision event 
* [new] closemost with valid scheme

game/world/worldobjects.cpp
```diff
@@ -791,18 +791,18 @@ void WorldObjects::marchInteractives(DbgPainter &p) const {
...
     if(i->checkMobName(dest))
       return i;
     }
 
-  float curDist=dist*dist;
+  Interactive* ret     = nullptr;
+  float        curDist = dist*dist;
   interactiveObj.find(pl.position(),dist,[&](Interactive& i){
-    if(i.isAvailable() && i.checkMobName(dest)) {
+    if(!filter(i))
```
Try, line 803: Would be nice to call `filter` last, as caller side might assume that argument to filter is already matching scheme/distance

thokkat:
Ok `Wld_GetMobState` already checks availability so this extra handling is not needed.

> Should this line be a part of detectedMob?

Hmm question is what happens if scheme name doesn't match collision mob. If the npc is interacting with a mob the mob's state is returned even if scheme is different but this seems illogical. Gorn opening the door isn't working in vanilla so couldn't test what a scheme mismatch would do there.

`Wld_GetMobState` is only used for G1 gates and for Gorn moving out of jail. In latter case the collision guarantees  that a search for closest door would find the collision object. 

Script comments indicate `Npc_GetDetectedMob` is only allowed in conjunction with `PERC_MOVEMOB` while `Wld_GetMobState` returns just the state of the nearest found mob. I'd like to keep it as is until there's some possible non-working case.

Try:
> Script comments indicate Npc_GetDetectedMob is only allowed in conjunction with PERC_MOVEMOB while Wld_GetMobState returns just the state of the nearest found mob. I'd like to keep it as is until there's some possible non-working case.

Agreed, lets keep as-is, and revisit if something breaks in future. Merged, thanks!

Issue #557, opened 2023-12-30
Gothic 1 blocking does not work


rokuroo32:
Hello,

I'm currently playing Gothic1 with "./Gothic2Notr.sh -g1 -rt 1 -g ~/Documents//Gothic". There is one bug in the fighting system: I can't block. When I played the steam version blocking worked.
There is no blocking animation when Im in fight mode and press the back key and action key which works fine in normal Gothic.

PR #630, opened 2024-05-16
Fix Gothic 1 melee blocking

thokkat:
Problem was a character confusion. For G2 the number 0 is used in `T_1HPARADE_0` but in G1 the letter O is used in `T_1HPARADE_O`.

What I'm unsure about is the ` FIXME: seems like name check is not needed` part. Testing shows this is necessary because the `DefWindow` can be reached during attack animations. But `DefWindow` refers to a combo window and `DefParWindow` should be the correct one to choose. But then again this one is always empty.

For now I checked against `DefParWindow` and consider empty window as blocking always enabled during animation. 

Fixes https://github.com/Try/OpenGothic/issues/557

### DIFF
--- a/game/graphics/mdlvisual.cpp
+++ b/game/graphics/mdlvisual.cpp
@@ -646,7 +646,7 @@ const Animation::Sequence* MdlVisual::startAnimAndGet(Npc& npc, AnimationSolver:
         bs = BS_RUN;
       break;
     case AnimationSolver::Anim::MoveBack:
-      if(st!=WeaponState::NoWeapon)
+      if(st==WeaponState::Fist || st==WeaponState::W1H || st==WeaponState::W2H)
         bs = BS_PARADE; else
         bs = BS_RUN;
     case AnimationSolver::Anim::RotL:
--- a/game/graphics/mesh/animationsolver.cpp
+++ b/game/graphics/mesh/animationsolver.cpp
@@ -4,6 +4,7 @@
 #include "world/world.h"
 #include "game/serialize.h"
 #include "utils/fileext.h"
+#include "gothic.h"
 #include "skeleton.h"
 #include "pose.h"
 #include "resources.h"
@@ -117,8 +118,10 @@ const Animation::Sequence* AnimationSolver::implSolveAnim(AnimationSolver::Anim 
         return solveFrm("T_FISTATTACKMOVE");
       return solveFrm("S_FISTATTACK");
       }
-    if(a==Anim::AttackBlock)
-      return solveFrm("T_FISTPARADE_0");
+    if(a==Anim::AttackBlock) {
+      bool g2 = Gothic::inst().version().game==2;
+      return g2 ? solveFrm("T_FISTPARADE_0") : solveFrm("T_FISTPARADE_O");
+      }
     }
   else if(st==WeaponState::W1H || st==WeaponState::W2H) {
     if(a==Anim::Attack && pose.hasState(BS_RUN))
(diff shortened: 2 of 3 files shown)


game/graphics/mesh/animation.cpp
```diff
@@ -263,7 +263,7 @@ bool Animation::Sequence::isInComboWindow(uint64_t t, uint16_t comboLen) const {
 
 bool Animation::Sequence::isDefParWindow(uint64_t t) const {
   if(data->defParFrame.size()!=2)
-    return false;
+    return true;
```
Try, line 266: `Pose::isDefence` uses `isDefWindow`, not `isDefParWindow`.

Also, after checking the source, I've find out that `isDefParWindow` is not in use. Was this change made by mistake or this is some kind of leftover?

thokkat: `isDefWindow` is also used for attack anis which made no sense for me in parade if there's `DEF_PAR_FRAME` extra for it. Though this is never used for humans so my plan was to just check the animation and leave def window/parade out completely. Decided to check against `isDefParWindow` later in case this is actually used somewhere. At least form testing it felt parade was always blocking.

Did some search now for `DEF_WINDOW` and modding wiki gives:
> Set a “window” in the animation - an interval of frames during which you need to press the “up” (G1) or the left mouse button (G2) to continue the combo strike.

https://gothic-modding-community.github.io/gmc/zengin/anims/events/#def_window

Try:
After testing this PR:

It looks to me that in opengothic block is handled conceptually wrong.
* In G2 we missing `T_2HPARADE_0_A3`, maybe more 
* `T_2HPARADE_0_A3` show that there is no clear naming pattern
* check for `BS_PARADE` is way more simple

My suggesting is to:
1. `BS_PARADE` test instead of name test + `isDefParWindow (old implementation)` for normal block
2. `BS_PARADE` + lack of `defParFrame` for jump-back block

case 2 will help with Gothic1 and also makes sense in G2. Not sure about animals in G1

thokkat:
>   1.  BS_PARADE test instead of name test + isDefParWindow (old implementation) for normal block
>   2.  BS_PARADE + lack of defParFrame for jump-back block

Do you mean `DefWindow`?  Asking because `defParFrame` is always empty and thus `Animation::Sequence::isDefParWindow` returns always false.

Vanilla is using `t_1hParadeJumpB` instead of `t_1hJumpB` for jump block. Animation is the same but `t_1hParadeJumpB` comes with a DEF_WINDOW  so suggested check for jump wouldn't work then which means vanilla must have some different jump detection. Maybe go back to name testing for jump only and check BS_PARADE for `isDefence`?

Try:
> Do you mean DefWindow?

Yes def-window, sorry. Maybe it's easier to post the code:
```c++
bool Pose::isDefence(uint64_t tickCount) const {
  for(auto& i:lay) {
    if(i.bs==BS_PARADE && i.seq->isDefWindow(tickCount-i.sAnim))
      return true;
  return false;
  }
```
> Asking because defParFrame is always empty

This will fall-thru to jump-back case. Here My assumption that vanilla script is slightly buggy, and engine should interpret such cases as jump-back (aka unconditional block)
```c++
bool Pose::isJumpBack() const {
  for(auto& i:lay) {
    if(i.bs==BS_PARADE && i.seq->data->defParFrame.empty())
      return true;
    }
  return false;
  }
```

Reason, why I'm thinking that this can be correct way: no name checks + simplicity of code. Only issue here, is propagating BS in ` Pose::processLayers`: "T_2HJUMPB" has stand animation as next, resulting into infinite stand-block. Probably it's a bug with BS propagation here.

thokkat:
`BS_PARADE` is now set if npc is in jump back animation and in melee fight mode. Technically this is wrong because in G1 for humans jump back is always BS_RUN and non blocking. For G1 it looks like only monsters block in jump back but haven't tested much more. I'd leave that for later.

I let animation names as is because it already works and avoids the bug you mentioned. One more reason is waran in G1 uses both `JUMPB` and `PARADEJUMPB` which makes it unclear when to pick which.

Try:
I've added 
```
        if(lay[i].bs==BS_PARADE)
          lay[i].bs = BS_RUN;
```
to animation solver, to avoid issues with `T_2HJUMPB`, rest looks good

Try:
Merged, thanks!

PR #631, opened 2024-05-16
`changeAttribute`: don't reduce hitpoints of immortal npcs

thokkat:
The orc in the new mine who makes the Ulu-Mulu would have his health reduced instead of increased when taking his medicine. This is due to a script bug giving him the wrong guild and making this check fail.
```
	if (C_NpcIsOrc(self))
	{
		self.attribute[ATR_HITPOINTS] = self.attribute[ATR_HITPOINTS_MAX];
	}
	else
	{
		Npc_ChangeAttribute	(self,ATR_HITPOINTS,-self.attribute[ATR_HITPOINTS_MAX]);
	};
```
 In vanilla death is prevented by immortality. OG checks immortality only when damage is dealt. This check is now moved to `Npc::changeAttribute` like it was done for godmode.


Try:
LGTM, thanks @thokkat !

PR #633, opened 2024-05-22
Fixed CI/CD bug by mirroring configuration of shadowDownsample

ist182069:
Hey @Try,

Was trying to fix the most recent bug where the pipelines were failing.
So I decided on mirroring the configuration of `shadowDownsample` and do the same for the initialized variable `shadowRq`.

I introduced the same on the `Shader` class where the `shadowRq` function call will invoke `Tempest::ComputePipeline`.

I do not know which issue you are going for right now, could you point me to it if this fix is not enough?

Cheers,

Jose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,17 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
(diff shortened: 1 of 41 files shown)


Try:
Hi, @ist182069 !

Sorry, for breaking CI this week - I accidentally pushed a part of code, that belongs to rectified-shadowmap. No need to do PR - just ping me on discord next time :)

PR #634, opened 2024-05-24
Add `goto vob` and `goto camera` marvin commands

thokkat:
`goto vob` teleports the player to a vob that matches the given name.  A index number can be added to teleport to the n_th vob in case there exist multiple, often the case for items. Only npcs and items are tested for now but in vanilla it works for all kind of vobs.

`goto camera` teleports the player to the camera position. I found this came in handy when I wanted to observe a npc without percs being triggered but then move quickly to the scene to look what would happen.

### DIFF
--- a/game/world/world.h
+++ b/game/world/world.h
@@ -104,8 +104,8 @@ class World final {
 
     auto                 takeHero() -> std::unique_ptr<Npc>;
     Npc*                 player() const { return npcPlayer; }
-    Npc*                 findNpcByInstance(size_t instance, uint32_t n = 1);
-    Item*                findItemByInstance(size_t instance, uint32_t n = 1);
+    Npc*                 findNpcByInstance(size_t instance, size_t n = 0);
+    Item*                findItemByInstance(size_t instance, size_t n = 0);
     std::string_view     roomAt(const Tempest::Vec3& arr);
 
     void                 scaleTime(uint64_t& dt);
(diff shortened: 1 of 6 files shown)


Try (approved):

thokkat:
> will merge after finishing my work with sky
Sure no hurry. Just some pr/issue spam incoming from my side. Want to wrap up stuff lying around :)

game/world/worldobjects.cpp
```diff
@@ -452,10 +452,29 @@ Npc *WorldObjects::findHero() {
   return nullptr;
   }
 
-Npc *WorldObjects::findNpcByInstance(size_t instance) {
-  for(auto& i:npcArr)
-    if(i->handle().symbol_index()==instance)
-      return i.get();
+Npc *WorldObjects::findNpcByInstance(size_t instance, uint32_t n) {
+  if(n==0)
```
Try, line 456: Normally in C++ indexing is normally zero-based. 
I'm suggesting to have this function as zero-based, and in that case `-1` passed from caller will be casted to `UINT_MAX` and result into null result.

game/world/world.h
```diff
@@ -104,7 +104,8 @@ class World final {
 
     auto                 takeHero() -> std::unique_ptr<Npc>;
     Npc*                 player() const { return npcPlayer; }
-    Npc*                 findNpcByInstance(size_t instance);
+    Npc*                 findNpcByInstance(size_t instance, uint32_t n = 1);
+    Item*                findItemByInstance(size_t instance, uint32_t n = 1);
```
Try, line 108: In project indices are nominated in `size_t`, not uint

Try (changes requested):

Try:
Merged, thanks!

PR #635, opened 2024-05-24
Remove obsolete code

thokkat:
When first entering Valley of Mines npc like militita Den are still alive and have 1hp max health. They are killed by script when the dragon attack video sequence is triggered. The npc code is probably some remains before trigger were added to make sure these npc are in fact dead.

Trigger code is only used for adding triggers to `triggersZn` but this vector is never used later again. Looks like a leftover when trigger collision was moved to collision zone object `collisionZn`.


Try:
Merged, thanks!

PR #636, opened 2024-05-24
Trigger improvements

thokkat:
This aims at fixing some more trigger quirks.

Apparently movers ignore `fireDelay` and `sendUntrigger`. In https://github.com/Try/OpenGothic/discussions/622 issues13.1 the stone gate opens immediately instead of respecting delay and the metal gate doesn't have `sendUntrigger` set to true, blocking it from opening. I added `retriggerDelay` here  as well without testing.

Same goes for collision. Part of code in https://github.com/Try/OpenGothic/pull/635 suggests movers can't have it and I confirmed it by testing. CsCamera behave the same way. In sleeper temple are some that could be wrongly triggered by normal walking.
I changed the test for number of npcs in a trigger zone to be any positive number instead of exactly one. There was a case where a number of skeletons following the player would prevent the trigger.

In another case a trigger activated by intersection had a `fireDelay`. Because it was not called by `WorldObjects::execTriggerEvent` it did not work. This made it necessary to move the call for population of `triggersDef` object inside `AbstractTrigger` class.

### DIFF
--- a/game/world/triggers/abstracttrigger.cpp
+++ b/game/world/triggers/abstracttrigger.cpp
@@ -82,7 +82,7 @@ void AbstractTrigger::processEvent(const TriggerEvent& evt) {
   if(fireDelay>0) {
     TriggerEvent ex(evt.target, evt.emitter, world.tickCount() + fireDelay, evt.type);
     delayedEvent = std::move(ex);
-    world.addDefTrigger(*this);
+    world.enableDefTrigger(*this);
     return;
     }
   implProcessEvent(evt);
@@ -173,8 +173,11 @@ void AbstractTrigger::load(Serialize& fin) {
   fin.read(emitCount,disabled);
   fin.read(emitTimeLast);
 
-  if(fin.version()>=47)
+  if(fin.version()>=47) {
     delayedEvent.load(fin);
+    if(hasDelayedEvents())
+      world.enableDefTrigger(*this);
+    }
   if(fin.version()>=48) {
     fin.read(ticksEnabled);
     if(ticksEnabled)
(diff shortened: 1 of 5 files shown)


game/world/triggers/abstracttrigger.cpp
```diff
@@ -78,6 +82,7 @@ void AbstractTrigger::processEvent(const TriggerEvent& evt) {
   if(fireDelay>0) {
     TriggerEvent ex(evt.target, evt.emitter, world.tickCount() + fireDelay, evt.type);
     delayedEvent = std::move(ex);
+    world.addDefTrigger(*this);
```
Try, line 85: why this code with `WorldObjects::addDefTrigger` exists?
tracking if deferred triggers is already implemented outsize of trigger class, that should be enough.

thokkat: Problem is `AbstractTrigger::processEvent` is called directly from `onIntersect` and `onCollide` functions for triggers activated by touch/collision which prevents them being added to `defTriggers`.  Idea was to move adding directly after fireDelay check.

game/world/worldobjects.cpp
```diff
@@ -490,6 +486,10 @@ void WorldObjects::addTrigger(AbstractTrigger* tg) {
   triggers.emplace_back(tg);
   }
 
+void WorldObjects::addDefTrigger(AbstractTrigger& t) {
+  triggersDef.push_back(&t);
```
Try, line 490: It's better to check here for `triggersDef` not containing `&t` already

game/world/worldobjects.cpp
```diff
@@ -400,11 +400,7 @@ bool WorldObjects::execTriggerEvent(const TriggerEvent& e) {
     auto& t = *i;
     if(t.name()!=e.target)
       continue; // NOTE: trigger name is not unique - more then one trigger can be activated
-
-    const bool hadDelayedEvt = t.hasDelayedEvents();
     t.processEvent(e);
-    if(!hadDelayedEvt && t.hasDelayedEvents())
```
Try, line 406: Same code exists in `WorldObjects::load`. Since now it's Trigger responsibility to track deferred stuff, should be moved to `AbstractTrigger::load`

Try:
Merged, thanks!

Issue #640, opened 2024-05-24
[Gothic 1] Npcs in `ZS_Teaching` don't align to their waypoint


thokkat:
Example npc is Baal Cadar, looking in the wrong direction. This is caused by a `AI_AlingnToWp` call missing in the script state enter function.

In vanilla this can be observed after talking to him. There's no repositioning. But after a walk to his waypoint positioning would work which suggests npcs are aligned automatically by the game once they reach their destination waypoint.

PR #644, opened 2024-05-27
Align npcs to their freepoint once it's reached

thokkat:
Basically the code is copied from `AI_AlignToWp`. I left out the non-zero check for readability because it's checked later again down the function chain.
Test npc: `goto vob GUR_1208_BaalCadar`.

Removed empty line accidentally made from the gotoVob PR. 

Fixes https://github.com/Try/OpenGothic/issues/640


Try:
Merged, thanks!

PR #648, opened 2024-06-27
Don't use a time limit for non-daily routine states

thokkat:
If a new state is started the end time is not set to infinite but the current daily routine end time is kept instead. So if a npc is fighting but a new daily routine begins the npc would quit the fight and go home. There's already code that excludes `zs_talk` from early ending.

This has been changed in https://github.com/Try/OpenGothic/commit/7af7f0ce28f53532d5fab29e0524e73f8556bedc to avoid daily routines breaking. In my testing everything worked correctly with the pr change, likely code change over time made daily routines  more robust.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -2245,7 +2245,7 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
         }
       break;
     case AI_StartState:
-      if(startState(act.func,act.s0,aiState.eTime,act.i0==0)) {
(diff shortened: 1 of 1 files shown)


Try:
Thanks for PR, @thokkat !

> If a new state is started the end time is not set to infinite but the current daily routine end time is kept instead.

Can you please add this as a comment to code? In case if we would run into another state-tracking issue, we better keep knowledge of why `gtime::endOfTime()` is important.

Try (approved):

thokkat:
Added comment.

Try:
Merged, thanks!

PR #650, opened 2024-07-02
Allow exit from load menu after trying to load empty save

Pinemarten:
It's not possible to exit the load menu if you try to load an empty save game slot (stumbled upon this accidentally). 

This PR fixes it. In case loading save failed, the `closeFlag` is not set to true.
Also minor refactor to use `string_frm` instead of `std::snprintf`.

### DIFF
--- a/game/ui/gamemenu.cpp
+++ b/game/ui/gamemenu.cpp
@@ -5,8 +5,6 @@
 #include <Tempest/TextCodec>
 #include <Tempest/Dialog>
 
-#include <format>
-
 #include "utils/string_frm.h"
 #include "world/objects/npc.h"
 #include "world/world.h"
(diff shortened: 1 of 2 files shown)


Try:
Hi, @Pinemarten and thanks for the fix!

> Also minor refactor to use std::format (available in C++20) instead of std::snprintf (I guess some left-over?).

Not all C++ compilers de facto support `std::format`, unfortunately:
```
[ 94%] Building CXX object CMakeFiles/Gothic2Notr.dir/game/ui/gamemenu.cpp.o
/Users/appveyor/projects/opengothic/game/ui/gamemenu.cpp:922:37: error: no member named 'format' in namespace 'std'
  const std::string fileName = std::format("save_slot_{}.sav",int(id));
                               ~~~~~^
1 error generated.
make[3]: *** [CMakeFiles/Gothic2Notr.dir/game/ui/gamemenu.cpp.o] Error 
```

Note, that opengothic has it's own square wheel for string formation: `string_frm` .

Try:
Merged, thanks!

PR #651, opened 2024-07-09
Use key down and support key repeat in game menu

Pinemarten:
This PR includes two changes:

- Use `keyDown` instead of `keyUp` in game menu. In vanilla Gothic 2 NOTR it's actually key down that triggers the menu action (e.g., going through menu items with up/down arrows), so this change makes OpenGothic in line with vanilla.
- Add support for key repeat. In vanilla, it's possible to hold your key to decrease/increase value on the slider in the menu. For example, it's possible to hold left arrow key to decrease the music volume. It was previously not possible in OpenGothic.

### DIFF
--- a/game/ui/menuroot.cpp
+++ b/game/ui/menuroot.cpp
@@ -161,7 +161,6 @@ void MenuRoot::keyRepeatEvent(Tempest::KeyEvent &e) {
   else if(e.key == Event::K_S || e.key == Event::K_Down)
     current->onKeyboard(KeyCodec::Back);
   }
-  
 
 void MenuRoot::keyDownEvent(KeyEvent &e) {
   size_t sz = std::extent_v<decltype(cheatCode)>;
@@ -203,4 +202,3 @@ void MenuRoot::keyDownEvent(KeyEvent &e) {
       popMenu();
     }
   }
-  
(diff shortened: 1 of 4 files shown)


Try:
Hi, @Pinemarten !

Thanks for PR, looks good - will test once I'm back at home, roughtly by the end of this week

Try:
Tested PR today:

* In the main-menu vanilla has keyRepeat for UP/DOWN keys, but it's not achieved by this path\
* For no/yes options LEFT/RIGHT keys do not have loop behavior in vanilla, but still can toggle no/yes only by holding only left key
* For % options case is similar. For example view-range: in vanilla you can hold left and go as low as 20% and then it stops, in this path it loops to 300%
* For sliders - works well

game/ui/menuroot.cpp
```diff
@@ -151,6 +151,13 @@ void MenuRoot::mouseWheelEvent(MouseEvent &event) {
     }
   }
 
+void MenuRoot::keyRepeatEvent(Tempest::KeyEvent& e) {
+    if(e.key==Event::K_A || e.key==Event::K_Left)
```
Try, line 155: code-style: tabulation in project is 2 spaces, not 4

Try:
Merged, thanks!

Try:
Just to note: `keyRepeatEvent` in Tempest has repeat period same as OS. And usually it adjusted for text typing, not for video-games. So it's significantly faster than in vanilla-game.

PR #652, opened 2024-07-10
Fix weapon undraw sound not being played

Pinemarten:
Seems to be a regression introduced in https://github.com/Try/OpenGothic/commit/488b607ad42856c34ef19a478069f39ed59cd270

The `auto ws = visual.fightMode();` variable was already set to `NoWeapon` after `visual.setFightMode` call, so no undraw sound was played.


Try:
Merged, thanks!

Issue #305, opened 2022-09-13
Earthquake can be heard from large distance


Dehela:
Looks like there is an issue with earthquake, which happens near temple ruins which leads to Jharkendar, that it can be heard from large (or even unlimited) distance. Like, starting the game in Xardas Tower, you can already hear it from time to time.

I suppose it might something to do with full life simulation (like mentioned in FPS ticket) for this version so the event happens despite the distance. It case if it matters, I was testing master branch on Linux.

Try:
> I suppose it might something to do with full life simulation

Almost: actual root cause of earthquakes in game is Saturas. In his state-script he has 5% chance to summon global earth-quake.

Script callstack:
```
"WLD_PLAYEFFECT"
"B_EVENT_PORTAL_EARTHQUAKE"
"ZS_STUDY_WP_LOOP"
--- 
Npc::tickRoutine
```

Dehela:
So, is it actually a normal thing to happen? I haven't play the game for quite a while so I might recall things wrong.

Try:
I'm also not 100% sure how it was in vanilla.

So far my point is:
1. Earthquakes are related to npc-processing
2. It' likely that original game doesn't execute npc-state-scripts, if npc is outside of certain distance 
3. Script is crap to begin with :)

Dehela:
Sounds like a complicated issue to fix unless things are altered. Probably excluding world effect from that script is the least evil, which is unfortunate if you aim to keep game logic intact as it was in original (but might be not too bad if it's not the first time to make a workaround patch).

PolyMeilex:
Random idea, perhaps we could check source of the effect and fade it smoothly based on position, so instead of unloading like gothic did it woul be a lot nicer and slowly decay over distance.

sirsteve1:
> > I suppose it might something to do with full life simulation
> 
> Almost: actual root cause of earthquakes in game is Saturas. In his state-script he has 5% chance to summon global earth-quake.
> 
> Script callstack:
> 
> ```
> "WLD_PLAYEFFECT"
> "B_EVENT_PORTAL_EARTHQUAKE"
> "ZS_STUDY_WP_LOOP"
> --- 
> Npc::tickRoutine
> ```

Dear __Try__, I tried to find the callstack of the earthquake routine but had no chance of finding anything. In which files would I have to lookup? I would like to play around and see if I can find a solution for this. Checked out both the OpenGothic repo as well as the Gothic2 game folder but found nothing...

[...]

PR #653, opened 2024-07-16
Add some vanilla behavior for npcs

thokkat:
If a npc comes into visibility range:

- Best weapons are automatically equipped. For ranged weapons ammo has to be in inventory. From testing turns out this is checked in `equipBestRangedWeapon`. Bartok can be used as test npc.

- Script function `B_RefreshAtInsert` that restores health is called in G2.

Additionally clean up some leftover of 1hp npc handling and setting the damage type to 2 if its default was 0. Needed for the hero in G1. Tested as behaving the same in G2 for some monsters and humans by setting their damage type to 0 in script.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1128,6 +1128,14 @@ void GameScript::invokePickLock(Npc& npc, int bSuccess, int bBrokenOpen) {
   vm.call_function<void>(fn, bSuccess, bBrokenOpen);
   }
 
+void GameScript::invokeRefreshAtInsert(Npc& npc) {
+  auto fn = vm.find_symbol_by_name("B_RefreshAtInsert");
+  if(fn==nullptr)
+    return;
+  ScopeVar self(*vm.global_self(), npc.handlePtr());
+  vm.call_function<void>(fn);
+  }
+
 CollideMask GameScript::canNpcCollideWithSpell(Npc& npc, Npc* shooter, int32_t spellId) {
   auto fn   = vm.find_symbol_by_name("C_CanNpcCollideWithSpell");
   if(fn==nullptr)
--- a/game/game/inventory.cpp
+++ b/game/game/inventory.cpp
@@ -946,6 +946,8 @@ Item* Inventory::bestItem(Npc &owner, ItmFlags f) {
       continue;
     if(!i->checkCond(owner))
       continue;
+    if(itData.munition>0 && findByClass(size_t(itData.munition))==nullptr)
+      continue;
 
     if(itData.value>g){
       ret=i.get();
--- a/game/game/inventory.cpp
+++ b/game/game/inventory.cpp
@@ -905,13 +905,9 @@ void Inventory::invalidateCond(Item *&slot, Npc &owner) {
     }
   }
 
-void Inventory::autoEquip(Npc &owner) {
-  sortItems();
-
-  auto a = bestArmour     (owner);
+void Inventory::autoEquipWeapons(Npc &owner) {
   auto m = bestMeleeWeapon(owner);
   auto r = bestRangeWeapon(owner);
-  setSlot(armour,a,owner,false);
   setSlot(melee ,m,owner,false);
   setSlot(range ,r,owner,false);
   }
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -474,6 +474,10 @@ bool Npc::resetPositionToTA() {
 
   if(!isDead)
     attachToPoint(at);
+
+  if(g2)
+    owner.script().invokeRefreshAtInsert(*this);
+
   return true;
   }
 
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -168,6 +168,10 @@ Npc::Npc(World &owner, size_t instance, std::string_view waypoint)
 
   owner.script().initializeInstanceNpc(hnpc, instance);
   hnpc->wp       = std::string(waypoint);
+
+  // vanilla forces non-zero damage type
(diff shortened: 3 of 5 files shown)


game/world/objects/npc.cpp
```diff
@@ -471,8 +472,15 @@ bool Npc::resetPositionToTA() {
...
 
-  if(!isDead)
+  if(!isDead) {
     attachToPoint(at);
+    if(!isMonster())
+      invent.autoEquipWeapons(*this);
+    }
+
+  if(g2)
+    owner.script().invokeRefreshAtInsert(*this);
```
Try, line 482: do you need to call this for dead npc's as well?

game/game/gamescript.cpp
```diff
@@ -1128,6 +1128,14 @@ void GameScript::invokePickLock(Npc& npc, int bSuccess, int bBrokenOpen) {
   vm.call_function<void>(fn, bSuccess, bBrokenOpen);
   }
 
+void GameScript::invokeRefreshAtInsert(Npc& npc) {
+  auto fn = vm.find_symbol_by_name("B_RefreshAtInsert");
+  if(fn==nullptr)
```
Try, line 1133: it would be good idea to cache function in `GameScript`, similar to `ItKE_lockpick`, as there will be burst of this calls after "sleep".

game/game/inventory.cpp
```diff
@@ -950,6 +946,8 @@ Item* Inventory::bestItem(Npc &owner, ItmFlags f) {
       continue;
     if(!i->checkCond(owner))
       continue;
+    if(itData.munition>0 && findByClass(size_t(itData.munition))==nullptr)
+      continue;
```
Try, line 950: > For ranged weapons ammo has to be in inventory.

How this interact with `B_CreateAmmo` ? Does game requires to have some ammo upfront, but then adds ammo automatically?

Try (commented):

Can you also have a look here: https://github.com/Try/OpenGothic/issues/305 (last few comments), as this seem to be related?

game/world/objects/npc.cpp
thokkat: Good Point. Testing shows script is called for dead npcs. But first line directly returns in that case.
```C
func void B_RefreshAtInsert()
{
	// EXIT IF...
	
	// ------ Tote = Exit ------
	if (self.attribute[ATR_HITPOINTS] <= 0)
	{
		return;
	};
```

game/game/inventory.cpp
thokkat: This is only a check in `AI_EquipBestRangedWeapon`. If npc has a strong bow equipped without ammo and weak crossbow with bolts in inventory, `AI_EquipBestRangedWeapon` would equip the crossbow. 

`EquipItem` doesn't care for ammo.

Try:
I've tested path today: it seems we missing a case, when `autoEquipWeapons` should be invoked. In G2 `insert PC_ROCKEFELLER` inserts an paladin with berserkers-axe and dragon-crossbow. In opengothic - default paladin sword. Would it be correct to also call `autoEquipWeapons` in Npc constructor? Not sure about different cases, such as summons.

game/world/objects/npc.cpp
```diff
@@ -169,7 +169,9 @@ Npc::Npc(World &owner, size_t instance, std::string_view waypoint)
   owner.script().initializeInstanceNpc(hnpc, instance);
   hnpc->wp       = std::string(waypoint);
 
-  // vanilla forces non-zero damage type
+  // vanilla behavior: equip best weapon and set non-zero damage type
+  if(!isMonster() && !isPlayer())
```
Try, line 173: One more question: `isMonster` is truly relevant? For wolfs/snappers should not mater; but maybe relevant for skeletons

thokkat: Yes best weapon autoequip works for skeletons and from what I tested it seems to only work for skeletons. Other monsters just have the weapon in their inventory.

For like a wolf weapon inserted with `CreateInvItem` is equipped and visible in OpenGothic, that's why I excluded monsters in general.

thokkat: After some thinking it's maybe best to prevent weapon equipping for monsters somewhere in the equip function directly in case someone adds a `EquipItem` call in script. 

Skeleton is detected based on guild in vanilla. Changing the guild of a golem to skeleton resulted in a draw/remove weapon stall.

Try: Since it not directly related to `isMonster` would be nice to at least understand how it intended to work.
I've done some testing with marvin - guild by itself seem to be irrelevant. Maybe it's related to `npc` having some builtin bones, such as `ZS_SWORD`/`ZS_RIGHTHAND` ?

Can you test what will happen if you assign `GIL_SHEEP` to skeleton or "Ske_Body" to sheep, while having weapon in inventory?

Try: UPD:
Tested by editing Nikolas guild: as `GIL_SKELETON` or `GIL_SHEEP` he still equips best weapon.

thokkat: > GIL_SHEEP to skeleton

Nothing.

> "Ske_Body" to sheep

Nothing but if `Mdl_SetVisual (self,	"HumanS.mds")`  with Sheep_Body sword is visible. 

For default sheep both `Npc_HasEquippedWeapon` and `Npc_HasEquippedMeleeWeapon` return true.

Try: > GIL_SHEEP to skeleton
Nothing.

You mean nothing changed, or skeleton does not equip weapons any more?)

thokkat: Nothing changed.

PR #654, opened 2024-07-17
Drop items from right hand instead of left hand

Pinemarten:
If you take a look at the drop animation, the item should be actually coming out of the right hand instead of the left hand.


Try:
Merged, thanks @Pinemarten !

PR #655, opened 2024-07-17
CMAA2 integration

KirillAlekseeenko:
-Integration implemented with different quality modes (higher mode means lower edge detection threshold)
-Tonemaping switches to compute when CMAA2 is on in order to avoid additional copy from RT resource to UAV.
-Created additional HDR luminance resource, but currently it's not used, since there is a problem, that in some locations HDR range is so small, that almost no edges are detected even on max quality preset.


game/graphics/renderer.cpp
```diff
@@ -596,8 +671,23 @@ void Renderer::drawTonemapping(Encoder<CommandBuffer>& cmd) {
...
+
+    tonemapping.uboTone.set(2, cmaa2.sceneTonemappedUav);
+    tonemapping.uboTone.set(3, cmaa2.sceneHdrLumaUav);
+    cmd.setUniforms(*tonemapping.computePso, tonemapping.uboTone, &p, sizeof(p));
+
+    const uint32_t threadGroupSize = 8;
+    const uint32_t groupsCountX = (cmaa2.sceneTonemappedUav.w() + threadGroupSize - 1) / threadGroupSize;
+    const uint32_t groupsCountY = (cmaa2.sceneTonemappedUav.h() + threadGroupSize - 1) / threadGroupSize;
+
+    cmd.dispatch(groupsCountX, groupsCountY, 1);
```
Try, line 685: You don't need to calculate workgroup count manually, there is: `cmd.dispathThreads`.

Also `Y` and `Z` axis are `1` by default, so no need to specify `Z` in 2 dimensional workload.

> const uint32_t threadGroupSize = 8;

Also, if really needed there is `ComputePipeline::workGroupSize`

game/graphics/renderer.cpp
```diff
@@ -626,6 +716,49 @@ void Renderer::drawFxaa(Encoder<CommandBuffer>& cmd) {
   cmd.draw(Resources::fsqVbo());
   }
 
+void Renderer::applyCmaa2(Tempest::Encoder<Tempest::CommandBuffer>& cmd)
+{
+  // TODO: initialization that is needed only on the first run. Make it run only in the first run
```
Try, line 721: TODO

shader/lighting/tonemapping.frag
```diff
@@ -10,6 +10,32 @@
 
 #include "upscale/lanczos.glsl"
 
+#if defined(COMPUTE)
+const uint THREADGROUP_SIZE = 8;
```
Try, line 14: Any reason to have a constant for `THREADGROUP_SIZE`? `gl_WorkGroupSize` should be just fine.

shader/antialiasing/cmaa2/compute_dispatch_args.comp
```diff
@@ -0,0 +1,50 @@
+#version 460
+
+#extension GL_ARB_separate_shader_objects       : enable
+#extension GL_GOOGLE_include_directive          : enable
+#extension GL_EXT_control_flow_attributes       : enable
+
+#include "cmaa2_common.glsl"
+
+layout(binding = 8) buffer UboWorkingExecuteIndirectBuffer {
+  uint g_workingExecuteIndirectBuffer[];
```
Try, line 10: Need to add explicit memory layout: `std430`.
Also why an array? Would it be nicer to have:
```
uvec3 shapeCandidateIndirectArg;
```
?

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,196 @@
...
+#define CMAA_MSAA_SAMPLE_COUNT                      1       // for now force mode without MSAA
+#define CMAA_MSAA_USE_COMPLEXITY_MASK               1
+#define CMAA2_EDGE_UNORM                            0
+#define CMAA2_EXTRA_SHARPNESS                       0
+
+const float c_symmetryCorrectionOffset = 0.22;
+
+#if CMAA2_EXTRA_SHARPNESS
+  const float c_dampeningEffect = 0.11;
+  #define g_CMAA2_LocalContrastAdaptationAmount       0.15f
```
Try, line 18: why sometimes use `const` and sometimes `#define` ? 

Note: generally `const` is preferred in opengothic, when possible.

shader/antialiasing/cmaa2/compute_dispatch_args.comp
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+
+void main() {
+  uvec3 groupID = gl_WorkGroupID;
+
+  if(groupID.x == 1) {
```
Try, line 18: Here `gl_WorkGroupID` is uses as essentially a flag. Better to just use push-constant.

shader/CMakeLists.txt
```diff
@@ -196,6 +196,9 @@ add_shader(tonemapping.vert          copy.vert -DHAS_UV)
 add_shader(tonemapping.frag          lighting/tonemapping.frag)
 add_shader(tonemapping_up.frag       lighting/tonemapping.frag -DUPSCALE)
 
+add_shader(tonemapping.comp          lighting/tonemapping.frag -S comp -DCOMPUTE)
+add_shader(tonemapping_up.comp       lighting/tonemapping.frag -S comp -DCOMPUTE -DUPSCALE)
```
Try, line 200: Does it make sense to have need anti-aliasing + upscale together?

Try (commented):
Have had only quick look yet. Some overall notes:
* Do we really need so many quality presets? Can we simplify code a bit by selection only one or two that do fit well?
* Definitely need to remove MSAA cases, MSAA is anti-aliasing by itself and also not supported in engine.
* Need to update submodules..

Also good finding in `shader/copy_img.comp`!

shader/CMakeLists.txt
KirillAlekseeenko: In FSR1 documentation there's a note that it's better to use antialiased input to FSR1 and we have the same Lanczos filter for upscaling. But since any MLAA algorithm doesn't provide additional samples unlike TAA or MSAA, it seems that edge reconstruction is just a reverse process.

KirillAlekseeenko:
1) 4 presets is too much, I just repeated Intel's demo here; MIDDLE and ULTRA should be enough.
2) I thought about potential integration with MSAA in a future where MSAA is for geometry aliasing and CMAA2 is for anything else.

game/graphics/renderer.cpp
KirillAlekseeenko: Is there something like a current frame number to determine if the current frame is the first one?

game/graphics/renderer.cpp
```diff
@@ -626,6 +711,56 @@ void Renderer::drawFxaa(Encoder<CommandBuffer>& cmd) {
...
+
+    uint32_t groupCountX = static_cast<uint32_t>((cmaa2.sceneTonemappedUav.size().w + outputGroupSize.x * 2 - 1) / (outputGroupSize.x * 2));
+    uint32_t groupCountY = static_cast<uint32_t>((cmaa2.sceneTonemappedUav.size().h + outputGroupSize.y * 2 - 1) / (outputGroupSize.y * 2));
+
+    cmd.setUniforms(*cmaa2.detectEdges2x2, cmaa2.detectEdges2x2Ubo);
+    cmd.dispatch(groupCountX, groupCountY, 1);
+    }
+
+  // set indirect for processCandidates pass
+  {
```
Try, line 739: No need to wrap `cmd.` dispatches into brackets

game/graphics/renderer.cpp
```diff
@@ -626,6 +711,56 @@ void Renderer::drawFxaa(Encoder<CommandBuffer>& cmd) {
...
   }
 
+void Renderer::applyCmaa2(Tempest::Encoder<Tempest::CommandBuffer>& cmd)
+{
+  uint32_t processCandidatesSetupFlag = 1;
+
+  static bool isFirstRun = true;
+  // initialization that is needed only on the first run. Make it run only in the first run
+  if (isFirstRun) {
+    cmd.setUniforms(*cmaa2.prepareDispatchIndirectArguments, cmaa2.prepareDispatchIndirectArgumentsUbo, &processCandidatesSetupFlag, sizeof(uint32_t));
```
Try, line 721: line feel a bit wordy, maybe `prepareDispatchIndirectArguments` -> `indirectArgs` ?

game/graphics/renderer.cpp
```diff
@@ -626,6 +711,56 @@ void Renderer::drawFxaa(Encoder<CommandBuffer>& cmd) {
...
+    cmd.dispatch(1);
+    isFirstRun = false;
+    }
+
+  // detect edges
+  {
+    const IVec3 inputGroupSize = cmaa2.detectEdges2x2->workGroupSize();
+    const IVec3 outputGroupSize = inputGroupSize - IVec3(2, 2, 0);
+
+    uint32_t groupCountX = static_cast<uint32_t>((cmaa2.sceneTonemappedUav.size().w + outputGroupSize.x * 2 - 1) / (outputGroupSize.x * 2));
```
Try, line 731: In c++ no need to use `static_cast` - it can be replaced with constructor-cast:
`static_cast<uint32_t>(epr) -> uint32_t(expr)`

game/graphics/renderer.cpp
```diff
@@ -626,6 +711,56 @@ void Renderer::drawFxaa(Encoder<CommandBuffer>& cmd) {
   cmd.draw(Resources::fsqVbo());
   }
 
+void Renderer::applyCmaa2(Tempest::Encoder<Tempest::CommandBuffer>& cmd)
+{
+  uint32_t processCandidatesSetupFlag = 1;
+
+  static bool isFirstRun = true;
```
Try, line 718: See `gi::fisrtFrame`. 
In fact to handle general case you need not the very first frame, but first, after effect been enabled.

game/graphics/renderer.cpp
```diff
@@ -187,7 +214,9 @@ void Renderer::initSettings() {
 
   auto prevVidResIndex = settings.vidResIndex;
   settings.vidResIndex = Gothic::inst().settingsGetF("INTERNAL","vidResIndex");
-  settings.fxaaEnabled = (Gothic::options().fxaaPreset > 0) && (settings.vidResIndex==0);
+  settings.cmaa2Enabled = (Gothic::options().cmaa2Preset>0) && (settings.vidResIndex==0);
+  settings.fxaaEnabled = (Gothic::options().fxaaPreset>0) && (settings.vidResIndex==0) && !settings.cmaa2Enabled;
```
Try, line 218: Is FXAA still relevant? Any case, when use can prefer FXAA to CMAA?

shader/lighting/tonemapping.frag
```diff
@@ -10,6 +10,31 @@
...
+layout(binding = 2) uniform writeonly image2D tonemappedOutput;
+layout(r32f, binding = 3) uniform writeonly image2D hdrLumaOutput;
+
+/* layout:
+0 1 4 5 ... 
+2 3 6 7 ...
+16 17 ...
+18 19 ...
+*/
+uvec2 getRemmappedPixelPos(uvec2 groupID, uint localThreadIndex)
```
Try, line 24: Does this remapping works? So far wasn't able to find any performance gain on my machine (Intel + NVidia)

shader/antialiasing/cmaa2/edge_color2x2.comp
```diff
@@ -0,0 +1,256 @@
+#version 460
```
Try, line 1: Shader code feel like HLSL in many ways...

Is it from Intel's demo, or is it rewrite of original DX code? If rewrite, need to adjust style and language to glsl:
* `g_` opengothic doesn't use it
* function name state with lower-case
* avoid bindings in header file (if unavoidable - put all of them into single header, such as cmaa2_common.glsl)

shader/antialiasing/cmaa2/edge_color2x2.comp
```diff
@@ -0,0 +1,256 @@
...
+#if CMAA2_EDGE_DETECTION_LUMA_PATH == 2
+  layout(binding = 8) uniform sampler2D g_inLumaReadonly;
+#endif
+
+#define CMAA2_CS_INPUT_KERNEL_SIZE_X                16
+#define CMAA2_CS_INPUT_KERNEL_SIZE_Y                16
+#define CMAA2_CS_OUTPUT_KERNEL_SIZE_X               (CMAA2_CS_INPUT_KERNEL_SIZE_X-2)
+#define CMAA2_CS_OUTPUT_KERNEL_SIZE_Y               (CMAA2_CS_INPUT_KERNEL_SIZE_Y-2)
+
+layout(local_size_x = CMAA2_CS_INPUT_KERNEL_SIZE_X, local_size_y = CMAA2_CS_INPUT_KERNEL_SIZE_Y, local_size_z = 1) in;
```
Try, line 18: `CMAA2_CS_INPUT_KERNEL_SIZE_X*CMAA2_CS_INPUT_KERNEL_SIZE_Y = 16*16 = 256`

Vulkan guaranties only 128 compute invocations.

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+  ret.w = float((value & 0x08) != 0);
+  return ret;
+  }
+
+vec3 LoadSourceColor(ivec2 pixelPos, ivec2 offset, int sampleIndex) {
+  vec3 color = texelFetch(g_inoutColorReadonly, pixelPos + offset, 0).xyz;
+  return color;
+  }
+
+uint Pack_R11G11B10_FLOAT(vec3 rgb) {
```
Try, line 83: pack/unpack functions seem to be good candidate for `common.glsl`.

Also naming can be more glsl-like. By glsl-like I mean "8.4. Floating-Point Pack and Unpack Functions", such as `packR11G11B10Float`

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+  if(val < 0.0031308) val *= 12.92;
+  else val = 1.055 * pow(abs(val), 1.0 / 2.4) - 0.055;
+  return val;
+  }
+
+vec3 LINEAR_to_SRGB(vec3 val) {
+  return vec3(LINEAR_to_SRGB(val.x), LINEAR_to_SRGB(val.y), LINEAR_to_SRGB(val.z));
+  }
+
+uint FLOAT4_to_R8G8B8A8_UNORM(vec4 unpackedInput) {
```
Try, line 147: Is it same as `packUnorm4x8` ?

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+  return Pack_R11G11B10_FLOAT(color);
+#else
+  return Pack_R11G11B10_E4_FLOAT(color);
+#endif
+  }
+
+float LINEAR_to_SRGB(float val) {
+  if(val < 0.0031308) val *= 12.92;
+  else val = 1.055 * pow(abs(val), 1.0 / 2.4) - 0.055;
+  return val;
```
Try, line 140: OpenGothic uses more simple curve in `srgbEncode`.
Should we use more advance gamma-correction and tonemapping is orthogonal question, for now let's keep code coherent.

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+
+uint Pack_R11G11B10_E4_FLOAT(vec3 rgb) {
+  rgb = clamp(rgb, 0.0, uintBitsToFloat(0x3FFFFFFF));
+  uint r = ((PackFloat32AsFloat16AndConvertToUint(rgb.r) + 4) >> 3) & 0x000007FF;
+  uint g = ((PackFloat32AsFloat16AndConvertToUint(rgb.g) + 4) << 8) & 0x003FF800;
+  uint b = ((PackFloat32AsFloat16AndConvertToUint(rgb.b) + 8) << 18) & 0xFFC00000;
+  return r | g | b;
+  }
+
+vec3 Unpack_R11G11B10_E4_FLOAT(uint rgb) {
```
Try, line 114: any reason to support `R11G11B10_E4` in shader? Engine doesn't support it, and there is not plans to implements shared exponent formats.

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+
+vec3 Unpack_R11G11B10_FLOAT(uint rgb) {
+  float r = uintBitsToFloat((rgb << 4) & 0x7FF0);
+  float g = uintBitsToFloat((rgb >> 7) & 0x7FF0);
+  float b = uintBitsToFloat((rgb >> 17) & 0x7FE0);
+  return vec3(r, g, b);
+  }
+
+uint PackFloat32AsFloat16AndConvertToUint(float arg) {
+  return packHalf2x16(vec2(arg, 0.));
```
Try, line 99: why introduce proxy function, instead of calling `packHalf2x16` directly?

shader/antialiasing/cmaa2/process_candidates.comp
```diff
@@ -0,0 +1,349 @@
...
+void StoreColorSample(uvec2 pixelPos, vec3 color, bool isComplexShape, uint msaaSampleIndex) {
+  uint counterIndex = atomicAdd(g_workingControlBuffer[12], 1);
+
+  uvec2 quadPos = pixelPos / uvec2(2, 2);
+  uint offsetXY = (pixelPos.y % 2) * 2 + (pixelPos.x % 2);
+  uint header = (offsetXY << 30) | (msaaSampleIndex << 27) | (uint(isComplexShape) << 26);
+
+  uint counterIndexWithHeader = counterIndex | header;
+
+  uint originalIndex = imageAtomicExchange(g_workingDeferredBlendItemListHeads, ivec2(quadPos), counterIndexWithHeader);
```
Try, line 29: NOTE: Engine emulates `imageAtomic*` on Metal 2.x and only for `r32ui`.
Should not be an issue in this code, just a note.

Try (commented):
About compy_img.comp: would you mind if  I'll commit fix for `local_size_y` separately?
Asking as review of PR might take a bit of time.

shader/lighting/tonemapping.frag
```diff
@@ -10,6 +10,31 @@
 
 #include "upscale/lanczos.glsl"
 
+#if defined(COMPUTE)
+layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;
+layout(binding = 2) uniform writeonly image2D tonemappedOutput;
+layout(r32f, binding = 3) uniform writeonly image2D hdrLumaOutput;
```
Try, line 16: `writeonly` doesn't require explicit pixel-format specifier

game/graphics/renderer.cpp
```diff
@@ -84,8 +84,22 @@ void Renderer::resetSwapchain() {
 
   if(settings.fxaaEnabled) {
     fxaa.sceneTonemapped = device.attachment(TextureFormat::RGBA8, w, h);
+    } else if(settings.cmaa2Enabled) {
+    cmaa2.sceneTonemappedUav = device.image2d(TextureFormat::RGBA8, w, h);
```
Try, line 88: Better not to use `UAV` sufix here, as it is DX12 terminology and engine largely follows Vulkan terms

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,176 @@
...
+
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// VARIOUS QUALITY SETTINGS
+//
+// Longest line search distance; must be even number; for high perf low quality start from ~32 - the bigger the number, 
+// the nicer the gradients but more costly. Max supported is 128!
+const uint c_maxLineLength = 86;
+// 
+
+layout(binding = 0) uniform sampler2D g_inoutColorReadonly;
```
Try, line 32: `g_inoutColorReadonly` -> `sceneTonemapped` ?

shader/lighting/tonemapping.frag
Try: Also compute variant can be removed, if you segregate `sceneTonemapped` (pre CMAA) and anti-aliased `sceneTonemapped`, that is written by `deferred_color_apply_2x2.comp`

shader/lighting/tonemapping.frag
KirillAlekseeenko: https://developer.nvidia.com/blog/optimizing-compute-shaders-for-l2-locality-using-thread-group-id-swizzling/
It should increase perfornamce on gpus that have vram throughput as a bottleneck

game/graphics/renderer.cpp
KirillAlekseeenko: Just in case someone prefer more blurry image. I doubt there many people with such preferences though.

shader/lighting/tonemapping.frag
Try: I'm aware of that article. Problem, that so far wasn't able to confirm it: even with locked GPU-frequency performance change in memory-bound algorithm (fog occlusion) is in a noise range.

And in general doesn't "feel" right, to reason about cache-locality of non-opaque data with unknown(to application) memory layout.

shader/antialiasing/cmaa2/cmaa2_common.glsl
KirillAlekseeenko: It's taken from https://github.com/Microsoft/DirectX-Graphics-Samples/blob/master/MiniEngine/Core/Shaders/PixelPacking_R11G11B10.hlsli

This format is used for packing LDR range more efficiently than standard R11G11B10 by stripping the exponent.

shader/antialiasing/cmaa2/cmaa2_common.glsl
KirillAlekseeenko: To not repeat packHalf2x16(vec2(arg, 0.)) for each channel and make it like f32tof16 in HLSL.

shader/antialiasing/cmaa2/cmaa2_common.glsl
KirillAlekseeenko: Yes

KirillAlekseeenko:
About compy_img.comp, yeah, it's better to commit separately

shader/lighting/tonemapping.frag
KirillAlekseeenko: No, it's the same `sceneTonemapped ` as used in cmaa2

shader/antialiasing/cmaa2/cmaa2_common.glsl
KirillAlekseeenko: I've renamed to more general `inputColorReadonly`, since input could be in HDR potentially

shader/antialiasing/cmaa2/edge_color2x2.comp
KirillAlekseeenko: I put all common (usages in more than one pass) bindings in `cmaa2_common.glsl`

shader/antialiasing/cmaa2/cmaa2_common.glsl
Try: Probably name should be something ala `packHalf1x16`, to be consistent with existing GLSL functions?

shader/lighting/tonemapping.frag
Try: In cmaa2_common.glsl;
`layout(binding = 0) uniform sampler2D inputColorReadonly;` - not a storage-image, but a simple texture. Also good to rename it to `sceneTonemapped`, in order to math with c++ side.

In deferred_color_apply_2x2.comp `outputColor` is indeed a writable image, but you dont have to write into a same one.
Maybe deferred_color_apply_2x2 can later be converted into fragment-shader? It's not heavy one compute features, other then a unordered-write.

shader/antialiasing/cmaa2/cmaa2_common.glsl
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+layout(binding = 5) buffer UboWorkingDeferredBlendItemList {
+  uvec2 workingDeferredBlendItemList[];
+  };
+
+layout(r32ui, binding = 6) uniform uimage2D workingDeferredBlendItemListHeads;
+
+layout(binding = 7) buffer UboWorkingControlBuffer {
+  uint workingControlBuffer[];
```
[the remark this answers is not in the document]
KirillAlekseeenko: No, extra variants didnt require additional counters. I suppose counters are aligned for faster access. 3 and 4 are close probably because they are used together in the indirect setup pass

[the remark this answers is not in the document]
Try: There is a publication about atomics performance profile: https://www.youtube.com/watch?v=VaE_uKPfjv0
I don't remember details now, yet my takeaway was that only single variable case is fast-path. Anything else - unpredictable across multiple gpu vendors.

In any case no point of using array, if you really need padding - use struct:
```c++
buffer {
  uint numCandidates;
  uint shapeCandidateCount;
  uint padding0[3];
  uint blendLocationCount;
  ...
  }
```

Try:
@KirillAlekseeenko Pr generally looks good. Last few changes can do by myself.

Let me know, when you happy with current state of PR, so I can start testing/finalizing

KirillAlekseeenko:
Seems that it's ready for integration

shader/antialiasing/cmaa2/compute_dispatch_args.comp
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+layout(push_constant, std140) uniform PushConstant {
+  uint setupProcessCandidatesPassFlag;
+  } push;
+
+layout(local_size_x = 1) in;
+
+void main() {
+  if(push.setupProcessCandidatesPassFlag != 0) {
```
Try, line 20: In C++ side:
```c++
  uint32_t processCandidatesSetupFlag = 1;

  // initialization that is needed only on the first run. Make it run only in the first run
  if(cmaa2.isFirstFrame) {
    cmd.setUniforms(*cmaa2.indirectArgsSetup, cmaa2.indirectArgsSetupUbo, &processCandidatesSetupFlag, sizeof(uint32_t));
    cmd.dispatch(1);
    cmaa2.isFirstFrame = false;
    }
```
You have extra run of this shader for first frame, writing only `g_workingExecuteIndirectBuffer`. Did you mean `processCandidatesSetupFlag = 0` ?

KirillAlekseeenko: Yes, the 'else' path should be run here, the point is just clearing the buffer, but it also could be done explicitly on initialization

Try:
Experimenting with edge detection in HDR space:
| Final image(HDR)  |  Footprint   | 
|--------------|------------|
| изображение| изображение |
| Final image(LDR)  |  Footprint   | 
| изображение | изображение |

In LDR path, I've changed encoding from `packR11G11B10E4F` to `packUnorm4x8`, as image is already tonemapped and in gamma-space.
In HDR path I'm using `packR11G11B10F`, to math frame buffer format.

> HDR luminance resource, but currently it's not used, since there is a problem

Computing luminance in flight via `CMAA2_SUPPORT_HDR_COLOR_RANGE==1` seem to work.

At this point I'm thing to make HDR as default and only path, since it fits better into existing pipeline (0.06ms performance saved on RTX, compare to LDR!) and some VRAM bandwidth savings.

@KirillAlekseeenko let me know what you think about it.

KirillAlekseeenko:
I wanted to use HDR luminance but encountered cases where HDR is much darker than LDR like on the shots below. So finally decided to stay with LDR. The game was run without cmd args except for -aa.
HDR:
HDR
LDR: 
LDR

Try:
> I wanted to use HDR luminance but encountered cases where HDR is much darker

It seems that tonemapping been applied twice or not applied at all on you image.

Here same spot on my working branch (9:00am):
изображение

Try:
> I didnt look deeply into this but for some reason after sky and fog passes color values are around 0.1.

Here how `sceneLinear` (pre tonemapping) look on my end:
изображение

if we divide it by exposure value in a frame = `0.0000088533`, we will get sky brightness = `18 072`lum, what is fine for day-time; clear sky.

Can you check if it's still an issues with my changes?

KirillAlekseeenko:
It works fine but there's two-three times less candidate pixels than with LDR pass, maybe it's worth trying adaptive edge detection threshold based on exposure value

Try:
> maybe it's worth trying adaptive edge detection threshold based on exposure value

Probably no, at least I don't see how it can be better. Right now input is pre-exposed, close to [0..1] range. Maybe raise `cmaa2EdgeThreshold`; but I'm not sure how to reason about quality then - as it's gonna be subjective.

KirillAlekseeenko:
Seems that there's no need to raise threshold, I just took a look at nsight trace and it's even faster than fxaa

Try:
Merged, thanks!

PR #659, opened 2024-07-26
Fix spell charging animation

thokkat:
Fixes a c&p leftover where parameter `run` was checked twice in `AnimationSolver::solveAnim`. Introduced here https://github.com/Try/OpenGothic/commit/825bf12b32e000c28dca4585d9b545b57ea19a96.

### DIFF
--- a/game/graphics/mesh/animationsolver.cpp
+++ b/game/graphics/mesh/animationsolver.cpp
@@ -393,7 +393,7 @@ const Animation::Sequence* AnimationSolver::solveAnim(std::string_view scheme, b
     name = string_frm("T_MAGMOVE_2_",scheme,"CAST");
   else if(run)
     name = string_frm("T_MAGMOVE_2_",scheme,"SHOOT");
-  else if(run)
+  else if(invest)
     name = string_frm("T_MAGRUN_2_",scheme,"CAST");
   else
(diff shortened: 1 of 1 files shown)


Try:

Can you please, provide a test case, for me to do the testing?

Try (approved):

thokkat:
Charging Spells:
```
ItSc_Pyrokinesis
ItSc_ChargeFireball
ItSc_Windfist
ItSc_Thunderball
```
After change all are fixed.

thokkat:
In Gothic 1 shoot animation is used as invest animation for spells like teleport. Stumbled upon pr issue while testing  G1 spells so wanna look into the new issue first.

thokkat:
Ok in G1s  non-charging spells like firestorm are called with `invest=true` which gives the CAST instead of the SHOOT animation. Fallback S_SHOOT kicks in an gives the right animation. Should this be changed to find the transition animation or is state animation sufficient?

Try:
> Should this be changed to find the transition animation or is state animation sufficient?

OpenGothic animation system usually prefer state-animations (maybe not best design). PR looks good, testing is fine, are you OK with merging it?

Try:
Merged, thanks!

PR #660, opened 2024-07-27
Don't consider npc type for attitudes

thokkat:
Special handling for npc type friend in `GameScript::personAttitude`  is preventing a guru from having the default dialog lines. The "don't talk to me" Gurus have this dialog check:
```C++
FUNC INT DIA_BaalTyon_NoTalk_Condition()
{
	if ( Npc_IsInState(self,ZS_TALK) && (BaalTyon_Ansprechbar==FALSE) && (Npc_GetPermAttitude(self,other)!=ATT_FRIENDLY) )
	{
		return 1;
	};
};
```
This is a problem for Baal Tyon who has npc type friend and therefore the attitude check returns false. Other gurus are npc type main and dialogs have no issues.

For testing: `goto vob gur_1210_baaltyon`

In `GameScript::isFriendlyFire` there's the check  `if(personAttitude(src, dst)==ATT_FRIENDLY)`. Don't know if this should check for npc type now instead of attitude considering that like for gururs npc type seems a bit arbitrary.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1300,8 +1300,6 @@ Attitude GameScript::personAttitude(const Npc &p0, const Npc &p1) const {
   att = npc.attitude();
   if(att!=ATT_NULL)
     return att;
-  if(npc.isFriend())
-    return ATT_FRIENDLY;
   att = guildAttitude(p0,p1);
   return att;
   }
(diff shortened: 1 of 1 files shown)


Try:
> Don't know if this should check for npc type now instead of attitude considering that like for gururs npc type seems a bit arbitrary.

Removing this check from fight, will causes problems when interacting with Diego in oldworld and maybe with other archers:
```c++
instance PC_ThiefOW(Npc_Default)
{
	name[0] = "Diego";
	guild = GIL_NONE;
	...
	npcType = NPCTYPE_FRIEND;
```
Afaik `GIL_NONE` to `GIL_NONE` has neutral relation, and without `NPCTYPE_FRIEND` Diego might shoot you in a back.
Lets' keep `isFriend` check in `isFriendlyFire`, at least until we found a case when it doesn't work correct.

thokkat:
> Lets' keep isFriend check in isFriendlyFire

Makes sense, done.

Try:
Merged, thanks!

Issue #639, opened 2024-05-24
More perc issues


thokkat:
Have to bring this up again. In G2 old world Mitten is automatically talking to the player at first meeting if in close distance. But this is happening through a wall when Milten is in the right room and player in the left room in old mage building. Can be tested with `-w oldworld.zen` and `goto pos 0 0 0`.

Cause is this line in `Npc::canSenseNpc`
```C++
  if(isNoisy) {
    // no need to be in same room: https://github.com/Try/OpenGothic/issues/420
    ret = ret | SensesBit::SENSE_HEAR;
    }
```
Commenting out works.

---
Two open points from https://github.com/Try/OpenGothic/issues/604
> Npcs wake up if player is picking a lock in sneak mode. 

My suggestion was 
```C++
- if(ev.groundSounds>0 && isPlayer() && (bodyStateMasked()!=BodyState::BS_SNEAK))
+ if(ev.groundSounds>0 && isPlayer() && (wlkMode&WalkBit::WM_Sneak)!=WalkBit::WM_Sneak)
  world().sendPassivePerc(*this,*this,*this,PERC_ASSESSQUIETSOUND);
```
I'm asking because this was part of rejected https://github.com/Try/OpenGothic/pull/589 and hasn't been moved to https://github.com/Try/OpenGothic/commit/fccda5bec4afede9d31d90889eec4c99e80fc2d6.

Secon question  was
> AssessEnterRoom not sent while sneaking is incorrect. Test case can be house behind Zuris. Guard reacts on enter if sneaking.

This is blocked by the sneak check

```C++
void WorldObjects::sendImmediatePerc(Npc& self, Npc& other, Npc& victum, Item* itm, int32_t perc) {
  const auto pl = owner.player();
  if(pl==nullptr || pl->bodyStateMasked()==BS_SNEAK)
    return;
```

Thesneak check could be removed because `ASSESSQUIETSOUND` isn't sent in that case. Is this special treatment for `QUIETSOUND/ENTERROOM` as immediate perc needed at all? Is there a difference in executing immediately or later in the same tick together with other passive percs?

PR #673, opened 2024-08-20
fix. lockpicking reaction.

Domarion:
Hi!
[Npcs wake up if player is picking a lock in sneak mode](https://github.com/Try/OpenGothic/issues/639) because every lock-picking action calls `snd_play3d` which unconditionally sends `PERC_ASSESSQUIETSOUND`. 

In G2NotR scripts `snd_play3d` function is used only for lock-picking status sounds. And script sends `PERC_ASSESSQUIETSOUND` only when lock-pick was broken:

```
func void G_PickLock(var int bSuccess,var int bBrokenOpen)
{
	var int rnd;
	if(bSuccess)
	{
		if(bBrokenOpen)
		{
			Snd_Play3d(self,"PICKLOCK_UNLOCK");
			Print(PRINT_PICKLOCK_UNLOCK);
		}
		else
		{
			Snd_Play3d(self,"PICKLOCK_SUCCESS");
			Print(PRINT_PICKLOCK_SUCCESS);
		};
	}
	else if(bBrokenOpen)
	{
		Snd_Play3d(self,"PICKLOCK_BROKEN");
		Print(PRINT_PICKLOCK_BROKEN);
		rnd = Hlp_Random(100);
		if(rnd <= 25)
		{
			Npc_SendPassivePerc(hero,PERC_ASSESSQUIETSOUND,hero,hero);
		};
	}
	else
	{
		Snd_Play3d(self,"PICKLOCK_FAILURE");
		Print(PRINT_PICKLOCK_FAILURE);
	};
};
```

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -3188,7 +3188,6 @@ void GameScript::snd_play3d(std::shared_ptr<zenkit::INpc> npcRef, std::string_vi
     c = char(std::toupper(c));
   auto sfx = ::Sound(*owner.world(),::Sound::T_3D,file,npc->position(),0.f,false);
   sfx.play();
-  owner.world()->sendImmediatePerc(*npc,*npc,*npc,PERC_ASSESSQUIETSOUND);
   }
 
 void GameScript::exitsession() {


Try:
Make sense. Merged, thanks @Domarion !

PR #676, opened 2024-08-24
Magic improvements

thokkat:
Some refactor/cleanup of the switch statements e.g. fallback for teleport spells is no longer needed after changes to animation start function. 

PR has fixes for these issues:

- `SPL_DONTINVEST` allows the player to hold the charged spell until key is released. `investedMana` is then no longer accumulated.

Gothic 1 exclusive:

- A lot of spells ( i.e teleport, firerain) use the SHOOT animation as cast animation. If `castLevel` is in `CS_Cast` range SHOOT animation would be started a second time after spell is already sent. To prevent this play animation only if CAST animation  exists.
- When a teleport spell is aborted transition to idle animation could be skipped by walking forward. `CS_Finalize` is now handled like `CS_Emit` except spell is not committed.
- For `code=SPL_NEXTLEVEL` change `castLevel` only if `castLvl<15` instead of `castLvl<=15`. Otherwise it was possible to jump from `CS_Invest` to `CS_Cast` which broke spells like summon demon with a long cast phase.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -3497,7 +3497,7 @@ bool Npc::beginCastSpell() {
   if(active==nullptr)
     return false;
 
-  if(owner.version().game==1 && attribute(ATR_MANA)==0) {
+  if(attribute(ATR_MANA)<=0) {
     setAnim(Anim::MagNoMana);
     return false;
     }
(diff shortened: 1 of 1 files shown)


Try:
Look good - merged.
Thanks @thokkat !

Issue #661, opened 2024-08-01
[Gothic 1] Magic Quirks


Abendlied:
- Has not been fixed by PR #659 yet
- is partly linked to discussion #622 but separated in a standalone ticket

**Describe the bug**
Multiple Magic spells not working as intended. 

- [ ] 1. fire rain animation broken
- [ ] 2. Ice wave does no damage, does not freeze enemies
- [ ] 3. Telekinesis cannot focus on items (e.g. sleeper temple statuette) while attack spells can focus on items
- [ ] 4. Pyrokinesis does no damage, has no spell effect, broken enemy animation

**Save-game/Log-file/Screenshots/Video**
1. Fire Rain 
    savegame: [save_slot_14.sav.tar.gz](https://github.com/user-attachments/files/16457732/save_slot_14.sav.tar.gz)
    mp4-video: 7 sec, 6.1 MiB: [fire rain quirky animation](https://github.com/user-attachments/assets/29e7256f-bc88-4b68-bf06-d19dd51bf392)

2. Ice wave
    savegame: [save_slot_14.sav.tar.gz](https://github.com/user-attachments/files/16457732/save_slot_14.sav.tar.gz)
    mp4-video: 6 sec, 6.6 MiB: [ice wave no freezing no dmg](https://github.com/user-attachments/assets/073e1851-948d-4a3f-b3c0-2173be41fe29)

3. Telekinesis
    savegame: [save_slot_14.sav.tar.gz](https://github.com/user-attachments/files/16457732/save_slot_14.sav.tar.gz)
    mp4-video: 10 sec, 8.9 MiB:[telekinesis cannot focus on item](https://github.com/user-attachments/assets/78464ad7-41e1-4b63-8ef2-58d33f87bf5a)

4. Pyrokinesis
    savegame: [save_slot_14.sav.tar.gz](https://github.com/user-attachments/files/16457732/save_slot_14.sav.tar.gz)
    mp4-video: 6 sec, 6.6 MiB: [pyrokinesis no dmg, no](https://github.com/user-attachments/assets/1110b691-8f35-4f71-8f6e-3f5a4965c713)

**Application log**
<details>
<summary>fire rain animation log (no crash, just log)</summary>

> OpenGothic v1.0 dev
> no "SystemPack.ini" file in path - using default settings
> GPU = NVIDIA GeForce GTX 1050
> Depth format = Depth32F Shadow format = Depth16
> [zenkit] Parsing object [MeshAndBsp % 0 0]
[...]

PR #682, opened 2024-09-06
Fix some Gothic 1 magic bugs

thokkat:
This PR fixes issues 1. and 2. from https://github.com/Try/OpenGothic/issues/661 and  18.1 and 18.2 from G1 bug [discussion](https://github.com/Try/OpenGothic/discussions/622).

(1) Broken firerain effect is fixed by adding it's `Target_collect_algo` attribute to `Item::isSpellShoot`. I proposed this change some time ago but withdrew because it would break Stormfist. Found out now that projectile property is hardcoded ([source](https://forum.worldofplayers.de/forum/threads/1460092-Stuck-in-a-charge-spell/page2?p=24786462&viewfull=1#post24786462)) and used that for Stormfist.

(2) Added a missing perc trigger on collision to make icewave freezing work.

(18.1/2.) Spell targets can be restricted based on attribute `Target_collect_type`, meaning orc priest just cannot be targeted. For `TARGET_TYPE_UNDEAD`  I took `C_NpcIsUndead` script function from G2 as reference since it doesn't exist in G1. 
Target type is not always correctly set in vanilla. For example icewave can target items.

While at it also limited focus search for interactives in combat to only bow and crossbow as magic can't target those.

### DIFF
--- a/game/graphics/effect.cpp
+++ b/game/graphics/effect.cpp
@@ -306,6 +306,9 @@ void Effect::onCollide(World& world, const VisualFx* root, const Vec3& pos, Npc*
   if(npc!=nullptr)
     vfx = root->emFXCollDyn;
 
+   // Gothic 1 uses emFXCollDyn->sendAssessMagic as check for PERC_ASSESSMAGIC
+   // Gothic 2 instead introduced new vfx emFXCollDynPerc specific for this purpose
+  bool g2 = world.version().game==2;
   if(vfx!=nullptr) {
     Effect eff(*vfx,world,pos,SpellFxKey::Collide);
     eff.setSpellId(splId,world);
@@ -314,7 +317,7 @@ void Effect::onCollide(World& world, const VisualFx* root, const Vec3& pos, Npc*
 
     if(npc!=nullptr) {
       npc ->runEffect(std::move(eff));
-      if(vfx->sendAssessMagic) {
+      if(!g2 && vfx->sendAssessMagic) {
         auto oth = other==nullptr ? npc : other;
         npc->perceptionProcess(*oth,npc,0,PERC_ASSESSMAGIC);
         }
@@ -323,7 +326,7 @@ void Effect::onCollide(World& world, const VisualFx* root, const Vec3& pos, Npc*
       }
     }
 
-  if(npc!=nullptr && root->emFXCollDynPerc!=nullptr) {
+  if(g2 && npc!=nullptr && root->emFXCollDynPerc!=nullptr) {
     const VisualFx* vfx = root->emFXCollDynPerc;
     Effect eff(*vfx,world,pos,SpellFxKey::Collide);
     eff.setActive(true);
--- a/game/graphics/effect.cpp
+++ b/game/graphics/effect.cpp
@@ -312,9 +312,15 @@ void Effect::onCollide(World& world, const VisualFx* root, const Vec3& pos, Npc*
     eff.setOrigin(other);
     eff.setActive(true);
 
-    if(npc!=nullptr)
-      npc ->runEffect(std::move(eff)); else
+    if(npc!=nullptr) {
+      npc ->runEffect(std::move(eff));
+      if(vfx->sendAssessMagic) {
+        auto oth = other==nullptr ? npc : other;
+        npc->perceptionProcess(*oth,npc,0,PERC_ASSESSMAGIC);
+        }
+      } else {
       world.runEffect(std::move(eff));
+      }
     }
 
   if(npc!=nullptr && root->emFXCollDynPerc!=nullptr) {
--- a/game/world/objects/item.cpp
+++ b/game/world/objects/item.cpp
@@ -210,8 +210,9 @@ bool Item::isSpellShoot() const {
   // Only hardcode Stormfist for now since other spells work with the heuristic based on target_collect_algo
   if(!isSpellOrRune())
     return false;
-  bool g1 = world.version().game==1;
-  if(g1 && spellId()==47)
+  bool          g1            = world.version().game==1;
+  const int32_t SPL_STORMFIST = 47;
+  if(g1 && spellId()==SPL_STORMFIST)
     return true;
   auto& spl = world.script().spellDesc(spellId());
   return spl.target_collect_algo!=TargetCollect::TARGET_COLLECT_NONE &&
--- a/game/world/objects/item.cpp
+++ b/game/world/objects/item.cpp
@@ -4,6 +4,7 @@
 
 #include "game/serialize.h"
 #include "game/gamescript.h"
+#include "utils/versioninfo.h"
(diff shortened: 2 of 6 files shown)


game/world/objects/item.cpp
```diff
@@ -204,12 +205,19 @@ bool Item::isMulti() const {
   }
 
 bool Item::isSpellShoot() const {
+  // Whether a spell is a projectile is hardcoded in vanilla, see
+  // https://forum.worldofplayers.de/forum/threads/1460092-Stuck-in-a-charge-spell/page2?p=24786462&viewfull=1#post24786462
+  // Only hardcode Stormfist for now since other spells work with the heuristic based on target_collect_algo
   if(!isSpellOrRune())
     return false;
+  bool g1 = world.version().game==1;
+  if(g1 && spellId()==47)
```
Try, line 214: nitpick: better to decouple `47` in something like `const int32_t SPL_STORMFIST = 47`

game/world/objects/item.cpp
```diff
@@ -204,12 +205,19 @@ bool Item::isMulti() const {
   }
 
 bool Item::isSpellShoot() const {
+  // Whether a spell is a projectile is hardcoded in vanilla, see
```
Try, line 208: In link you provided:
```c++
  'bIsProjectile' TRUE (if visual effect name doesn't contain "_SPREAD"):
    38, 87
  'bIsProjectile' TRUE (always):
    2, 6, 19, 20, 23, 26, 30, 33, 34, 36, 43, 71, 72, 73, 80, 81, 82, 86
  'bIsProjectile' default is FALSE.
```

Can't find `47` there. Am I missing something?

game/world/worldobjects.cpp
```diff
@@ -995,6 +999,27 @@ static bool checkFlag(Interactive& i,WorldObjects::SearchFlg f){
   return true;
   }
 
+template<class T>
+bool checkTargetType(T&, TargetType) { return true; }
+
+static bool checkTargetType(Npc& n, TargetType t) {
+  if(bool(t&(TARGET_TYPE_ALL|TARGET_TYPE_NPCS)))
```
Try, line 1006: Is this code also G1 only?

game/graphics/effect.cpp
```diff
@@ -312,9 +312,15 @@ void Effect::onCollide(World& world, const VisualFx* root, const Vec3& pos, Npc*
     eff.setOrigin(other);
     eff.setActive(true);
 
-    if(npc!=nullptr)
-      npc ->runEffect(std::move(eff)); else
+    if(npc!=nullptr) {
+      npc ->runEffect(std::move(eff));
+      if(vfx->sendAssessMagic) {
+        auto oth = other==nullptr ? npc : other;
+        npc->perceptionProcess(*oth,npc,0,PERC_ASSESSMAGIC);
```
Try, line 319: There is another call to `PERC_ASSESSMAGIC` in code bellow.
Would it cause double perception, if both `vsm!=nullptr` and `root->emFXCollDynPerc!=nullptr` ?

Try (commented):

> (18.1/2.) Spell targets can be restricted ...

This part seem to be a reference?

Try:
not about red ci: microsoft decided that they want to rename some stuff in dx.
should be fixed by: 1af5396e084b6bf6a4b69f6a71f90aeb790bb21a

game/world/objects/item.cpp
thokkat: These are for G2, G1 has different ids. Except Stormfist `TARGET_COLLECT_ALL_FALLBACK_NONE` is used for icewave, firerain and uriziel rune which are all aoe effects.

Originally I had another source which gave only the hardcoded fact, but couldn't find it anymor. Found one from a dev instead :)

game/world/worldobjects.cpp
thokkat: No can be used for G2. `SPL_WindFist` has `TARGET_TYPE_NPCS`. If changed to `TARGET_TYPE_UNDEAD` novices can't be targeted anymore.

Here it's just a leftover because  non-npcs can't be targeted anyway. Mostly if G2 does something different G1 stuff is still functional but unused,

game/graphics/effect.cpp
thokkat: I went the lazy way and thought the modder is responisble to make sure only one is triggered. But reality is `emFXCollDynPerc` attribute doesn't exist in G1 and is only used in G2.  And in my testing `emFXCollDyn` perc couldn't be triggered in G2, so each Gothic has it's own vfx for percs.

Tested the spells in rockefellers inventory and there wasn't a single spell where `sendAssessMagic` was true for both `emFXCollDyn` and `emFXCollDynPerc`.

thokkat:
> > (18.1/2.) Spell targets can be restricted ...
> 
> This part seem to be a reference?

Meant issues 18.1 and 18.2 in bug discussion

> 18.1 Destroy Undead instantly kills Grash-Varrag-Arushat (5th undead priest), long before player has Uriziel
> savegame: [save_slot_7.sav.tar.gz](https://github.com/user-attachments/files/16228882/save_slot_7.sav.tar.gz)
> no ticket created
> screenshot: n/a

> 18.2 Destroy Undead kills all undead and regular orcs and minecrawler warriors and Demons. Is that base game behaviour?
> savegame: [save_slot_4.sav.tar.gz](https://github.com/user-attachments/files/16228883/save_slot_4.sav.tar.gz)
> no ticket created
> screenshot: n/a

Try:
Merged, thanks!

Abendlied:
Not sure if it is correct to comment after succesful merge.
> This PR fixes issues 1. and 2. from https://github.com/Try/OpenGothic/issues/661 and 18.1 and 18.2 from G1 bug https://github.com/Try/OpenGothic/discussions/622.

>18.1 Destroy Undead instantly kills Grash-Varrag-Arushat (5th undead priest), long before player has Uriziel
    savegame: [save_slot_7.sav.tar.gz](https://github.com/user-attachments/files/16228882/save_slot_7.sav.tar.gz)
    no ticket created
    screenshot: n/a

>    18.2 Destroy Undead kills all undead and regular orcs and minecrawler warriors and Demons. Is that base game behaviour?
    savegame: [save_slot_4.sav.tar.gz](https://github.com/user-attachments/files/16228883/save_slot_4.sav.tar.gz)
    no ticket created
    screenshot: n/a

what I understood you have fixed here is that destroy undead cannot focus on enemies. While it appears to fix the problem, the player can still kill Grash-Varrag-Arushat as well as MC warriors etc. with that spell by aiming free (without focusing). Standing directly infront of G-V-A and charging the spell still kills him even if not focused.

Abendlied:
BIG thank you for this PR @thokkat!
fire rain and ice wave work flawlessly now thanks to your work.

PR #685, opened 2024-09-13
Follow up fix for `TARGET_TYPE`

thokkat:
https://github.com/Try/OpenGothic/pull/682 allowed focus only if npc guild matches the spell's `TARGET_TYPE`. But that spell could still hit and cause damage. 

Tested both G1&G2. Projectile spells hit only if `TARGET_TYPE` allows for, while spells like firerain ignore type and always hit.

I guess in theory there's same behavior for items and interactives but they can't be damaged anyway so ignore them.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1138,6 +1138,13 @@ void GameScript::invokeRefreshAtInsert(Npc& npc) {
   }
 
 CollideMask GameScript::canNpcCollideWithSpell(Npc& npc, Npc* shooter, int32_t spellId) {
+  if(owner.version().game==1) {
+    auto& spl = spellDesc(spellId);
+    if(npc.isTargetableBySpell(TargetType(spl.target_collect_type)))
+      return COLL_DOEVERYTHING; else
+      return COLL_DONOTHING;
+    }
+
   auto fn   = vm.find_symbol_by_name("C_CanNpcCollideWithSpell");
   if(fn==nullptr)
     return COLL_DOEVERYTHING;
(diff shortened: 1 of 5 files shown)


Try:
> https://github.com/Try/OpenGothic/pull/682 allowed focus only if npc guild matches the spell's TARGET_TYPE. But that spell could still hit and cause damage.

Are you sure, that engine also need to check target type on collision?  To me checking only `C_CanNpcCollideWithSpell` makes more sense, and `TARGET_TYPE` is essentially a hint for game-interface.

thokkat:
This is how fireball looks with undead type.

https://github.com/user-attachments/assets/3cea4310-220d-483d-99fc-9a659a4e6b61

In contrast firerain with undead type makes everyone angry

https://github.com/user-attachments/assets/4f0b5fe2-6eb9-44a3-9168-4567d3c4eb74

Other reason I found for preventing collision for projectile spells is that AssessDamage perc is not sent. Undead priest in G1 orc temple triggers a dialog after some hits but those spells with non-matching target type never trigger it.

Try:
> This is how fireball looks with undead type.

Hm, yeah. Still in case of G2 this feel-like leftover from G1, that wasn't intended to be in game:
* in G2-scripts `targetCollectType` is effectively unused
* G2 has `C_CanNpcCollideWithSpell` api to achieve same purpose
While change is vital for G1, for G2 it falls into "bug reproduction" category.

Can we instead make no change to current G2 implementation, and provide a default implementation of `GameScript::canNpcCollideWithSpell` for G1?

thokkat:
> Can we instead make no change to current G2 implementation, and provide a default implementation of GameScript::canNpcCollideWithSpell for G1?

Done. Just wasn't getting what you were up to initially :)

Try:
Merged, thanks!

PR #688, opened 2024-09-28
Update `dmusic`

lmichaelis:
This patch updates [dmusic](https://github.com/GothicKit/dmusic) to fix some music playback issues. In particular, GothicKit/dmusic@7b44736f37cede64b742239d07a25a25d8e3e9f6 fixes an issue which caused drum kits sounds to not be considered during playback, causing the incorrect instruments to be chosen.

Here's an example of how `MO_DayStd` sounds with this update:

<figure>
  
https://github.com/user-attachments/assets/7bcde0aa-e459-4d65-84c9-17e5e40dd8fe

</figure>


Try:
~Looks~ Sounds good to me - merged :)

Thanks!

PR #691, opened 2024-10-18
Added arm64 as supported architecture for edd-dbg

pawcijo:
Fix for macos build, not sure if this is the right place for PR :)

### DIFF
--- a/lib/edd-dbg/src/osx/frames.cpp
+++ b/lib/edd-dbg/src/osx/frames.cpp
@@ -9,7 +9,7 @@
 #include <pthread.h>
 #include <stdint.h>
 
-#if !defined(__i386__) && !defined(__amd64__)
+#if !defined(__i386__) && !defined(__amd64__) && !defined(__arm64__)
 #   error "unsupported architecture :("
 #endif
 


Try:
Hi, @pawcijo !

> added arm64 as supported architecture

Can you clarify this statement? Arm64 for mac been supported for a while, with no build issues.  Or do you mean crash-logger support?

pawcijo:
When i try to build the project on m1 macbook  i get  "unsupported architecture :(", if frames.cpp is from crash-logger than yes your right :)

Try:
How do you build opengothic on mac? `edd` library is enabled only on windows targets

pawcijo:
The same as it goes in Readme :
`cmake -H. -Bbuild -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo
cmake --build ./build --target Gothic2Notr`

Try:
Hm, something is not quite right here. Can you please provide build log, cmake version, and more details overall?

In build script there is:
```
add_subdirectory(lib/edd-dbg)
include_directories(lib/edd-dbg/include)
if(WIN32)
  target_link_libraries(${PROJECT_NAME} edd_dbg)
endif()
```
Maybe `add_subdirectory` behaves different across different versions.

pawcijo:
Cannot reproduce anymore, it might benn some wild cmake error on macos

Issue #287, opened 2022-08-31
Joystick support (for turning)


Nindaleth:
This is a Steam Deck issue (probably a gamepad issue in general) - as there appears to be no joystick support, the top Steam Deck community layout for Gothic II Gold uses a workaround - the physical joystick control emulates a d-pad, i.e. when moving, you either turn left/right 100% or not at all.

I don't really mind that this applies to the forward/backward movement too (at worst I can switch to walk mode), but It's the turning left and right that's really hard to get used to and there's no workaround for that.

Try:
Marked as help-wanted, since there is no Steam-Deck in Norway for me :D

Nindaleth:
I can sell you mine if you'd want one :smile: I just think you don't actually want one, but the offer is serious.

Try:
:D haha 
Thanks, but no - I'll buy it myself, just later. On the other side, some PR would be nice ;)

ezamelczyk:
You don't need a steamdeck to write support for a controller. Steamdeck is just another xinput device just like dualshock or xbox controller. I wanted to write support for this myself but I quickly realized that I'm just a java dev and c++ is too complicated for me.

Try:
I need at least some controller to progress :D 
While at it can you post here, what controller can be good for windows/mac (or controller that you prefer to use)?

ezamelczyk:
I just use my ps5 controller but xbox would be fine as well. The new ones are rather expensive tho so maybe just go for dualshock 4 (the ps4 one).

On the other hand I looked a bit more through the code and I figured that this would have to be implemented into tempest since it handles the event loop so maybe we should move this conversation over there?

I'm bad at c++ but I'm gonna try to implement this. I don't have anything better to do anyway :p

ezamelczyk:
Also it might be easier than I thought since there's already a pretty robust cross-platform library for this https://www.libsdl.org/

I don't know what's your stance on 3rd party libraries tho.  You think this might be a good idea?

Try:
@ezamelczyk 
> I don't have anything better to do anyway :p

Glad to head it :D 

> I just use my ps5 controller but xbox would be fine as well.

OK, cool... actually still kinda want to buy steamdeck instead, for the sake of playing but will see .

> I don't know what's your stance on 3rd party libraries tho.

Generally better to avoid, unless they do something irreplaceable and in good quality. Unfortunately, `libsdl` however is not something that can be associated with quality.

> have to be implemented into tempest since it handles the event loop

[...]

PR #695, opened 2024-11-15
Update playercontrol.cpp

Schlumpf7:
Outsourcing Gamepad


Try:
Hi, @Schlumpf7 and thanks for PR!

Unfortunately, this cannot be merged into upstream. Generally in project I'm very conservative on third-party libraries following white-list approach essentially. Linking SDL/SDL2 in particular is definitely not acceptable, due to low quality of the library.

Schlumpf7:
What about libgamepad for controller integration?

Try:
As touched in https://github.com/Try/OpenGothic/issues/287, current direction is to take a native api and built on top.
For windows: https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/input/overviews/input-readings
For apple: https://developer.apple.com/documentation/gamecontroller/handling_input_events?language=objc

libgamepad can be used as test-reference/inspiration for api.

Nindaleth:
@Try do you have any preference for Linux as well? Here's an archwiki page on it: [Gamepad](https://wiki.archlinux.org/title/Gamepad)

niklaskorz:
> Linking SDL/SDL2 in particular is definitely not acceptable, due to low quality of the library.

That's a very bold/rude statement that in my experience does not hold. Care to elaborate with sources, not unfounded opinions?
SDL has Valve's backing nowadays and even has become their go-to recommendation for any cross platform input handling. I don't think it's realistic or wise to attempt supporting gamepad input on Linux without it.

Edit: I recommend reading <https://blogs.gnome.org/alicem/2024/10/24/steam-deck-hid-and-libmanette-adventures/> to understand why trying to do everything from scratch using evdev et al is probably a bad idea. The Gnome project's [libmanette](https://gnome.pages.gitlab.gnome.org/libmanette/) is another implementation that supports SDL's widely established control mapping format.

PR #701, opened 2024-11-28
Rudimental Controller support with libgamepad (small C++ lib)

Schlumpf7:
Very basic. Still learning.
Whats not working: Menu Navigation, Looking up/down. 
Whats working: Movment, Sword Combat, interacting, jumping (maybe more?)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "lib/dmusic"]
 	path = lib/dmusic
 	url = https://github.com/GothicKit/dmusic.git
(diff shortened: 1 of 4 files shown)


Try:
Hi, @Schlumpf7 !

As been mentioned in previous PR, thirparty libraries are not welcome, we should stick to native api instead.
Another thing to point out - it's trivial to write code, but before that we have to have agreement on top-level api:
should controller be immediate(async) mode or queue based and why?

Try:
converted to draft, as this PR wont be merged

PR #714, opened 2025-01-11
Handle Gothic 1 guilds

mmind:
I was wondering why I couldn't cast the "kill undead" spell on the Zombie-Keeper in the Stone-Circle when retrieving the magical focus.

Tracked it down to the Zombie-Keeper being of guild "Lurker" ;-) .

I've added the G1 guilds to the enum and added the version check to the Npc::isTargetableBySpell().
This seems to be the only location right now that needs to look at the guild in detail.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1209,7 +1209,9 @@ uint32_t Npc::guild() const {
   }
 
 bool Npc::isMonster() const {
-  return Guild::GIL_SEPERATOR_HUM<guild() && guild()<Guild::GIL_SEPERATOR_ORC;
+  const bool g2 = owner.version().game==2;
+  const auto SEPERATOR_ORC = g2 ? GIL_SEPERATOR_ORC : GIL_G1_SEPERATOR_ORC;
+  return GIL_SEPERATOR_HUM<guild() && guild()<SEPERATOR_ORC;
   }
 
 void Npc::setTrueGuild(int32_t g) {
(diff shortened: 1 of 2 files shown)


game/world/objects/npc.cpp
```diff
@@ -2818,16 +2818,24 @@ void Npc::runEffect(Effect&& e) {
...
   if(bool(t&(TARGET_TYPE_ALL|TARGET_TYPE_NPCS)))
     return true;
+  bool g2 = owner.version().game==2;
   Guild gil = Guild(trueGuild());
   if(bool(t&TARGET_TYPE_HUMANS) && gil<GIL_SEPERATOR_HUM)
     return true;
-  if(bool(t&TARGET_TYPE_ORCS) && gil>GIL_SEPERATOR_ORC)
-    return true;
+  if(bool(t&TARGET_TYPE_ORCS)) {
+    if(g2 && gil>GIL_SEPERATOR_ORC)
```
Try, line 2826: Nitpick:
Can we do, for sake of readability something like:
```c++
const bool g2 = owner.version().game==2;
const auto SEPERATOR_ORC = g2 ? GIL_SEPERATOR_ORC : GIL_G1_SEPERATOR_ORC;
... // other separators
```
and then keep most of original code - otherwise `if-else` tower becomes too complex.

mmind: I've adapted the change for the separator. Question would be, do you also want to do the undead-thing differently? Because there it isn't a single seperator-value but comparing to a bunch of ids.

We could of course do
```
const auto G1_UNDEAD = (gil == GIL_G1_ZOMBIE || ...)
const auto G2_UNDEAD = (gil == GIL_GOBBO_SKELETON || ...)
...
if(bool(t&TARGET_TYPE_UNDEAD) && !g2 && G1_UNDEAD)
  return true;
if(bool(t&TARGET_TYPE_UNDEAD) && g2 && G2_UNDEAD)
  return true;
```
That would even shave one level off the if-tree :-)

Try: I was thinking about:
`const auto SKELETON = g2 ? GIL_SKELETON : GIL_G1_SKELETON`

However your solution looks even better, considering goblin-skeletons.

mmind: Ok thanks ... I'll adapt the change accordingly :-)

Meanwhile I also found the other SEPERATOR_ORC occurence in Npc::isMonster(), that I missed earlier

mmind: I think the updated patches from just now might do it :-) .

Not sure how you're convention is for resolving review comments. Do resolve it as done or do you resolve them as being satisfied? I know projects doing it either way, so just want to find out what you prefer.

Try: > Not sure how your convention is for resolving review comments.

No strict rules, for me it's better to resolve them myself - make it easier to follow

Try:
Red job - apparently is due to we are using old vulkan sdk.

Will merge PR as-is, since change is non-graphics and pretty small. There is also one line with `GIL_SEPERATOR_HUM` in `Npc::checkHealth` - going to adjust it locally.

Thanks!

game/world/objects/npc.cpp
mmind: > > Not sure how your convention is for resolving review comments.
> 
> No strict rules, for me it's better to resolve them myself - make it easier to follow

Good to know - also matches how we do it at my work - so don't even need to switch gears :-D

PR #715, opened 2025-01-11
Implement some function calls used in Gothic 1 scripts

mmind:
GetNextWp is used for bloodflies and wolfes roaming around their paths.
GetInvItemBySlot is used by a number of scripts for a Npc to find a weapon to defend themself.

The other two sounded sensible, but I'm less sure about the implementation, especially if the "in-cutscene" check is correct.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -169,6 +169,7 @@ void GameScript::initCommon() {
   bindExternal("npc_percenable",                 &GameScript::npc_percenable);
   bindExternal("npc_percdisable",                &GameScript::npc_percdisable);
   bindExternal("npc_getnearestwp",               &GameScript::npc_getnearestwp);
+  bindExternal("npc_getnextwp",                  &GameScript::npc_getnextwp);
   bindExternal("npc_clearaiqueue",               &GameScript::npc_clearaiqueue);
   bindExternal("npc_isplayer",                   &GameScript::npc_isplayer);
   bindExternal("npc_getstatetime",               &GameScript::npc_getstatetime);
@@ -2121,6 +2122,14 @@ void GameScript::npc_percdisable(std::shared_ptr<zenkit::INpc> npcRef, int pr) {
 std::string GameScript::npc_getnearestwp(std::shared_ptr<zenkit::INpc> npcRef) {
   auto npc = findNpc(npcRef);
   auto wp  = npc ? world().findWayPoint(npc->position()) : nullptr;
+  if(wp)
+    return wp->name;
+  return "";
+  }
+
+std::string GameScript::npc_getnextwp(std::shared_ptr<zenkit::INpc> npcRef) {
+  auto npc     = findNpc(npcRef);
+  auto wp      = npc ? world().findSecondNearestWayPoint(npc->position()) : nullptr;
   if(wp)
     return wp->name;
   return "";
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -225,6 +225,7 @@ void GameScript::initCommon() {
   bindExternal("npc_isdetectedmobownedbyguild",  &GameScript::npc_isdetectedmobownedbyguild);
   bindExternal("npc_ownedbynpc",                 &GameScript::npc_ownedbynpc);
   bindExternal("npc_canseesource",               &GameScript::npc_canseesource);
+  bindExternal("npc_isincutscene",               &GameScript::npc_isincutscene);
   bindExternal("npc_getdisttoitem",              &GameScript::npc_getdisttoitem);
   bindExternal("npc_getheighttoitem",            &GameScript::npc_getheighttoitem);
   bindExternal("npc_getdisttoplayer",            &GameScript::npc_getdisttoplayer);
@@ -2704,6 +2705,26 @@ bool GameScript::npc_canseesource(std::shared_ptr<zenkit::INpc> npcRef) {
   if(!self)
     return false;
   return self->canSeeSource();
+  }
+
+// Used (only?) in Gothic 1 in B_AssessEnemy, to prevent attacks during cutscenes.
+// Check the global cutscene lock to check if we're in a cinematic
+// Currently there is only the global aiIsDlgFinished() to check for running dialog,
+// combine this a player-check to at least prevent the player from getting attacked.
+// This could use refinement later, to allow per-npc in-dialog checks.
+bool GameScript::npc_isincutscene(std::shared_ptr<zenkit::INpc> npcRef) {
+  auto npc = findNpc(npcRef);
+  auto w = Gothic::inst().world();
+  if(w==nullptr)
+    return false;
+
+  if(w->isCutsceneLock())
+    return true;
+
+  if(npc!=nullptr && npc->isPlayer() && !owner.aiIsDlgFinished())
+    return true;
+
+  return false;
   }
 
 int GameScript::npc_getdisttoitem(std::shared_ptr<zenkit::INpc> npcRef, std::shared_ptr<zenkit::IItem> itmRef) {
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -159,6 +159,7 @@ void GameScript::initCommon() {
   bindExternal("npc_hasitems",                   &GameScript::npc_hasitems);
   bindExternal("npc_hasspell",                   &GameScript::npc_hasspell);
   bindExternal("npc_getinvitem",                 &GameScript::npc_getinvitem);
+  bindExternal("npc_getinvitembyslot",           &GameScript::npc_getinvitembyslot);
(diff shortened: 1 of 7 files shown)


game/game/gamescript.cpp
```diff
@@ -2643,6 +2671,11 @@ bool GameScript::npc_canseesource(std::shared_ptr<zenkit::INpc> npcRef) {
   return self->canSeeSource();
   }
 
+bool GameScript::npc_isincutscene(std::shared_ptr<zenkit::INpc> npcRef) {
+  auto w = Gothic::inst().world();
+  return w->isCutsceneLock();
```
Try, line 2676: `w` might be `nullptr`, if script calling this function from main-menu

game/game/gamescript.cpp
```diff
@@ -2793,6 +2826,12 @@ void GameScript::ai_turntonpc(std::shared_ptr<zenkit::INpc> selfRef, std::shared
     self->aiPush(AiQueue::aiTurnToNpc(npc));
   }
 
+void GameScript::ai_whirlaround(std::shared_ptr<zenkit::INpc> selfRef, std::shared_ptr<zenkit::INpc> npcRef) {
+  // This is supposed to be a fast turn to npcRef,
+  // for now, just also do a regular turn
```
Try, line 2831: Why not implement it as `fast` turn right away?
PS: is my understanding correct, that "fast"=="immediate"?

game/game/inventory.h
```diff
@@ -108,6 +108,7 @@ class Inventory final {
     Item*  currentRangeWeapon()    { return range;      }
     Item*  currentSpell(uint8_t s) { return numslot[s]; }
     const Item*  currentSpell(uint8_t s) const { return numslot[s]; }
+    Item*  getNthCategoryItem(enum InvCategory cat, uint32_t num) const;
```
Try, line 111: Nitpick:
`getNthCategoryItem` -> `findByCategory`, similar to existing `findByClass`.

game/game/inventory.cpp
```diff
@@ -938,6 +938,59 @@ Item *Inventory::findByClass(size_t cls) {
   return nullptr;
   }
 
+Item* Inventory::getNthCategoryItem(enum InvCategory cat, uint32_t num) const {
+  ItmFlags f = ITM_CAT_NONE;
+  switch(cat) {
```
Try, line 943: We probably shouldn't introduce a new `enum` for a single use-case.
It's probably better to move this translation switch-case to `GameScript::npc_getinvitembyslot`, and use `ItmFlags` here.

game/game/gamescript.cpp
```diff
@@ -2126,6 +2143,17 @@ std::string GameScript::npc_getnearestwp(std::shared_ptr<zenkit::INpc> npcRef) {
   return "";
   }
 
+std::string GameScript::npc_getnextwp(std::shared_ptr<zenkit::INpc> npcRef) {
+  auto npc     = findNpc(npcRef);
+  auto nearest = npc ? world().findWayPoint(npc->position()) : nullptr;
+  auto filter  = [&](const WayPoint& p) { return nearest != &p; };
+  auto wp      = nearest ? world().findWayPoint(npc->position(), filter) : nullptr;
+
+  if(wp)
```
Try, line 2152: This function is very similar to `WayMatrix::findNextPoint`. for sake of similarity, probably better to move this code to `WayMatrix`.

Another thing to note, in filter you probably want to discard points that are already taken by other npc's. `p.isLocked()` can achive that.

game/game/gamescript.cpp
mmind: Probably me "just" not knowing enough about what to touch yet ;-) .

For the function I'm going with the description from https://github.com/Try/OpenGothic/wiki/Daedalus-scripts and https://www.worldofgothic.de/modifikation/index.php?go=g1functions ... both saying "fast turn to other".

Looking at the G1 script a great cinematic where it is used is with the Daemon in Xardas tower, where I also need to implement ai_turnaway as well :-) .

```
ai_turnaway(hero, self);
ai_output(hero, self, "Info_XardasDemon_INTRO_15_01");
ai_whirlaround(hero, self);
ai_output(hero, self, "Info_XardasDemon_INTRO_15_02");

```

I guess I'll dig a bit deeper into how that stuff works

game/game/inventory.cpp
mmind: From what I've read the INV_* values were sort of global constants, but you're right, nothing else is using them.
I'll streamline things a bit, like you suggested.

game/game/inventory.h
mmind: findByFlags() might be even more appropriate when taken together with your comment about dropping the "category" thing?

game/game/gamescript.cpp
Try: > I guess I'll dig a bit deeper into how that stuff works

Sounds good. If possible would be nice to have a video on how script above works in vanilla,as a reference.
As for sake of merging, `ai_whirlaround` can be decoupled to different PR

game/game/gamescript.cpp
Try: > ... I did some "nature studies" during my lunch break.

If vanilla doesn't check occupancy - than we good, only need to move code into `WayMatrix`.

game/game/inventory.cpp
mmind: Question: do you prefer num starting at 1 or 0 ? Right now the script "slot" is leaking into the implementation (num = 1 -> first item),

game/game/gamescript.cpp
mmind: I found a Let's play :-D ... https://youtu.be/GKpt0Y07gJw?t=1538 ... when talking to Xardas' daemon, which is this sequence from above

game/game/gamescript.cpp
mmind: I've added a nullptr check

game/game/gamescript.cpp
mmind: This one is still unresolved - hopefully this evening :-)

game/game/gamescript.cpp
mmind: I've actually managed to implement ai_whirlaround (+ai_turnaway) and moved both to a separate pull request.

game/game/inventory.h
mmind: done

game/game/gamescript.cpp
```diff
@@ -2643,6 +2710,13 @@ bool GameScript::npc_canseesource(std::shared_ptr<zenkit::INpc> npcRef) {
   return self->canSeeSource();
   }
 
+bool GameScript::npc_isincutscene(std::shared_ptr<zenkit::INpc> npcRef) {
```
Try, line 2713: I've been looking into what can be expected behavior in this function and can't find good explanation. Meanwhile:
* `zenkit::VCutsceneCamera` - trigger, takes control of camera move for a time
* `zenkit::CutsceneLibrary/CutsceneBlock` - dialogs with npc's as usual

Your implementation handles case 1, but I'm not really sure if this is what needed. Checking [scripts](https://github.com/VaanaCZ/gothic-1-classic-scripts/releases) of G1, only use-case seem to be in `B_AssessEnemy`:
```c++
//-------- Cutscene-Teilnehmer wird das Waffeziehen verziehen --------
if (Npc_IsInCutscene(other)) // Befindet sich der Feind in einer Cutscene?
{	
  PrintDebugNpc	(PD_ZS_CHECK, "...Feind ist in Cutscene!" );				
  return;  //... dann verzeihen wir ihm nat�rlich -> Abbruch des Checks
};
```

There is a `aiIsDlgFinished` to do a general check for dialog (and it's not great at what it doing, as it reaches to UI). 

Note: you don't have to resolve this, as it requires bigger refactoring - just comment in code is fine

game/game/gamescript.cpp
Try: > This one is still unresolved - hopefully this evening :-)
> only need to move code into WayMatrix.

Maybe I need to correct my previous comment: can you move implementation of of point search outside of `GameScript`? `World::findNextFreePoint` is a good reference.

Try (changes requested):

game/game/gamescript.cpp
mmind: Move the implementation to World now, not sure what you think about the function naming, but we'll see :-)

game/game/gamescript.cpp
mmind: I think with the new change, I at least reached half-way :-) .

Using a combination of npc->isPlayer() + aiIsDlgFinished() this should at least prevent the player themself to not get attacked.

I've of course also added a comment explaining the situation and possible later improvement.

Try:
Merged, thanks!

Moved `npc_isincutscene` implementation to https://github.com/Try/OpenGothic/issues/721

PR #716, opened 2025-01-12
Try a MDL resource as alternative for script resources

mmind:
Fixes an error

`loading error: failed to open resource: WASHMOB.MDH`

Because washmob does not need to be split into MHM and MDH files, but instead can also use the MDL type that contains both elements in one file.

If someone stumbles upon "sequel" datafiles by accident, this allows loading the world data and go exploring.

### DIFF
--- a/game/resources.cpp
+++ b/game/resources.cpp
@@ -506,28 +506,42 @@ std::unique_ptr<ProtoMesh> Resources::implLoadMeshMain(std::string name) {
       mdhName = name;
     FileExt::assignExt(mdhName,"MDH");
 
-    const auto* entry = Resources::vdfsIndex().find(mdhName);
-    if(entry==nullptr)
-      throw std::runtime_error("failed to open resource: " + mdhName);
+    if(const auto* entryMdh = Resources::vdfsIndex().find(mdhName)) {
+      // Find a MDH hirarchy file and separate mesh
+      zenkit::ModelHierarchy mdh;
+      auto reader = entryMdh->open_read();
+      mdh.load(reader.get());
 
-    zenkit::ModelHierarchy mdh;
-    auto reader = entry->open_read();
-    mdh.load(reader.get());
+      std::unique_ptr<Skeleton> sk{new Skeleton(mdh,anim,name)};
+      std::unique_ptr<ProtoMesh> t;
 
-    std::unique_ptr<Skeleton> sk{new Skeleton(mdh,anim,name)};
-    std::unique_ptr<ProtoMesh> t;
+      if(const auto* entry = Resources::vdfsIndex().find(mesh)) {
+        auto reader = entry->open_read();
 
-    if(const auto* entry = Resources::vdfsIndex().find(mesh)) {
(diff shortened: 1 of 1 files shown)


game/resources.cpp
```diff
@@ -506,28 +506,47 @@ std::unique_ptr<ProtoMesh> Resources::implLoadMeshMain(std::string name) {
...
+        t.reset(new ProtoMesh(mdm,std::move(sk),name));
+        }
+      else
+        t.reset(new ProtoMesh(mdh,std::move(sk),name));
+
+      return t;
       }
-    else
-      t.reset(new ProtoMesh(mdh,std::move(sk),name));
+      catch(...) {
```
Try, line 533: In opengothic I'm using exceptions only for exceptional cases (aka errors). Based on your PR description, absence of `mdh` is not abnormal case, but just a 'special' one.

It would be better to rewrite `try-catch` part as:
```c+++
if(const auto* entry = Resources::vdfsIndex().find(mdhName)) {
  // MDH path
  } else {
  // MDL fallback
  }
```

Try:
Hi, @mmind and thanks for PR! I left comment in review, and I need bit of time to test it on my end.

CC @lmichaelis , as he is the author of ZenKit library

game/resources.cpp
mmind: I've adapted the code accordingly now to not needlessly throw exceptions around.

game/resources.cpp
```diff
@@ -506,28 +506,43 @@ std::unique_ptr<ProtoMesh> Resources::implLoadMeshMain(std::string name) {
...
-      auto reader = entry->open_read();
+        zenkit::ModelMesh mdm {};
+        mdm.load(reader.get());
+        t.reset(new ProtoMesh(mdm,std::move(sk),name));
+        }
+      else
+        t.reset(new ProtoMesh(mdh,std::move(sk),name));
+
+      return t;
+      } else {
```
Try, line 529: no need to have `} else {` immediately after `return`.
This code can be expressed as:
```c++
if(const auto* entryMdh = Resources::vdfsIndex().find(mdhName)) {
  ...
  }
FileExt::exchangeExt(mdhName,"MDH","MDL");
if(const auto* entryMdl = Resources::vdfsIndex().find(mdhName)) {
  ...
  }
throw std::runtime_error("failed to open resource: " + anim->defaultMesh());
```

mmind: Hopefully fixed now :-) .

mmind:
I guess the broader question is, what is your stance regarding the sequel :-) .

For me it's an interesting historical artifact, so it's fun that OpenGothic can also "just" run it. But there were some more changes necessary for that version "1.12" ... https://github.com/mmind/OpenGothic/tree/dev/gothic1_12 .

I do not plan on implementing all the differences and with the changes from the branch above, it's possible to navigate the new world and interact with things.

But I also can fully understand if you don't want to water down the main OpenGothic with such obscure things ;-) .

Try:
Merged, thanks!

PR #718, opened 2025-01-15
2 additional AI turn actions

mmind:
AI_WhirlAround is doing the same as ai_turntonpc, just faster.
The script does not contain specific values for whirl-speed, so just use
double of the NPc's turn-speed.

AI_turnaway does what the name implies and turns the back to the npc.
There might be a nicer way to calculate that angle than in my code.

### DIFF
--- a/game/game/constants.h
+++ b/game/game/constants.h
@@ -385,6 +385,7 @@ enum Action:uint32_t {
   AI_PrintScreen,
   AI_LookAt,
   AI_WhirlToNpc,
+  AI_TurnAway,
   };
 
 
--- a/game/game/constants.h
+++ b/game/game/constants.h
@@ -383,7 +383,8 @@ enum Action:uint32_t {
   AI_PointAt,
   AI_StopPointAt,
   AI_PrintScreen,
-  AI_LookAt
+  AI_LookAt,
+  AI_WhirlToNpc,
   };
 
 
--- a/game/graphics/mesh/pose.cpp
+++ b/game/graphics/mesh/pose.cpp
@@ -812,7 +812,7 @@ Vec2 Pose::headRotation() const {
   return Vec2(headRotX,headRotY);
   }
 
-void Pose::setAnimRotate(const AnimationSolver &solver, Npc &npc, WeaponState fightMode, int dir) {
+void Pose::setAnimRotate(const AnimationSolver &solver, Npc &npc, WeaponState fightMode, AnimationSolver::TurnType turn, int dir) {
   const Animation::Sequence *sq = nullptr;
   if(dir==0) {
     if(rotation!=nullptr) {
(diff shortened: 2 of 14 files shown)


game/world/objects/npc.cpp
```diff
@@ -1352,6 +1367,16 @@ bool Npc::implTurnTo(float dx, float dz, bool noAnim, uint64_t dt) {
   return rotateTo(dx,dz,step,noAnim,dt);
   }
 
+bool Npc::implWhirlTo(const Npc &oth, uint64_t dt) {
+  if(&oth==this)
+    return true;
+  auto dx = oth.x-x;
+  auto dz = oth.z-z;
+  auto  gl   = guild();
+  float step = float(owner.script().guildVal().turn_speed[gl]) * 2;
```
Try, line 1376: Based on video evidence you shared previously: 
> https://youtu.be/GKpt0Y07gJw?t=1538 ... when talking to Xardas' daemon, which is this sequence from above

Rotation speed appears(not 100% sure) to be the same, but with different animation.

game/world/objects/npc.cpp
```diff
@@ -2191,6 +2216,27 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
...
+      if(interactive()==nullptr) {
+        stopWalkAnimation();
+        visual.stopDlgAnim(*this);
+        }
+      if(act.target!=nullptr && implTurnAway(*act.target,dt)) {
+        queue.pushFront(std::move(act));
+        break;
+        }
+      // Not looking quite correct in dialogs, when npc turns around
+      // Example: Esteban dialog
```
Try, line 2235: Copy-paste?

game/world/objects/npc.cpp
mmind: Ahhhh ... you're right ... thanks for noticing that discrepancy.

Digging deeper, the animation data does have t_surprise_ccw and t_surprise_cw states and G1 data comes uncompiled with "//WhirlAround" comments intact above those :-) .

Additionally, it looks like in G1 everybody including humans have the whirl animation, where in G2, monsters have it - including the newly added ones - but not humans.

So I'll continue digging :-) .

mmind:
With the underlying rotation issues out of the way I think I found a nice way for those two turn variants now.

game/world/objects/npc.cpp
mmind: gone now

game/world/objects/npc.cpp
mmind: People are whirling really nicely now and the code looks quite a bit different :-)

game/graphics/mesh/pose.cpp
```diff
@@ -812,7 +812,7 @@ Vec2 Pose::headRotation() const {
   return Vec2(headRotX,headRotY);
   }
 
-void Pose::setAnimRotate(const AnimationSolver &solver, Npc &npc, WeaponState fightMode, int dir) {
+void Pose::setAnimRotate(const AnimationSolver &solver, Npc &npc, WeaponState fightMode, enum TurnAnim turn, int dir) {
```
Try, line 815: > `enum TurnAnim turn`

In opengothic we are not using `enum` or `struct` as part of function signature. In C++ you don't  need to.

game/game/constants.h
```diff
@@ -519,3 +521,9 @@ enum class AaPreset : uint32_t {
   ULTRA,
   PRESETS_COUNT
   };
+
+enum TurnAnim : uint8_t {
```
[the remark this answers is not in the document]
mmind: done :-)

game/graphics/mesh/pose.cpp
mmind: done too

game/world/objects/npc.cpp
```diff
@@ -3234,23 +3280,43 @@ bool Npc::rotateTo(float dx, float dz, float step, bool noAnim, uint64_t dt) {
...
+        break;
+      case AnimationSolver::TurnType::Whirl:
+        visual.setAnimWhirl(*this,+1);
+        break;
+      }
     setDirection(angle-step);
     } else
   if(sgn>0) {
-    setAnimRotate(noAnim ? 0 : -1);
+    switch(anim) {
```
Try, line 3309: This code might be simplified to cover both positive and negative `sign`, with single switch-case

Try (commented):
looks good, only minor nitpick.

will merge a bit later, as working today on: https://github.com/Try/Tempest/issues/71

Try (approved):

game/world/objects/npc.cpp
mmind: I've adapted the code now to only use one switch

Try:
Merged, thanks!

Issue #646, opened 2024-06-09
[Gothic 1] Game freezes in Xardas' tower


Rez2557:
I'm playing on Fedora Linux, GOG version. I just got to the Xardas' tower in the fourth chapter. When I enter the tower, to the left there is a Lesser Demon, that seems to autotalk with me. When in does that my character and the demon freeze mid-animation, no text pops up, but I can still move the camera up and down with the mouse.
I can't seem to attach my save file, but here's the log.txt:

```
OpenGothic v1.0 dev
no "SystemPack.ini" file in path - using default settings
GPU = AMD Radeon Graphics (RADV RENOIR)
Depth format = Depth32F Shadow format = Depth16
[zenkit] Parsing object [MeshAndBsp % 0 0]
[zenkit] Parsing chunk c000
[zenkit] Parsing chunk c010
[zenkit] Parsing chunk c040
[zenkit] Parsing chunk c045
[zenkit] Parsing chunk c050
[zenkit] Parsing chunk c0ff
[zenkit] Parsing object [VobTree % 0 0]
[zenkit] Parsing object [WayNet % 0 0]
[zenkit] Parsing object [EndMarker % 0 0]
[zenkit] detected invalid use of KEYWORD "ani" in "Model" block. Ignoring rest of script.
unable to load sound fx: WOOD_NIGHT2
[zenkit] detected invalid use of KEYWORD "*eventSFX" in "Model" block. Ignoring rest of script.
[zenkit] Syntax error (line 138, column 55): String not terminated
comb not found: t_FIREPLACE_Stand_2_S0 -> c_FP_STAND_2_S0_(c_FP_STAND_2_S0_2)
comb not found: t_FIREPLACE_S0_2_Stand -> c_FP_STAND_2_S0_(c_FP_STAND_2_S0_2)
[zenkit] Syntax error (line 162, column 56): String not terminated
alias not found: r_Roam1 -> r_Scratch
alias not found: t_SwimF_2_Dive -> t_swim_2_dive
alias not found: t_FistWalkBL_2_FistWalk -> t_FistWalk_2_FistWalkL
[zenkit] Syntax error (line 3186, column 2): comments must start with two slashes
comb not found: t_FIREPLACE_Stand_2_S0 -> c_FP_STAND_2_S0_(c_FP_STAND_2_S0_2)
comb not found: t_FIREPLACE_S0_2_Stand -> c_FP_STAND_2_S0_(c_FP_STAND_2_S0_2)
[...]

PR #720, opened 2025-01-17
Prevent deadlocking the perceptionProcess when perceptiontime is not set

mmind:
It seems some Npc do not get a perception-time set, which lets it stay at 0. If the perceptionProcess runs with that value, it will immediately rerun itself again, potentially leading to a deadlock.

The Npc where this happens in practice is the Firedemon in Xardas' tower in Gothic 1. When the player reaches it, all Npc (and player) motion will stop and with enabled logging, everthing will be flooded with messages about the Firedemon perception process - rerunning the AssessPlayer routine over and over.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -3837,6 +3837,12 @@ bool Npc::perceptionProcess(Npc &pl) {
   if(isPlayer())
     return true;
 
+  // Prevent deadlocking the perception handling
+  if(perceptionTime == 0) {
+    Log::e("Fixing 0 perception time of Npc ",displayName());
+    setPerceptionTime(1000);
+    }
+
   bool ret=false;
   if(processPolicy()!=Npc::AiNormal) {
     perceptionNextTime = owner.tickCount()+perceptionTime;


Try:
Hi, @mmind !

Can you please provide a *.sav file for me to test locally? Thanks!

> setPerceptionTime(1000);

Hm, this might not be great solution. I'm not 100% that overriding per-time will be transparent to script - would like to test locally to see what are possible options.

mmind:
[xardasdemon.zip](https://github.com/user-attachments/files/18470217/xardasdemon.zip)

I've attached a save of Gothic 1, inside Xardas' tower, before the first talk to the demon. So just walk up to it and the game will stall - light keeps on changing though, just both the demon and player completely stop moving.

Yesterday evening I realized that this is probably the thing described in #646

thokkat:
Vanilla is using 5s as default. But changing it to 1s can make sense. Brings it closer to other perc times used and as a player you don't have to wait that long till the demon reacts.

Try:
Fixed by 065ce79b
I've set default value to `5s` and add clamping to minimum of `1s` if script overrides it to smaller value. Value provided by script stored unchanged, mainly just-in-case if it's possible to read this value by something like LeGo.

Thanks @mmind and @thokkat for looking into the root-cause.

PR #723, opened 2025-01-20
Fix AIV_PARTYMEMBER difference between Gothic 1 and Gothic 2

mmind:
The index for AIV_PARTYMEMBER is 15 in Gothic 2, but 36 in Gothic 1, so set it accordingly.


Try:
Thanks for PR @mmind !

Merged.

PR #724, opened 2025-01-20
Protect the renderer against not finding any valid shadow or zbuffer formats

mmind:
While unlikely, there may be cases with broken or not yet feature-complete graphics drivers, where the renderer might not find valid formats for shadows or zbuffers.

When that happens the default values will stay in place and the game will fail later with very obscure exceptions.

The TextureFormat enum in Tempest already has a "Last" element, so use that as a sentinel to catch the no-valid-format case.

As there is no way to recover from this and it shouldn't happen in the first place, throw an exception.

### DIFF
--- a/game/graphics/renderer.cpp
+++ b/game/graphics/renderer.cpp
@@ -46,6 +46,7 @@ Renderer::Renderer(Tempest::Swapchain& swapchain)
     TextureFormat::Depth16,
     TextureFormat::Depth24x8,
     TextureFormat::Depth32F,
+    TextureFormat::Last,
     };
 
   for(auto& i:sfrm) {
@@ -53,18 +54,23 @@ Renderer::Renderer(Tempest::Swapchain& swapchain)
       shadowFormat = i;
       break;
       }
+    if(i==TextureFormat::Last)
+      throw std::runtime_error("no valid shadow format - broken graphics driver?");
     }
 
   static const TextureFormat zfrm[] = {
     TextureFormat::Depth32F,
     TextureFormat::Depth24x8,
     TextureFormat::Depth16,
+    TextureFormat::Last,
     };
   for(auto& i:zfrm) {
     if(device.properties().hasDepthFormat(i) && device.properties().hasSamplerFormat(i)){
       zBufferFormat = i;
       break;
       }
+    if(i==TextureFormat::Last)
+      throw std::runtime_error("no valid zBuffer format - broken graphics driver?");
     }
 
   // crappy rasbery-pi like hardware


mmind:
OpenGothic is a nice complex example I wanted to use to just peek at where the open-source Vulkan driver stack for ARM Mali GPUs is at (panvk as part of panfrost in Mesa) - running vkmark is boring ;-) . And yes that driver is not finished.

Tempest currently checks (all) formats against the flags `VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT|VK_FORMAT_FEATURE_BLIT_DST_BIT|VK_FORMAT_FEATURE_BLIT_SRC_BIT;`

If I'm reading the [profile-definitions correctly](https://vulkan.lunarg.com/doc/sdk/1.3.296.0/linux/profiles_definitions.html),  the `VK_FORMAT_FEATURE_BLIT_DST_BIT` should **not** be in that list at least for `VK_FORMAT_D16_UNORM`
Panfrost provides the format with the correct tilingfeatures, but this of course fails due to the BLIT_DST_BIT thing in Tempest.

This then fails with a completely unhelpful exception when starting a game:
```
---crashlog(std::terminate)---
GPU: Mali-G610 (Panfrost)
#1: std::unexpected() - /lib/aarch64-linux-gnu/libstdc++.so.6(_ZSt10unexpectedv+0) [0xffffab8b0aa4]
#2: std::unexpected() - /lib/aarch64-linux-gnu/libstdc++.so.6(+0xa98e8) [0xffffab8b98e8]
#3: std::unexpected() - /usr/games/opengothic/Gothic2Notr(+0x1408c0) [0xaaaad7d308c0]
#4: Renderer::prepareUniforms() - /usr/games/opengothic/Gothic2Notr(_ZN8Renderer15prepareUniformsEv+0xff0) [0xaaaad7e4c89c]
#5: MainWindow::onWorldLoaded() - /usr/games/opengothic/Gothic2Notr(_ZN10MainWindow13onWorldLoadedEv+0xa8) [0xaaaad7e6a4f0]
#6: Gothic::finishLoading() - /usr/games/opengothic/Gothic2Notr(_ZN6Gothic13finishLoadingEv+0xd0) [0xaaaad7df7f70]
#7: MainWindow::tick() - /usr/games/opengothic/Gothic2Notr(_ZN10MainWindow4tickEv+0x148) [0xaaaad7e6c8b0]
#8: MainWindow::render() - /usr/games/opengothic/Gothic2Notr(_ZN10MainWindow6renderEv+0x5c) [0xaaaad7e7013c]
#9: Tempest::X11Api::implProcessEvents(Tempest::SystemApi::AppCallBack&) - /usr/games/opengothic/libTempest.so(_ZN7Tempest6X11Api17implProcessEventsERNS_9SystemApi11AppCallBackE+0x60)
 [0xffffabbda80c]
#10: Tempest::X11Api::implExec(Tempest::SystemApi::AppCallBack&) - /usr/games/opengothic/libTempest.so(_ZN7Tempest6X11Api8implExecERNS_9SystemApi11AppCallBackE+0x34) [0xffffabbdae54]
#11: main - /usr/games/opengothic/Gothic2Notr(main+0x1d4) [0xaaaad7d31fb4]
```
I did track this down to to something about zbuffer copying not having a valid format.

So from a completenes standpoint I find it way nicer to tell people "your Vulkan implementation is broken", instead of dying later on with obscure exceptions ;-) .

Try:
I've double checked format spec(https://docs.vulkan.org/spec/latest/chapters/formats.html):
You right, `BLIT_DST_BIT` is not guarantied. In principle it's not needed for depth-format - so, we can remove this requirement on the engine side.
`BLIT_SRC_BIT` is used in test-suite, to readback depth

> So from a completenes standpoint I find it way nicer to tell people "your Vulkan implementation is broken", instead of dying later on with obscure exceptions ;-) .

I do agree - it also makes life easier on my end. `crashlog(std::terminate)` - this seem to be due to encounting general exception. Not sure where it is exactly by looking at the crash-log, but if you know - I can add a human readable exception in place of `throw` :)

mmind:
I've adapted the error message a bit, to be more explicit about the broken graphics driver thing, so the crashlog now says
```
---crashlog(std::runtime_error(no valid shadow format - broken graphics driver?))---
GPU: Mali-G610 (Panfrost)
#1: std::unexpected() - /lib/aarch64-linux-gnu/libstdc++.so.6(_ZSt10unexpectedv+0) [0xffffbbec0aa4]
#2: std::unexpected() - /lib/aarch64-linux-gnu/libstdc++.so.6(+0xa98e8) [0xffffbbec98e8]
#3: Renderer::Renderer(Tempest::Swapchain&) - /usr/games/opengothic/Gothic2Notr(_ZN8RendererC2ERN7Tempest9SwapchainE+0xc24) [0xaaaad97f53a8]
#4: MainWindow::MainWindow(Tempest::Device&) - /usr/games/opengothic/Gothic2Notr(_ZN10MainWindowC1ERN7Tempest6DeviceE+0xa4) [0xaaaad9810744]
#5: main - /usr/games/opengothic/Gothic2Notr(main+0x1c0) [0xaaaad96d1f60]
```

mmind:
> I've double checked format spec(https://docs.vulkan.org/spec/latest/chapters/formats.html): You right, `BLIT_DST_BIT` is not guarantied. In principle it's not needed for depth-format - so, we can remove this requirement on the engine side. `BLIT_SRC_BIT` is used in test-suite, to readback depth

Somehow I have the feeling that this is more involved than just removing the flag from the generic(?) imageflags bitfield?

Try:
Part 1:
I've added `BadTextureCastException` to the engine. Now exception will be more readable:
```
---crashlog(BadTextureCastException(Format "Depth16" is not sampling capable))---
```

Try:
Original issue should be fixed by [16f2143](https://github.com/Try/OpenGothic/commit/16f214366b643332116f93100f0e919ac10dacc7)

mmind:
> Root cause should be fixed by [16f2143](https://github.com/Try/OpenGothic/commit/16f214366b643332116f93100f0e919ac10dacc7)

Yep, format detection works now correctly and without exceptions.
And the BadTextureCastException also is a nicer safeguard.

So I guess we can close this merge-request :-) .

Issue #722, opened 2025-01-20
isPlayer() in Npc::Npc constructor will always be false


mmind:
The Npc constructor does
```
  // vanilla behavior: equip best weapon and set non-zero damage type
  if(!isPlayer())
    invent.autoEquipWeapons(*this);

```
with
```
bool Npc::isPlayer() const {
  return aiPolicy==Npc::ProcessPolicy::Player;
  }

```
But npc->setProcessPolicy(...) is only set after the constructor has finished, see
```
void World::createPlayer(std::string_view cls) {
...
  npcPlayer = wobj.insertPlayer(std::move(npc),waypoint);
  npcPlayer->setProcessPolicy(Npc::ProcessPolicy::Player);
...
  }

```
So right now all Npcs will not be recognized as player, but this works, because the player has an empty inventory at the beginning of the main games.

thokkat:
Good find. I proposed a fix in https://github.com/Try/OpenGothic/pull/725.

PR #725, opened 2025-01-20
Fix a non-working player check

thokkat:
As described in https://github.com/Try/OpenGothic/issues/722 we don't know if `aiPolicy==Player` in npc constructor which means best weapon equip has to be done later.

One of the `addNpc` functions already does this indirectly by calling `restPositionToTA`. The other one is never called for the player so weapon equip can be done here safely as well. This covers the case when marvin insert command is used which was the original motivation to do this in the constructor, see https://github.com/Try/OpenGothic/pull/653#issuecomment-2243476855.

Regarding the mentioned summons, I modified the wolf spell to summon the hero instead. Weapon was equipped while script had only a `createInvItem` line.

Fixes https://github.com/Try/OpenGothic/issues/722

### DIFF
--- a/game/world/worldobjects.cpp
+++ b/game/world/worldobjects.cpp
@@ -322,8 +322,7 @@ Npc* WorldObjects::addNpc(size_t npcInstance, const Vec3& pos) {
   Npc* npc   = new Npc(owner,npcInstance,pstr);
   npc->setPosition  (pos.x,pos.y,pos.z);
   npc->updateTransform();
-  if(owner.version().game==2)
-    owner.script().invokeRefreshAtInsert(*npc);
+  owner.script().invokeRefreshAtInsert(*npc);
 
   npcArr.emplace_back(npc);
   return npc;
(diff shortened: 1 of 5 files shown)


game/world/worldobjects.cpp
```diff
@@ -322,6 +322,7 @@ Npc* WorldObjects::addNpc(size_t npcInstance, const Vec3& pos) {
   Npc* npc   = new Npc(owner,npcInstance,pstr);
   npc->setPosition  (pos.x,pos.y,pos.z);
   npc->updateTransform();
+  npc->autoEquipWeapons();
```
Try, line 325: Another override for `WorldObjects::addNpc(size_t npcInstance, std::string_view at)` seem to be missing.

Try (commented):
Hi, @thokkat and thanks for the fix - one of overrides of `addNpc` appears to be missing.

game/world/worldobjects.cpp
thokkat: I left it out because this function calls `restPositionToTA` where it's already done. You want to have it added for consistency with other `addNpc` function?

Additionally I found that`B_RefreshAtInsert` is called as well and added it.

Try: > You want to have it added for consistency with other addNpc function?

Ideally, I think, we need to make it robust and leave no room for inconsistency. After giving more thought to it: maybe it's even better to add `ProcessPolicy` as a parameter to `npc` constructor. This will make any explicit/implicit use of `isPlayer` correct.

Try:
Merged, thanks!

PR #727, opened 2025-01-23
Implement Gothic1 behaviour for missing items

mmind:
The Gothic 2 script provides callback functions when the inventory of a npc is empty or too far away, the player is missing a key for a chest or another npc is using an object.

Vanilla Gothic 1 is less sophisticated and will just play the shoulder shrug animation (T_DONTKNOW). There is no textual output displayed in vanilla g1 for any of the mentioned cases, although the script seems to contain at least an unused text definition _STR_MESSAGE_INTERACT_NO_KEY for the chest-without-key/-lockpick case.

Implement this behaviour as a fallback when the engine can't find the symbol in the script, as right now the player character just does nothing.

The T_DONTKNOW animation is also the animation played in Gothic 2 from inside the callback functions.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -973,48 +973,60 @@ void GameScript::printCannotBuyError(Npc &npc) {
 
 void GameScript::printMobMissingItem(Npc &npc) {
   auto id = vm.find_symbol_by_name("player_mob_missing_item");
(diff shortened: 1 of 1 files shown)


Try:
I've changed it a bit: now builtin animation is player only for G1. I don't know for sure how exactly G2 react to absence of callback

PR #728, opened 2025-01-23
Implement opening the map via hotkey in Gothic 1

mmind:
Where Gothic 2 has a script function player_hotkey_screen_map() that gets called on the map-hotkey and is expected to provide the map to use, Gothic 1 is doing this in-engine.

In Gothic 1 this was of course easy, as there was only one world and experimenting with Vanilla Gothic 1 showed that the hotkey function seems ignores all maps except ITWRWORLDMAP and ITWRWORLDMAP_ORC - world maps with and without the orc-territory addition. The map with the orc-addition takes precedence over the regular map.

Also these two items are the only ITWR* items contained in the actual original Gothic.exe.

So implement a G1 specific script function that gets called when the player_hotkey_screen_map() cannot be found and we're on in Gothic 1.

### DIFF
--- a/game/game/gamescript.cpp
+++ b/game/game/gamescript.cpp
@@ -1169,10 +1169,28 @@ CollideMask GameScript::canNpcCollideWithSpell(Npc& npc, Npc* shooter, int32_t s
   return CollideMask(vm.call_function<int>(fn, spellId));
   }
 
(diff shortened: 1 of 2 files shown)


Try:
Merged, thanks!

PR #734, opened 2025-01-29
Implement npc_sendsingleperc

mmind:
The worldofgothic function list describes the function [0] as "send perception to a single Npc" and in another forum post [1] the function also is described as "make target react to * from caster"

So it seems correct to directly start the perception process of the target npc for the percid.

This function is for example used in Gothic 1 when the a monster should react to the shrinking spell.

This function does not seem to be used in Gothic 2.

[0] https://www.worldofgothic.de/modifikation/article_371_2.htm#231
[1] https://forum.worldofplayers.de/forum/threads/1544378-How-to-create-VFX-with-separated-origin-and-instigator?p=26210828&viewfull=1#post26210828


Try:
Looks good to me, thanks @mmind !

Issue #18, opened 2019-08-10
Registrate damage from falling on ground


Try:
1. Need to implement this
1.1 Make sure Immortal Npc shouldn't receive damage at all 
2 Fall animations looks alright already, but need to verify

PR #735, opened 2025-01-30
Don't set victim when picking up an item from the floor

### LINKED ISSUE #717: Does item pickup have a victim (picking up items marked as theft)
In Gothic 1 every time I pick up an item from the ground (i.e. a berry at the startpoint) while a Npc looks, he'll shout "Behind you!".

Tracking that down to b_assesstheft(), I can see:
```
    ai_quicklook(self, other);
    if (npc_canseenpc(self, other)) {
        printdebugnpc(pd_zs_check, "...NSC kann den Dieb sehen!");
        if (hlp_isvaliditem(item) && (!hlp_isvalidnpc(victim))) {
            printdebugnpc(pd_zs_check, "...Item wurde aufgehoben!"); // Item picked up
    ...
        }
        else {
            printdebugnpc(pd_zs_check, "...Taschendiebstahl!"); //pickpocket
            if (...) {
    ...
                ai_pointatnpc(self, other);
                b_say(self, other, "$BEHINDYOU");

```
So the condition (hlp_isvaliditem(item) && (!hlp_isvalidnpc(victim)) should be true to indicated something picked up, but isn't, leading to the pickpocket branch.

I'm not yet sure what part is wrong though.

mmind:
At least the Gothic 1 script assesses theft differently depending on whether a victim is set or not.

```
if (npc_canseenpc(self, other)) {
        printdebugnpc(pd_zs_check, "...NSC kann den Dieb sehen!");
        if (hlp_isvaliditem(item) && (!hlp_isvalidnpc(victim))) {
            printdebugnpc(pd_zs_check, "...Item wurde aufgehoben!"); // Item picked up
            ...
        }
        else {
            printdebugnpc(pd_zs_check, "...Taschendiebstahl!"); //pickpocket
            ...
        }
```
This causes NPC around an item to tell the player that something is getting stolen right now and also indicates that victim is not always set in the perception process.

So adapt the sent passivePerc accordingly to not transmit the player as victim when picking up items.

GameScript::invokeState() already removed the default setting of other=npc for the !other case in commit 3fc5c39e2095 ("damage from faling") as part of the fix for #18, so do the same for the current default victim setting.

This fixes #717.

### DIFF
--- a/game/world/world.cpp
+++ b/game/world/world.cpp
@@ -690,12 +690,20 @@ Bullet& World::shootBullet(const Item &itm, const Npc &npc, const Npc *target, c
   return b;
   }
 
-void World::sendPassivePerc(Npc &self, Npc &other, Npc &victim, int32_t perc) {
+void World::sendPassivePerc(Npc &self, Npc &other, int32_t perc) {
+  wobj.sendPassivePerc(self,other,nullptr,nullptr,perc);
+  }
+
+void World::sendPassivePerc(Npc &self, Npc &other, Npc* victim, int32_t perc) {
   wobj.sendPassivePerc(self,other,victim,nullptr,perc);
   }
 
-void World::sendPassivePerc(Npc &self, Npc &other, Npc &victim, Item& item, int32_t perc) {
+void World::sendPassivePerc(Npc &self, Npc &other, Npc* victim, Item& item, int32_t perc) {
   wobj.sendPassivePerc(self,other,victim,&item,perc);
+  }
+
+void World::sendPassivePerc(Npc &self, Npc &other, Item& item, int32_t perc) {
+  wobj.sendPassivePerc(self,other,nullptr,&item,perc);
   }
 
 void World::sendImmediatePerc(Npc& self, Npc& other, Npc& victim, int32_t perc) {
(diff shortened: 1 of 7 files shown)


game/world/world.cpp
```diff
@@ -690,12 +690,16 @@ Bullet& World::shootBullet(const Item &itm, const Npc &npc, const Npc *target, c
   return b;
   }
 
+void World::sendPassivePerc(Npc &self, Npc &other, Item& item, int32_t perc) {
+  wobj.sendPassivePerc(self,other,nullptr,&item,perc);
+  }
+
 void World::sendPassivePerc(Npc &self, Npc &other, Npc &victum, int32_t perc) {
-  wobj.sendPassivePerc(self,other,victum,nullptr,perc);
+  wobj.sendPassivePerc(self,other,&victum,nullptr,perc);
```
Try, line 698: Since original assumption about `victim` cannot be null is now know to be not valid, we also need to update code that call to his function, such as:
```
// Interactive::implTick
npc.world().sendPassivePerc(npc,npc,npc,PERC_ASSESSUSEMOB);


```

Try (commented):
Hi, @mmind !

Path looks good to me, but need to be updated due to https://github.com/Try/OpenGothic/pull/736

game/world/world.cpp
mmind: Hopefully I caught all of them now.

One interesting decision I had a hard time with, was wether or not to have the World::sendPassivePerc() variants.
Before the victim change, there were two sendPassivePerc variants, hiding the the possible item nullptr from users.

I've gone now with introducing similar variants for victim and victimless variants, but this already increases variants to 4.

The other option would be of course to just add the nullptr when calling World::sentPassivePerc(). I'll defer that to your preference :-) .

Try:
Merged, thanks!

Issue #523, opened 2023-10-12
Revisit `aiQueue` and `aiQueueOverlay`


Try:
Based on discussion in discord
https://discord.com/channels/989316194148433950/1006242446860107888/1161805115079008266

`AI_PointAt` should act as overlay, and we also have some similar cases in code:
`AI_LookAtNpc`, `AI_LookAtNpc`, `AI_PointAtNpc` - those are paired with similar stop-function

Example:
```
func void PC_Psionic_TRAIN_Info()
{
    AI_Output(self, other, "PC_Psionic_TRAIN_Info_05_01"); // output of dialog-line
    AI_PointAt(self, "PSI_PATH_9_14");                     // start pointing to point (overlay queue)
    AI_StopPointAt(self);                                  // <wait for "PC_Psionic_TRAIN_Info_05_01">; stop pointing 
    AI_Output(self, other, "PC_Psionic_TRAIN_Info_05_02"); // <wait for "PC_Psionic_TRAIN_Info_05_01">; output of next dialog-line
    ...
};
```

PR #743, opened 2025-02-15
Implement ai_quicklook

mmind:
The quicklook function is meant to have the Npc just move its head to the player for two seconds.

Testcase in vanilla Gothic 1 was at the exchange place, while Diego walks back to the old camp. Stand next to where he is walking and pick up one of the berries, arrows, etc. Diego will continue walking but turn his head over for those 2 seconds.

So the implementation does the same, not changing the currentLookAtNpc but storing the temporary one in a separate variable for the duration of the action. That way this cinematic should not influence any other parts or save/load mechanisms.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -1274,17 +1274,19 @@ bool Npc::implPointAt(const Tempest::Vec3& to) {
   }
 
 bool Npc::implLookAtWp(uint64_t dt) {
-  if(currentLookAt==nullptr)
+  // nothing to look at, or a quicklook is handled in implLookAtNpc
+  if(currentLookAt==nullptr || quickLookNpc!=nullptr)
     return false;
   auto dvec = currentLookAt->position();
   return implLookAt(dvec.x,dvec.y,dvec.z,dt);
   }
 
 bool Npc::implLookAtNpc(uint64_t dt) {
-  if(currentLookAtNpc==nullptr)
+  if(currentLookAtNpc==nullptr && quickLookNpc == nullptr)
     return false;
   auto selfHead  = visual.mapHeadBone();
-  auto otherHead = currentLookAtNpc->visual.mapHeadBone();
+  auto other = (quickLookNpc!=nullptr) ? quickLookNpc : currentLookAtNpc;
+  auto otherHead = other->visual.mapHeadBone();
   auto dvec = otherHead - selfHead;
   return implLookAt(dvec.x,dvec.y,dvec.z,dt);
   }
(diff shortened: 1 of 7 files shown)


game/world/objects/npc.cpp
```diff
@@ -3091,9 +3113,19 @@ void Npc::setToFistMode() {
...
 
 void Npc::aiPush(AiQueue::AiAction&& a) {
-  if(a.act==AI_OutputSvmOverlay)
-    aiQueueOverlay.pushBack(std::move(a)); else
-    aiQueue.pushBack(std::move(a));
+  switch(a.act) {
+    case AI_OutputSvmOverlay:
+      aiQueueOverlay.pushBack(std::move(a));
+      break;
+    case AI_QuickLook:
```
Try, line 3120: This looks like a partial workaround for:
https://github.com/Try/OpenGothic/issues/523

My understanding, that some commands, such as `AI_QuickLook` has to run in dedicated queue, with own command-level tracking

game/world/objects/npc.h
```diff
@@ -596,6 +596,9 @@ class Npc final {
     Npc*                           currentOther   =nullptr;
     Npc*                           currentVictim  =nullptr;
 
+    Npc*                           quickLookNpc=nullptr;
```
Try, line 599: Should it be mutually exclusive with `currentLookAtNpc` and occupy same variable?

Try (commented):
>  not changing the currentLookAtNpc but storing the temporary ... not influence any other parts or save/load mechanisms.

Unfortunately, you still need to store those in save file, as it's a part of npc state. Here opengothic is somewhat different to vanilla, by keeping as much state of the game as possible in the save.

game/world/objects/npc.cpp
```diff
@@ -3091,9 +3113,19 @@ void Npc::setToFistMode() {
...
+      // QuickLook is supposed to last for 2 seconds
+      // and should not interrupt other AI actions like walking
+      quickLookEnd = owner.tickCount()+2000;
```
Try, line 3123: State change should always occur when executing `Npc::nextAiAction`. Since `AI_QuickLook` should overlap with walking, `aiQueueOverlay` has to be utilized.
`

mmind:
> > not changing the currentLookAtNpc but storing the temporary ... not influence any other parts or save/load mechanisms.
> 
> Unfortunately, you still need to store those in save file, as it's a part of npc state. Here opengothic is somewhat different to vanilla, by keeping as much state of the game as possible in the save.

ha, you're right ... just checked by causing a quicklook and saving while it is ongoing. After loading, Diego keeps the turned head forever - he must get quite the neck-pain from that :-)

PR #745, opened 2025-02-17
Fix ammo checking for best weapons

thokkat:
Ranged weapon auto equip would accidentally unequip current equipped weapon if no best weapon could be found. 

In Cavalorn case this is because script function `EquipItem` doesn't check ammo availability but the `AI_BestWeapon` functions do. So the weapon is first equipped and later unequipped because no best weapon could be found. 

The fix is to just do nothing in this case. For consistency this has been added for armor and melee weapons too but should never be triggered.  I added a second commit to bring variable/function naming in line with script functions.

### DIFF
--- a/game/game/inventory.cpp
+++ b/game/game/inventory.cpp
@@ -116,7 +116,7 @@ void Inventory::implLoad(Npc* owner, World& world, Serialize &s) {
   s.read(stateSlot.slot);
   stateSlot.item = readPtr(s);
 
-  armour = readPtr(s);
+  armor  = readPtr(s);
   belt   = readPtr(s);
   amulet = readPtr(s);
   ringL  = readPtr(s);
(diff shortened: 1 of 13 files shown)


mmind:
Tiny nitpick: this nicely written explanation on how and why that fix works, would be really great as commit message in the first commit, because the merge-request conversation is way less accessible than a "git log" down the road.

game/game/inventory.cpp
```diff
@@ -116,7 +116,7 @@ void Inventory::implLoad(Npc* owner, World& world, Serialize &s) {
   s.read(stateSlot.slot);
   stateSlot.item = readPtr(s);
 
-  armour = readPtr(s);
+  armor = readPtr(s);
```
Try, line 119: alignment around ` = ` seem to be disturbed

game/game/inventory.cpp
```diff
@@ -460,7 +460,7 @@ bool Inventory::setSlot(Item *&slot, Item* next, Npc& owner, bool force) {
 void Inventory::updateView(Npc& owner) {
   auto& world = owner.world();
 
-  updateArmourView(owner);
+  updateArmorView(owner);
```
Try, line 463: `updateArmorView(owner);` -> `updateArmorView (owner);` to match lines bellow

Try (approved):
LGTM, can you please rebase on newer `master`, for sake of precommit please?

Also - some nitpicks in review about white-spaces (probably result of CTRL+R).

thokkat:
Whitespace misalignments should be resolved and commit message is now more detailed.

Try:
Merged, thanks!

Issue #637, opened 2024-05-24
Move trigger Problems


thokkat:
Some movers in sleeper temple have wrong behavior like finishing in wrong state, looks like same root cause. Examples taken from https://github.com/Try/OpenGothic/discussions/622. Wanted to report them as well.
 13.1 is fixed by https://github.com/Try/OpenGothic/pull/636.

> 13.2 Death trap with roof coming down, roof falls instantly, allows no escape. Timing issue.
> savegame: [save_slot_4.sav.tar.gz](https://github.com/Try/OpenGothic/files/15431672/save_slot_4.sav.tar.gz)
> no ticket created
> video 1 (using lever): , 4 sec, 7.3 MiB: https://github.com/Try/OpenGothic/assets/86957596/550fabdd-b436-4b5c-93bb-5d55a2519366
> video 2 (using lever): , 4 sec, 3.0 MiB: https://github.com/Try/OpenGothic/assets/86957596/ad49b3a6-b1e1-4df5-95a8-409c901e790a

> 13.3 Sleeper Temple Elevator floor clipping, causing player to fall to the ground into the elevating block
> savegame: [save_slot_11.sav.tar.gz](https://github.com/Try/OpenGothic/files/15432133/save_slot_11.sav.tar.gz)
> no ticket created
> video: 4 sec, 7.9 MiB: https://github.com/Try/OpenGothic/assets/86957596/7cedf0b4-6928-4b6e-8b98-57aff7dfac68

> 13.4 Uriziel room entry trap does not withdraw/pull back
> savegame: [save_slot_12.sav.tar.gz](https://github.com/Try/OpenGothic/files/15433204/save_slot_12.sav.tar.gz)
> no ticket created
> screenshot: [spike trap now withdrawn](https://github.com/Try/OpenGothic/assets/86957596/28299b4f-0273-400a-b350-5f5a70fc7b6a)

> 13.5 Sleeper Temple Arrow Target Puzzle - trap does not relax and ranged target clipping (and hitbox) from outside possible
> savegame: [save_slot_7.sav.tar.gz](https://github.com/Try/OpenGothic/files/15433561/save_slot_7.sav.tar.gz)
> no ticket created
> screenshot 1: [trap not withdrawn and aiming, even hitbox of ranged button from outside accessible](https://github.com/Try/OpenGothic/assets/86957596/9d1ffe81-8fae-4a90-a2f6-d4dae7a3a207)
[...]

PR #767, opened 2025-04-15
Fixing some mover bugs

thokkat:
Swapping `Close` with `Open` fixed  issues:

- **13.2** Death trap with roof coming down `goto waypoint TPL_044`
- **13.4** Uriziel room entry trap  `goto waypoint TPL_191`
- **13.5** Sleeper Temple Arrow Target Puzzle `goto waypoint TPL_085`

in sleeper temple from  https://github.com/Try/OpenGothic/issues/637.

Same issue has been fixed before for `TRIGGER_CONTROL` type movers in https://github.com/Try/OpenGothic/commit/2194ac2d935d0e1ff23ce4250db94a32df6add13

---

If there are multiple movers with same name an infinite retrigger loop could arise on a second  trigger.  Preliminary solution is to never retrigger if mover is in-non idle state. Problematic location: `goto waypoint TPL_062.`

Testing suggests that events are executed immediately and never retriggered e.g if a pillar moves up and the down switch is triggered the movement direction would change instantly. Couldn't come up with something working in general so far, I leave that for a follow-up PR.

### DIFF
--- a/game/world/triggers/movetrigger.cpp
+++ b/game/world/triggers/movetrigger.cpp
@@ -133,7 +133,10 @@ void MoveTrigger::onGotoMsg(const TriggerEvent& evt) {
 void MoveTrigger::processTrigger(const TriggerEvent& e, bool onTrigger) {
   if(moverKeyFrames.size()==0)
     return;
+  if(state==Loop)
+    return;
   if(state!=Idle) {
+    // FIXME: can lead to infinite amount of events if multiple movers with same name exist
     world.triggerEvent(e);
     return;
     }
--- a/game/world/triggers/movetrigger.cpp
+++ b/game/world/triggers/movetrigger.cpp
@@ -142,16 +142,16 @@ void MoveTrigger::processTrigger(const TriggerEvent& e, bool onTrigger) {
   switch(behavior) {
     case zenkit::MoverBehavior::TOGGLE: {
       if(frame==0) {
-        state = Close;
-        } else {
         state = Open;
+        } else {
+        state = Close;
         }
       break;
       }
     case zenkit::MoverBehavior::TRIGGER_CONTROL: {
       if(onTrigger)
-        state = Close; else
-        state = Open;
+        state = Open; else
+        state = Close;
       break;
       }
     case zenkit::MoverBehavior::OPEN_TIME: {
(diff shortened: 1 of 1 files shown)


game/world/triggers/movetrigger.cpp
```diff
@@ -134,7 +134,8 @@ void MoveTrigger::processTrigger(const TriggerEvent& e, bool onTrigger) {
   if(moverKeyFrames.size()==0)
     return;
   if(state!=Idle) {
-    world.triggerEvent(e);
+    // can lead to infinite amount of events if multiple movers with same name exist
+    // world.triggerEvent(e);
```
Try, line 138: While I do agree, that this code need to be changed, but we still need to be aware of case, when 2 triggers do overlap. 
Logic here is to postpone `e` until current trigger is done execution. Maybe we can use `delayedEvent` here instead.

game/world/triggers/movetrigger.cpp
```diff
@@ -155,7 +156,7 @@ void MoveTrigger::processTrigger(const TriggerEvent& e, bool onTrigger) {
       break;
       }
     case zenkit::MoverBehavior::OPEN_TIME: {
-      state = Open;
+      state = Close;
```
Try, line 159: This look very suspicious.
While such change may fix some particular cases, in overall it looks wrong to do `Close`, instead of open and vise-versa

Try (commented):

Can you confirm if those changes didn't affect Irdorath and Adanos temples in G2in a negative way?

thokkat:
> Can you confirm if those changes didn't affect Irdorath and Adanos temples in G2in a negative way?

Good thought, lazily I only tested newworld but luckily there's only one location more for G2.

- spike trap on bandit guild treasure island
- gate in a cave next to dexter's house
- gates at the side where undead dragon is in Irodorath

All work with pr but not before.

game/world/triggers/movetrigger.cpp
Try: > I think this works because Open and Close do the opposite thing in tick function

That would mean, tick `tick` function is wrong and need to be changed, not state <--> MoverBehavior relation.

At this point probably would be fine to do auto-rename, such as
1. Close -> Else
2. Open -> Close
3. Else -> Open

game/world/triggers/movetrigger.cpp
Try: Bump

Try (commented):
Tested locally on G2 - all good, still need to address review comments

thokkat:
Never retrigger for looping movers only for now because it doesn't do anything anyway. I would submit a proper solution in a future pr including missing sounds like `sfx_close_start`.

Try:
Merged, thanks!

PR #783, opened 2025-07-24
Fix undeclared pthread_setname_np() error

carlo-bramini:
I tried to compile OpenGothic but it stopped with this error:
```
game/utils/workers.cpp: In static member function ‘static void Workers::setThreadName(const char*)’: 
game/utils/workers.cpp:62:3: error: ‘pthread_setname_np’ was not declared in this scope
   62 |   pthread_setname_np(pthread_self(), threadName);
      |   ^~~~~~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```
According to:

https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html

the macro `_GNU_SOURCE` is required to be defined for using `pthread_setname_np()`.
Often, this macro is already defined, but in my case it is not, evidently.
So I did this quick patch and I fixed the problem.
It looks like other files include `pthread.h`, so `_GNU_SOURCE` must be declared on top on the file and not where `workers.cpp` includes `pthread.h`.

### DIFF
--- a/game/utils/workers.cpp
+++ b/game/utils/workers.cpp
@@ -1,8 +1,3 @@
-// Required by pthread_setname_np()
-#if defined(__GNUC__) && !defined(_GNU_SOURCE)
-#define _GNU_SOURCE
-#endif
-
 #include "workers.h"
 #include "utils/string_frm.h"
 
@@ -15,6 +10,11 @@
 #endif
 
 #if defined(__GNUC__)
+// Required by pthread_setname_np()
+#if !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+
 #include <pthread.h>
 #endif
 
--- a/game/utils/workers.cpp
+++ b/game/utils/workers.cpp
@@ -1,3 +1,8 @@
+// Required by pthread_setname_np()
+#if defined(__GNUC__) && !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+
 #include "workers.h"
 #include "utils/string_frm.h"
 


Try:
Thanks for PR @carlo-bramini , good catch!

Try:
I moved code around a bit, and merged it

carlo-bramini:
Thank you very much for the quick reply.

Unfortunately, your change made the patch not working anymore.
As I had written in my first comment, `pthread.h` is already included by something (I don't know where), so `_GNU_SOURCE` must be declared before the first inclusion, that's why I did it on top of the file.
It is also true that including `pthread.h` is redundant into `workers.cpp` because it is already included somewhere.
However it could be left, if it improves the readability of the code.

Try:
Hi, 

> so _GNU_SOURCE must be declared before the first inclusion, that's why I did it on top of the file.
Sorry, I didn't give enough attention to that part

After checking the sources - `pthread.h` is not really included explicitly anywhere, might be included indirectly by `<thread>` or other part of std.

Can you please check [d187cf0](https://github.com/Try/OpenGothic/commit/d187cf011faca88e24dc738308a8649c452df57e), if it helps? - I've decided to add `_GNU_SOURCE` globally at CMake level.

carlo-bramini:
> Can you please check [d187cf0](https://github.com/Try/OpenGothic/commit/d187cf011faca88e24dc738308a8649c452df57e), if it helps? - I've decided to add `_GNU_SOURCE` globally at CMake level.

I just cloned the latest sources from the repo and now it works perfectly.
Thank you very much!

PR #784, opened 2025-07-24
Fix Character Rotation When Stationary & Improve Dialogue Skipping Sync

rockysx27:
## Summary

Fixes two small but impactful issues related to player control and dialogue system:

### 1. Prevent Character Rotation When Stationary
- Updated `PlayerControl::assignRunAngle` to check for zero or negative movement on the X/Z axes.
- Prevents rotation updates when the player is idle but in attack or finishing animation states.
- This avoids unwanted snapping/clipping while stationary.

### 2. Improve Dialogue Skip Sync
- In `DialogMenu::skipPhrase`, added calls to `stopDlgAnim()` for both player and NPC.
- Ensures animations are stopped when skipping dialogue, keeping lip-sync and timing consistent.

## Files Modified
- `game/game/playercontrol.cpp`
- `game/ui/dialogmenu.cpp`

Let me know if anything needs adjustment!

### DIFF
--- a/game/ui/dialogmenu.cpp
+++ b/game/ui/dialogmenu.cpp
@@ -379,12 +379,14 @@ void DialogMenu::startTrade() {
 
 void DialogMenu::skipPhrase() {
   if(current.time>0) {
-    if(pl!=nullptr)
-      pl->setAiOutputBarrier(0,false);
+    if (pl != nullptr) {
+      pl->setAiOutputBarrier(0, false);
       pl->stopDlgAnim(); //skip pl-dialog phrase
-    if(other!=nullptr)
-      other->setAiOutputBarrier(0,false);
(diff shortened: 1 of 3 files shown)


Try:
Hi, @rockysx27 !

Can you please separate this PR into 2 different ones? It's not good to have 2 orthogonal things in a same commit.

game/ui/dialogmenu.cpp
```diff
@@ -381,8 +381,10 @@ void DialogMenu::skipPhrase() {
   if(current.time>0) {
     if(pl!=nullptr)
       pl->setAiOutputBarrier(0,false);
+      pl->stopDlgAnim(); //skip pl-dialog phrase
```
Try, line 384: missing `{` `}` around code-block

PR #786, opened 2025-07-24
Update dialogmenu.cpp - Fixed-Lip-Sync-When-SkipPhrase

rockysx27:
Fixes lip sync when skipping dialogues the lips sync now looks really good.

https://youtu.be/RpaaUUMX6r8

### DIFF
--- a/game/ui/dialogmenu.cpp
+++ b/game/ui/dialogmenu.cpp
@@ -379,10 +379,14 @@ void DialogMenu::startTrade() {
 
 void DialogMenu::skipPhrase() {
   if(current.time>0) {
-    if(pl!=nullptr)
-      pl->setAiOutputBarrier(0,false);
-    if(other!=nullptr)
-      other->setAiOutputBarrier(0,false);
+    if (pl != nullptr) {
+      pl->setAiOutputBarrier(0, false);
+      pl->stopDlgAnim(); //skip pl-dialog phrase
+    }
+    if (other != nullptr) {
+      other->setAiOutputBarrier(0, false);
+      other->stopDlgAnim(); //skip other-dialog phrase
+    }
     currentSnd   = SoundEffect();
     current.time = 1;
     }


Try:
Hi, @rockysx27 !

> https://youtu.be/RpaaUUMX6r8

Hm, this video look like opengothic and not vanilla and in the issue, I've asked about how it is looking in original game.
You sure, that video is correct?

PS:
It's fine on github to drag-drop small videos to comment - not necessary to go thru youtube upload

rockysx27:
Here is how the original game looks. The animations are transitioned not stopped but its better than them playing when the npc stopped speaking...
https://youtu.be/Tej9ehMn3uQ

Also, it shows how the runAngle should look like. In your version runAngle get applied still when you are stationary. If you don't want to merge my pulls delete them.
PS:
My videos are HQ and can't be uploaded to GitHub.

PS:
I found another bug, in your version the bandit has a spliff and in original gothic he doesn't, its only given to him as an invisible item so his animations can play out

Try:
Hi @rockysx27 !

> https://youtu.be/Tej9ehMn3uQ

Video is unavailable, due to restricted access... 

> The animations are transitioned not stopped but its better than them playing when the npc stopped speaking.

Usually it's better to follow vanilla and make behavior as similar as possible: usually it's hard to agree between many people on what is better - so keeping it same, unless there is a very strong reason for change.

Try:
> I found another bug, in your version the bandit has a spliff and in original gothic he doesn't

I've check this one. (assuming `spliff` is "A stalk of swampweed").
Npc's in the game have a random items in inventory. In case of bandits, there are 7 different sets for script to choose from. Sets number `2` and `4` do have a spliff.


> given to him as an invisible item so his animations can play out

Game has no invisibility status attached to specific items. Only items that are in use (runes usually) are hidden from ransack menu.

PR #788, opened 2025-07-24
Update playercontrol.cpp - Fixed-RunAngle-when-Stationary

rockysx27:
https://youtu.be/hhdAGLwb2Yk

### DIFF
--- a/game/game/playercontrol.cpp
+++ b/game/game/playercontrol.cpp
@@ -1031,6 +1031,16 @@ void PlayerControl::quitPicklock(Npc& pl) {
   }
 
 void PlayerControl::assignRunAngle(Npc& pl, float rotation, uint64_t dt) {
+  //Only assign run angle if player is running (to prevent weird angles when stationary)
+  if(pl.isInState(ScriptFn()) && pl.isAttackAnim() && !pl.isFinishingMove()) {
+    runAngleDest = 0;
+    return;
+  }
+  if (pl.animMoveSpeed(0).x < 0 || pl.animMoveSpeed(0).z < 0) {
+    runAngleDest = 0;
+    return;
+  }
+  
   float dtF    = (float(dt)/1000.f);
   float angle  = pl.rotation();
   float dangle = (rotation-angle)/dtF;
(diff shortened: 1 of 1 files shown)


game/game/playercontrol.cpp
```diff
@@ -1031,6 +1031,16 @@ void PlayerControl::quitPicklock(Npc& pl) {
   }
 
 void PlayerControl::assignRunAngle(Npc& pl, float rotation, uint64_t dt) {
+  //Only assign run angle if player is running (to prevent weird angles when stationary)
+  if(pl.isInState(ScriptFn()) && pl.isAttackAnim() && !pl.isFinishingMove()) {
```
Try, line 1035: Same check for null-state is already exists at line 661:
```
  if(!pl.isInState(ScriptFn()) || dlg.isActive()) {
    runAngleDest = 0;
    return;
    }
```

game/game/playercontrol.cpp
```diff
@@ -1031,6 +1031,16 @@ void PlayerControl::quitPicklock(Npc& pl) {
...
+    runAngleDest = 0;
+    return;
+  }
+  if (pl.animMoveSpeed(0).x < 0 || pl.animMoveSpeed(0).z < 0) {
```
Try, line 1039: There is already and explicit check for `ani==Npc::Anim::Move`, in place where function is called.
Also, calling `animMoveSpeed` with delta-time equals zero - this condition has no chance to be `true`.

game/game/playercontrol.cpp
rockysx27: > Same check for null-state is already exists at line 661:
> 
> ```
>   if(!pl.isInState(ScriptFn()) || dlg.isActive()) {
>     runAngleDest = 0;
>     return;
>     }
> ```

Well then, your 'null-state' check didn't work as intended. Because when you call it at 661 it doesn't set it to 0 as assignRunAngle function overrides the 0 value set at 661.

Try: After checking in your latest code:
Yes, I've been wrong when calling check same - they actually inverse of one-another due to `!`

Looking at your code case-by-case:
* `pl.isInState(ScriptFn())` == true. Practically always.
* `pl.isAttackAnim()` attack animation is not a movement and will result into `dangle==0` anyway.
* `isFinishingMove` is special-cased for some reason (why?)

So only thing that been changed is handling of `runAngleSmooth` timeout in very indirect and confusing way.

At this point my question: what is intent behind this code or what exact use-case suppose to be solved and why?

PR #789, opened 2025-07-28
Fix helper npc bugs

thokkat:
Helper npcs like Characterhelper don't have a valid waypoint but still have their state started in vanilla. This means a check  is not required in `Npc::currentRoutine` as this would only lead to a default routine. State start is needed to have talk perc enabled.

Waypoint for helpers routine is "XXX" which makes them walk to the bandit cave at the beginning of Gothic 2. In vanilla they don't walk at all. Reason is that waypoint search in `GameScript::ta_min` allows an inexact match. Npc's added by insert command get a waypoint/freepoint close to spawn location assigned if there is one. Could still lead to some walking but this is only a small nitpick.

### DIFF
--- a/game/marvin.cpp
+++ b/game/marvin.cpp
@@ -610,7 +610,7 @@ std::string_view Marvin::completeInstanceName(std::string_view inp, bool& fullwo
       }
     if(name.size()>match.size())
       name = name.substr(0,match.size());
-    for(size_t i=0; i<match.size(); ++i) {
+    for(size_t i=0; i<name.size(); ++i) {
       int n = std::tolower(name [i]);
       int m = std::tolower(match[i]);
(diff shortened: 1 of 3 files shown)


game/marvin.cpp
```diff
@@ -610,7 +610,7 @@ std::string_view Marvin::completeInstanceName(std::string_view inp, bool& fullwo
       }
     if(name.size()>match.size())
       name = name.substr(0,match.size());
-    for(size_t i=0; i<match.size(); ++i) {
+    for(size_t i=0; i<name.size(); ++i) {
```
Try, line 613: nit: better to segregate those into separate small pr, please :)

Try:
Merged, thanks @thokkat !

PR #790, opened 2025-07-28
Remove extra check for z-distance in waypoint search

thokkat:
This fixes some peasants close to Lefty in new camp who can now find their freepoint.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -2320,9 +2320,9 @@ void Npc::nextAiAction(AiQueue& queue, uint64_t dt) {
         }
       break;
     case AI_StartState:
-      // only daily routine states have a scheduled end time
-      // assigning finite end-time to ZS_TALK/ZS_ATTACK can cause bugs, when dialog or combat is interrupted by time-table routine
-      if(startState(act.func,act.s0,gtime::endOfTime(),act.i0==0)) {
+      // NOTE: a new state can be stater within a daly routiine, such as TA_Sleep, with: ZS_GotoBed -> ZS_Sleep.
+      // In such cases it's important to preserve aiState.eTime.
+      if(startState(act.func,act.s0,aiState.eTime,act.i0==0)) {
         setOther(act.target);
         setVictim(act.victim);
         }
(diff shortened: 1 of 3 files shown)


Try:
Looks good - I'll will test it a bit, before merge, since G2 is quite sensitive to way-points

Try (approved):

thokkat:
I played a bit around and manipulated script to only have relevant fp functions inside state loop. Value for `distanceThreshold` seems to be like ~800 regardless of gothic version. Doesn't change much for G2 though.

Try:
Currently somewhat blocked in testing.
My plan was to accelerate time and check that npc's way-point in city are consistent day-2-day. Unfortunately it's not possible: trader (and others) now sleep indefinitely.

Infinite sleep is not caused by this pr, but by one of your previous ones: https://github.com/Try/OpenGothic/pull/648
Need to figure how-to solve infinite sleep first :|

thokkat:
Actually this pr was a side effect while testing a fix for a G1 sleep problem. Wrong state end time was one of the issues :D.

If daily routine calls AI_StartState this new state is then treated like new routine and keeps the same end time. But states started from percs have infinite end time.

Try:
> If daily routine calls AI_StartState this new state is then treated like new routine and keeps the same end time. But states started from percs have infinite end time.

Sounds over-complicated. Basically you suggesting to unroll call-stack of the script to find how `AI_StartState` was originated?

In [711d073](https://github.com/Try/OpenGothic/pull/790/commits/711d0732b00df5a036d95e6f23b3802f81174abb) I've pushed a different solution: basically to threat attacking behavior or speaking as blockers for state termination.

thokkat:
My current solution is to check current endtime before AI_StartState is pushed to queue

```C++
self->aiPush(AiQueue::aiStartState(st.funcIni,self->isFiniteState(),state,oth,vic,wp));
```
and set infinite endtime  before perc execution
```C++
    auto time = aiState.eTime;
    aiState.eTime = gtime::endOfTime();
    owner.script().invokeState(this,&pl,victim,fn);
    aiState.eTime = time;
```

Try:
> My current solution is to check current endtime before AI_StartState is pushed to queue

That still basically means 2 different behaviors for a queue-function, based on context. Choosing between it and special casing - I think it's more likely, that vanilla have had a special case version.

Try:
Apparently Zuris can't find right place to stand in morning.
UPD:
appear to be not directly related to PR - debugging it now. https://github.com/Try/OpenGothic/issues/791

Try:
Vatras not going to pray at night, with `distanceThreshold = 800`. Need to be larger:
image

thokkat:
> Vatras not going to pray at night, with distanceThreshold = 800. Need to be larger:

Tested, Vatras doesn't go to freepoint in vanilla either. Did a rebase and changed back to current value `900` as that is a good  compromise already.

>  I think it's more likely, that vanilla have had a special case version.

Sure simpler solution is better

Try:
> Sure simpler solution is better

PR, looks good to me - should I merge?

thokkat:
Yes, best non-squashed like the last one.

Try:
Merged, thanks!

PR #792, opened 2025-07-31
hardcode allowed player states

thokkat:


Try:

Uhh, it's not cool that vanilla has such terrible piece of code deep inside... If it wont be game breaking, would like not to implement such.
Can we have a code for G2/G1, but effectively disable it in G2? AFAIK magic freeze works OK, for gothic2 and having while-list of states can be an issue for mod-makers.

thokkat:
Adding a clearstate to the magic freeze workaround could be done. This effectively already does the same as name filtering because the only use case I'm aware of is to not put player in some attack state after a magic state ends.
```C++

void GameScript::ai_startstate(std::shared_ptr<zenkit::INpc> selfRef, int func, int state, std::string_view wp) {
  auto self = findNpc(selfRef);
  if(self!=nullptr && func>0) {
    Npc* oth = findNpc(vm.global_other());
    Npc* vic = findNpc(vm.global_victim());
    if(!self->isInState(ScriptFn()) && self->isPlayer()) {
      // avoid issue with B_StopMagicFreeze
      self->aiPush(AiQueue::aiStandup());
      self->clearState(state); // <-- make sure current state ends
      return;
      }
```

> AFAIK magic freeze works OK, for gothic2 

Player stays in freeze animation because `AI_StandUp` is later removed by a clear queue in `B_AssessDamage` before attack state is started. Not a problem for npcs because `ZS_Attack` has `AI_StandUp` inside. So question is if you want to keep `self->aiPush(AiQueue::aiStandup());`  as it's actually a script bug which shows in vanilla as well.

```bash
func void B_StopMagicFreeze()
{	
	Npc_PercEnable	(self, PERC_ASSESSMAGIC, B_AssessMagic);
	
	Npc_ClearAIQueue(self);
	AI_StandUp		(self);

	if (self.guild < GIL_SEPERATOR_HUM)
	{
		B_AssessDamage();
	}
	else
	{
		Npc_SetTempAttitude (self, ATT_HOSTILE); //falls nicht schon Gilden-Attitüde hostile ist 
	};
	
	// nach Aufruf dieses Befehles wird die Loop über return LOOP_END beendet
};
```

Try:
OK, lets do name filter for both G1, G2

thokkat:
Addressed review comments.

Try:
I've changed use of `st.name()` - this is debug only function, and not intended to be used in actual code.
Also changed initialization of `playerEnabledStates` - static initialization based on game version means, we wont be capable to unload G2 and start G1, in the same runtime and vise-versa.

Other than thet - LGTM, merged.

thokkat:
> I've changed use of `st.name()` - this is debug only function, and not intended to be used in actual code.

What about storing symbols in `AiState` instead of indexes to have automatic name access?
>we wont be capable to unload G2 and start G1, in the same runtime and vise-versa.

Is this needed somewhere? I thought you have to close and restart if you want to run a different gothic.

Try:
> What about storing symbols in AiState instead of indexes to have automatic name access?

No really needed in any parts of code, and `find_symbol_by_index` is as simple as array lookup anyway.

> Is this needed somewhere?

For future use-cases: can be potentially helpful for level-editor

PR #794, opened 2025-08-20
macOS/Metal: gate VSM 3D storage texture allocation to fix Metal validation crash

arturkasperek:
### Summary
- **Fix**: Avoid early creation of Virtual Shadow Maps (VSM) 3D storage textures on macOS/Metal unless VSM is enabled and supported.
- **Why**: Prevents startup crash due to Metal rejecting linear 3D textures.

### Error
```text
_mtlValidateStrideTextureParameters:1824: failed assertion `Texture Descriptor Validation
Linear texture: must have depth == 1, depth (16) disallowed
Linear texture: must be of type MTLTextureType2D or linear MTLTextureType2DArray, textureType (MTLTextureType3D) disallowed
```

### Root cause
- On Metal where native image atomics aren’t available (MSL < 3.1), storage images fall back to buffer‑backed “linear” textures.
- Metal allows linear 2D/2D array only; a linear 3D texture (for VSM page table) triggers validation and aborts.

### Impact
- **macOS (VSM off by default)**: No performance/visual change; startup crash resolved.
- **Supported platforms with VSM on**: Unchanged; VSM resources still allocate on demand.
- **Unsupported macOS configs**: Keep VSM off (or run with `-vsm 0`).

### How to test
1. Build on macOS and run: `./Gothic2Notr.sh -g "<path-to-Gothic2>"`.
2. Expect normal startup (no Metal validation assert).
3. Optionally test with `-vsm 0` to force-disable VSM.

### DIFF
--- a/game/graphics/renderer.cpp
+++ b/game/graphics/renderer.cpp
@@ -172,7 +172,8 @@ void Renderer::resetSwapchain() {
     vsm.pageListTmp = StorageBuffer();
 
     //FIXME: used internally by DrawCommands
-    vsm.pageTbl  = device.image3d(TextureFormat::R32U, 32, 32, 16);
+    if(settings.vsmEnabled && Shaders::isVsmSupported())
+      vsm.pageTbl  = device.image3d(TextureFormat::R32U, 32, 32, 16);
   }
 
   // rtsm


arturkasperek:
@Try can you check?

Try:
Hi, @arturkasperek !

Interesting finding, what GPU are you using? For me - I've been playing with VSM on M1 for a while and it works.

> Metal allows linear 2D/2D array only;

Can't actually find any confirmation with MSL/Metal documentation, only indirect information, such as your validation error and MoltenVK code. Need to know what GPU family we need to exclude to start working on a workaround.

Try:
I've pushed a proper fix to `//FIXME: used internally by DrawCommands`, in [4bbd5c6](https://github.com/Try/OpenGothic/commit/4bbd5c67c9bbc1777f13c817ffd1a43a94f09a8c) - now it's safe to create page-table on demand.

Still need to investigate, why Metal validation complains.

arturkasperek:
I tested on MacBook pro 2016

Try:
@arturkasperek since there is now proper workaround on `master`, should we close this PR?

Also it seem, that there is no real solution to the problem, and we would have to deprecate Metal2 or have a pretty-much time-bomb in engine.

PR #796, opened 2025-08-24
CsCamera refactor and smooth phasing

Try:
Large refactor of CsCamera.

* Simplify code
* Most of things are now stateless
  * animation state no longer depends of `dt`
  * player 'god' state handled properly
* Added debug-draw for splines
* Remove old motion-scaling

While there is intent to make camera smooth, this are of time-demo is still an issue:
image

### DIFF
--- a/game/world/triggers/cscamera.cpp
+++ b/game/world/triggers/cscamera.cpp
@@ -2,7 +2,7 @@
 
 #include <Tempest/Log>
 #include <cassert>
-
+#include "utils/dbgpainter.h"
 #include "gothic.h"
 
 using namespace Tempest;
@@ -22,17 +22,18 @@ float CsCamera::KeyFrame::arcLength() const {
   return len;
   }
 
-CsCamera::KbSpline::KbSpline(const std::vector<std::shared_ptr<zenkit::VCameraTrajectoryFrame> >& frames, const float duration, std::string_view vobName) {
+
+CsCamera::KbSpline::KbSpline(const std::vector<std::shared_ptr<zenkit::VCameraTrajectoryFrame>>& frames, const float duration, std::string_view vobName) {
   keyframe.reserve(frames.size());
   for(auto& f : frames) {
     KeyFrame kF;
     kF.c[3]  = Vec3(f->original_pose[3][0],f->original_pose[3][1],f->original_pose[3][2]);
     kF.time  = f->time;
+    kF.motionType = f->motion_type;
     keyframe.push_back(kF);
     }
 
   const size_t size = keyframe.size();
-  float pathLength = 0;
   for(size_t i=0; i+1<size; ++i) {
     auto& kF  = keyframe[i];
     Vec3  p0  = i==0 ? kF.c[3] : keyframe[i-1].c[3];
@@ -44,8 +45,6 @@ CsCamera::KbSpline::KbSpline(const std::vector<std::shared_ptr<zenkit::VCameraTr
     kF.c[0] =  2 * p1 - 2*p2 +   dd + sd;
     kF.c[1] = -3 * p1 + 3*p2 - 2*dd - sd;
     kF.c[2] = std::move(dd);
-
-    pathLength += kF.arcLength();
     }
 
   if(size<2)
@@ -58,10 +57,9 @@ CsCamera::KbSpline::KbSpline(const std::vector<std::shared_ptr<zenkit::VCameraTr
     Log::e("CsCamera: \"", vobName, "\" - invalid sequence duration");
     }
 
-  const float ref    = 1.f/std::max(pathLength, 0.0001f);
   const float slow   = 0;
-  const float linear = duration * ref;
-  const float fast   = 2 * duration * ref;
+  const float linear = duration;
+  const float fast   = 2 * duration;
 
   const zenkit::CameraMotion mType0 = frames.front()->motion_type;
   const zenkit::CameraMotion mType1 = frames.back() ->motion_type;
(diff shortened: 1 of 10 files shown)


Try:
@thokkat may you have a quick look, since CsCamera is originally done by you?

Also I've been thinking that at some point camera was smoother, but wasn't able to find a revision where it was.

thokkat:
I think it is because progress was based on arclength initially. Means if 20% of rescaled time passed camera position was at 20% length (green curve) instead of the spline evaluated at t=0.2 (blue curve). Orange is measured vanilla average.

now

This non-smoothness became even worse if target curve was a spline and not just a point.

Vanilla is using [14] in https://splines.readthedocs.io/en/latest/euclidean/catmull-rom-properties.html#Wrong-Tangent-Vectors.
I haven't considered this back then because it's described there as bad.

This change gives the vanilla curve:
```
float dt  = kF2.time-kF1.time;
Vec3  dd  = (p2-p0) * dt/(kF2.time-kF0.time);
Vec3  sd  = (p3-p1) * dt/(kF3.time-kF1.time);
```
---
I have a branch with support for other motion types and some minor changes, can sent a pr after this one is merged. I extend the pre-pr approach of using a hermite spline and using derivatives 0 for slow, 2 for fast, and 1 for everything else. Mentioned it because it could slightly change fps numbers for bechmarking.

game/world/triggers/cscamera.cpp
```diff
@@ -44,8 +45,6 @@ CsCamera::KbSpline::KbSpline(const std::vector<std::shared_ptr<zenkit::VCameraTr
     kF.c[0] =  2 * p1 - 2*p2 +   dd + sd;
     kF.c[1] = -3 * p1 + 3*p2 - 2*dd - sd;
     kF.c[2] = std::move(dd);
-
-    pathLength += kF.arcLength();
```
thokkat: arclength function  is unused now.

Try:
> I think it is because progress was based on arclength initially. .. Orange is measured vanilla average.

Is this chart a camera speed at given point of time, from 0..30 seconds, for time-demo?
I've also experimented with uniform movement, with `arclength`, (ignoring time-point) - still non-smooth.


> This change gives the vanilla curve:
```
float dt  = kF2.time-kF1.time;
Vec3  dd  = (p2-p0) * dt/(kF2.time-kF0.time);
Vec3  sd  = (p3-p1) * dt/(kF3.time-kF1.time);
```
Em this change need to go where?)

> I extend the pre-pr approach of using a hermite spline and using derivatives 0 for slow, 2 for fast, and 1 for everything else

Sounds very promising: I've removed hermite, as it contradicts exact timestamps in key-frames.

game/world/triggers/cscamera.cpp
Try: Intentional - let's keep arcLength for now.
Also this implementation is very trivial - it just tessellates curve into 100 linear segments and sum their length.

thokkat:
> Is this chart a camera speed at given point of time, from 0..30 seconds, for time-demo?

yes

>Em this change need to go where?)

Where spline segments are calculated in kbspline constructor

> I've removed hermite, as it contradicts exact timestamps in key-frames.

you mean that node has like time 5s but camera would only reach at 7s?

Try:
> Where spline segments are calculated in kbspline constructor

Aha, done!

> you mean that node has like time 5s but camera would only reach at 7s?

Each `KeyFrame` in spline has exact time code. If we would apply any kind of time-scale before resolving frame (like in previous implementation), that would lead into mismatch between real-time and key-frame.
So, IMAO time modifier suppose to be applied only after, key-frame is resolved (see `CsCamera::KbSpline::position`)

thokkat:
Yes generally true but SMOOTH is the exception. This node is ignored except if it's a start or end node and then treated like LINEAR. Otherwise speed would be constant except first and last segment for timedemo.

two more minor review things:

Spline curve can be shown via `ztoggle showsplines`. Could be added separately together with waynet/freepoints marvin commands.

`active` variable can be removed and replaced by a function that queries `ticksEnabled`.

Try:
> Yes generally true but SMOOTH is the exception

But `LINEAR` are still exact right? So in theory, there can be `LINEAR` -> `SMOOTH` -> `SMOOTH` -> `LINEAR`. Maybe we can estimate speed of consecutive `SMOOTH` points, and recalculate their timings in setup?

> Spline curve can be shown via ztoggle showsplines.

Debug draw here only for debugging, not intended as marvin-feature

> active variable can be removed and replaced by a function that queries ticksEnabled.

done

thokkat:
> But LINEAR are still exact right?

Yes. Only STEP is a special case. Behaves like LINEAR if it's the start node of a segment, if endnode camera would jump to that position and wait until total time reaches the node time. Used in sleeper temple.

> recalculate their timings in setup

Problem is FAST node where in timedemo speed is accelerated for the first 15s and then deaccelerated till the end. With pre-calculation speeds per segment were different but constant. So it would be like a piecewise linear approximation, but I don't know if this would be a problem visually-wise.

Just fyi what I had in mind. Instead of the actual keyframe determine the closest previous/next non-smooth and calculate position from that time. Speed can be 0/1/2 like mentioned.

```C++
  float t   = float(time)/1000.f;
  auto  kF1 = &keyframes.front();
  auto  kF2 = &keyframes.back();
  for(auto& kF:keyframes) {
    if(kF.motionType==CameraMotion::UNDEFINED || kF.motionType==CameraMotion::CUSTOM || kF.motionType==CameraMotion::SMOOTH)
      continue;
    if(kF.time>t) {
      kF2 = &kF;
      break;
      } else {
      kF1 = &kF;
      }
    }

  if(kF1==kF2 || kF2->motionType==CameraMotion::STEP)
    return kF2->time;
  float d1      = kF1->speed();
  float d2      = kF2->speed();
  float dt      = kF2->time - kF1->time;
  float u       = (t - kF1->time) / dt;
  float tScaled = (((d1 + d2 - 2)*u -2*d1 - d2 + 3)*u + d1)*u*dt;
  return kF1->time + tScaled;
```

Try:
> Just fyi what I had in mind.

Let's merge this PR as-is, so you can look into `motionType`'s and speed :)
I can take care of rest of benchmark mode in the meantime

thokkat:
Sounds good. Non-smothness is gone or still a problem?

Try:
Smoothness is good since this code is added https://github.com/Try/OpenGothic/pull/796#issuecomment-3218386510

PR #806, opened 2025-09-19
Fix build issue; Cast correct type

versable:


Try:
Thanks for report!

`static_cast<T>` is not very welcome in project, so I've changed line in question to
`setFRate(CommandLine::inst().isBenchmarkMode()!=Benchmark::None)` 

In any case, thanks for pointing out, specially now,  when CI is down

Issue #713, opened 2025-01-11
Enabling / disabling subtitles


SovietMouse:
Hey there,

Is there a plan to allow possibility to enable / disable subtitles?

Reading whole sentence before it is being said aloud tempts me to skip the dialogue. Disabling subtitles in original game helped me stay immersed in the world (: so I'd be happy to see that feature in OpenGothic too!

Try:
Hi, @SovietMouse 

Yes, we actually missing all 4 toggle here:
```
[GAME]

subTitles=1
; ... choose if you want to see windows with spoken words (1) or not (0), default is 0

subTitlesAmbient=1
; ... set to 1 if you dont want to have subtitles for ambient talks (default: 1) [disabled if subTitles is off]

subTitlesPlayer=1
; ... activates (1) or deactivates (0) subtitles for the hero (default: 1) [disabled if subTitles is off]

subTitlesNoise=0
; ... activates (1) or deactivates (0) subtitles for surroundings npc ambient infos and talks (default: 0) [disabled if subTitles is off]

```

And systempack.ini:
```
[SUBTITLES]
Control=0
TimeMultiplier=2.0
MaxTimePerPhrase=30.0
TimePerChar=100.0
```

Shouldn't be too hard to implement, and most of the work can be handled in `dialogmenu.cpp`.

UPD:
`GAME / subTitles` - done by [#809](https://github.com/Try/OpenGothic/pull/809)

lezeSoftware:
Hello,
i would like to participate and try to solve this issue :)

Try:
@lezeSoftware 
Please do - PR is very welcome :)

PR #810, opened 2025-10-12
enable flag to turn on/off player subtitles

Nindaleth:
here's an addition of a player-specific subtitles toggle (another part of #713)

### DIFF
--- a/game/ui/dialogmenu.cpp
+++ b/game/ui/dialogmenu.cpp
@@ -373,9 +373,9 @@ bool DialogMenu::haveToWaitOutput() const {
   return false;
   }
 
-bool DialogMenu::isShowingSubtitles() const {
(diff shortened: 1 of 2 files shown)


Try:
Hi, @Nindaleth !

Patch generally look good. Would you mind to refactor condition `showSubtitles && (showSubtitlesPlayer || !curentIsPl)` into separate function, forsake of readability?

Nindaleth:
I've split out the condition into a separate function. I assumed it could stay private and a `isSomethingTrue` naming approach for bool-returning function is OK here, but please comment further if I should improve it.

Try:
> I assumed it could stay private and a isSomethingTrue naming approach

`is` generally fine; Changed to haveTo* notation, since it's not a property, but a check. 
Merged!

PR #822, opened 2025-10-22
Delay trigger sent by mobsi interaction

thokkat:
Matches vanilla behavior where a CsCamera sequence is only started after the switch animation to next state has ended.


Try:
Merged, thanks!

PR #823, opened 2025-10-22
Relax a condition for mover speed calculation

thokkat:
Distance between keyframe positions doesn't have to be exactly zero to use rotation speed, it seem there is some small threshold. Otherwise the boat close to Lares in L'Hiver mod would jump between two keyframes.

In general boats in L'Hiver require lower speeds than mover speed value suggests whereas in non-moded game higher values are needed. There's even one going completely haywire. I have no solution for this and leave it as is for now.


Try:
I've add some comments, so it will be more clear from looking at the code, why there is a threshold.

Merged, thanks!

Issue #638, opened 2024-05-24
[Gothic 1] Fleeing Npcs have player as goto target


thokkat:
If a digger is brought to <50% health he begins to flee. Instead of running away he is following the player.

Problem is that `AI_Wait` in script loop is interrupting the turning process, causing this weird behavior. In vanilla it looks like fleeing is following the waynet and using same logic as npc going to a different location. A solution could be to handle fleeing in `Npc::implGoTo` where it's currently a nop.
```C++
  if(go2.flag==GT_Flee) {
    // nop
    }
```

PR #824, opened 2025-10-22
Fix fleeing for G1 npcs

thokkat:
G1 script makes npcs flee if the their health drops below a certain percentage in a fight. As in many ZS loop functions there's a AI_Wait call in ZS_Flee.

The Engine is calling `Npc::adjustAttackRotation` function everytime a npc has to wait.
```C++
  if(waitTime>=owner.tickCount() || aniWaitTime>=owner.tickCount() || outWaitTime>owner.tickCount()) {
    if(!isPlayer() && faiWaitTime<owner.tickCount())
      adjustAttackRotation(dt);
```
 Because fleeing npcs have a target and a drawn weapon a rotation is executed. The rotation added from AI_Flee is overwritten and the npc follows the player instead of running away.

This is fixed by adding a check if npc is attacking to `Npc::adjustAttackRotation`. I used the same check as in `Npc::implAttack`. I’m not sure if there is a difference to `isAttack()` here.

fixes https://github.com/Try/OpenGothic/issues/638

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -2172,7 +2172,7 @@ void Npc::tick(uint64_t dt) {
     }
 
   if(waitTime>=owner.tickCount() || aniWaitTime>=owner.tickCount() || outWaitTime>owner.tickCount()) {
-    if(!isPlayer() && faiWaitTime<owner.tickCount())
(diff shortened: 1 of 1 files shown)


Try:
will cancel CI runs today, due to until #830 resolved

Try:
Merged, thanks!

PR #834, opened 2025-11-07
Update frames.cpp to fix compilation on WinARM64

ArminiusTux:
as mentioned in #827

This change should not affect other OSes and platforms.🤞


Try:
LGTM.

One last question: have I understood correctly, that thi PR only enables compilation and crash-log is not expected to be correct?

ArminiusTux:
Yes indeed, it's a compilation fix only.

Try:
Merged, thanks!

PR #839, opened 2025-11-19
Change name check in waypoint search

thokkat:
Byproduct of https://github.com/Try/OpenGothic/pull/831. Npcs would go to closest waypoint even if that had empty name.

Testing showed `AI_GoToPoint` wants exact name match but `AI_GoToFp` only does a loose check. Even single characters like "_" or "S" made the npc go to a freepoint if its name contained them.

Added a comment to range threshold indicating vanilla value is buggy (found in https://github.com/Try/OpenGothic/pull/790#issuecomment-3140773921) and early return in `AI_GoToPoint` to avoid two commands being sent to ai queue..


Try (approved):

Try:
Looks good,
Thanks for your investigation about way/free points!

PR #846, opened 2025-12-13
strafe while running

AIG-Livny:
just little improvement. Allow strafe until running. Strafe now have higher priority. It allows press A or D while W already pressed

### DIFF
--- a/game/marvin.cpp
+++ b/game/marvin.cpp
@@ -63,7 +63,7 @@ Marvin::Marvin() {
     {"set var %v %s",              C_SetVar},
 
     // animation
-    {"play ani %s",                C_Invalid},
+    {"play ani %s",                C_PlayAni},
     {"play faceani %s",            C_Invalid},
     {"remove overlaymds %s",       C_Invalid},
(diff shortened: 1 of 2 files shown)


game/marvin.cpp
```diff
@@ -63,7 +63,7 @@ Marvin::Marvin() {
     {"set var %v %s",              C_SetVar},
 
     // animation
-    {"play ani %s",                C_Invalid},
+    {"play ani %s",                C_PlayAni},
```
Try: Addition of `play ani` command seem to be orthogonal to strafe itself. Generally, I'm fine merging both things in one PR - just double checking

Try (commented):
LGTM. 
A small question: this strafe behavior is something what exists in vanilla as well?

AIG-Livny:
> LGTM. A small question: this strafe behavior is something what exists in vanilla as well?

Yes, this is exactly vanilla behavior. I checked it before commit

Try:
Merged, thanks!

Issue #872, opened 2026-01-18
DEB package


versable:
I'd be willing to build up a build system for a DEB package, as suggested by @Try [here](https://github.com/Try/OpenGothic/issues/127#issuecomment-792359603).

I was a [repository maintainer](https://launchpad.net/~versable/+archive/ubuntu/elementary-update) before, and am familiar with the ecosystem.

However, this is not as trivial as suspected, as I would like to do it the recommended way. This would mean I would need to split each dependency and mark them as a required package.

I'd need to start with Tempest, which depends on (Debian/Ubuntu packages):

- OpenAL (libopenal-dev)
- libpng (libpng-dev)
- metal-cpp (no package, header only, no need for a separate package)
- SPIRV-Cross (spirv-cross)
- Squish (libsquish-dev)
- zlib (zlib1g-dev)

After that I would need to tackle ZenKit. Bullet is already packaged in Debian (zlib1g-dev).

Alternatively I could just create a package as is from the existing build system, which would never get pulled upstream but would lower the entry barrier for Debian/Ubuntu (and derivatives) users.

Try:
Hi, @versable !

Can you please clarify why you have to split dependency? Is it requirement, for package to be accepted?
Ideally I would like to have all dependencies to be statically linked with Tempest, to match my windows setup. Also, worth to mention: in some cases local libraries (spirv-cross) specifically out of sync with main on purpose. Usually it to accommodate bug-fixes immediately, without waiting on them to be merged.

versable:
@Try 

It's Debian policy to split dependencies, see https://www.debian.org/doc/debian-policy/ch-sharedlibs.html (Which is why some people prefer Snap or Flatpaks, the upside of the split dependencies is that bug fixes tend to get pushed upstream quicker).

I am curious, for the release assets, do you manually add the files?

I will start to get the DEB build in as is for now, without splitting the dependencies.

Try:
> I am curious, for the release assets, do you manually add the files?

Do you mean, those which are in github release, right? This is implemented at `appveyor.yml` build script:
```
    if($IsLinux) {
      echo gothic-linux-v$env:APPVEYOR_BUILD_VERSION | Out-File "VERSION" -Encoding utf8

      $NAME = $BASE_NAME + '_linux.zip'
      7z a $NAME    VERSION
      7z a $NAME    "*.so"
      7z a $NAME    "Gothic2Notr"
      7z a $NAME    "Gothic2Notr.sh"
      }
```

[...]

PR #875, opened 2026-01-22
Add Debian Packaging Support / Use Github Actions

versable:
Hey @Try,

As discussed in https://github.com/Try/OpenGothic/issues/872, here is the initial deb build with AppVeyor support. (Not sure if it runs/works, we'll see)

The DEB installs the following files:

-  /usr/games/Gothic2Notr
-  /usr/lib/x86_64-linux-gnu/opengothic/libTempest.so
-  /etc/ld.so.conf.d/opengothic.conf

The directory `/usr/games` was chosen according to the [Filesystem Hierarchy man page](https://www.man7.org/linux/man-pages/man7/hier.7.html) (And because my [Sauerbraten](http://sauerbraten.org/) binary resides there as well 😀):

```
       /usr/games
              Binaries for games and educational programs (optional).
```

There is no more need for the LD_LIBRARY_PATH overwrite, as this is now handled with the ld.so.conf.d (runtime discovery)

As for the DEB AppVeyor build, it is handled in a different build VM with the flag `BUILD_DEB` set to true ([Separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,11 +19,11 @@ environment:
       VCVARSALL:  "C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Auxiliary/Build/vcvarsall.bat"
       PLATFORM:   x64
 
-    - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2404
+    - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2204
       GENERATOR:  ""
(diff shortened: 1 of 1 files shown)


versable:
If #877 gets merged, this needs to be revised/changed (specifically the LD hook.)

versable:
Deprecated PR

PR #879, opened 2026-01-23
Allow detection of CPU arch for GNU/CLang tools

carlo-bramini:
The macros `_M_AMD64`, `_M_IX86` and `_M_ARM64` are available into MSVC.
However, these macros are not available into GNU tools like MinGW-w64.
This patch doesn't modify any function, it just check which CPU architecture is used on Windows, by also adding the identifiers used by GCC.
To avoid some duplications, the checks into the code are now done by just using:
- `EDD_DBG_X86_64` for AMD64 and Intel x86-64 CPUs.
- `EDD_DBG_X86` for 80x86 CPUs.
- `EDD_DBG_ARM64` for Aarch64 CPUs.

Tested with MinGW-w64 tools with x86_64, i686 and aarch64 cross compiler.

### DIFF
--- a/lib/edd-dbg/src/windows/frames.cpp
+++ b/lib/edd-dbg/src/windows/frames.cpp
@@ -12,7 +12,13 @@
 
 #include <windows.h>
 
-#if !defined(_M_AMD64) && !defined(_M_IX86) && !defined(_M_ARM64)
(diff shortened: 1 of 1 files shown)


Try:
Hi, @carlo-bramini and thank for PR!

Looks good - merged. Hopefully, soon we will be capable drop this ancient crash-logger and keep only `std::stacktrace`, but not today :)

Issue #304, opened 2022-09-12
Suggestion: Add a .clang-format file


lmichaelis:
While working on #271, I've missed being able to automatically format code to the point that I sometimes just didn't do it. Having a `.clang-format` file [like the one I have in phoenix](https://github.com/lmichaelis/phoenix/blob/main/.clang-format) would make contributing easier and get rid of any ambiguity concerning formatting. You can even test out all the parameters it provides in [an online tool](https://zed0.co.uk/clang-format-configurator/) :)

lmichaelis:
Yeah I think that's not possible. But changing that part of the code style shouldn't be too bad since it mostly just affects lines with closing braces, right? I'm not aware of too many cases where block closing braces are on the same line as some other code.

Reveares:
I also think that it would be useful to add a `.clang-format`, just because CLion does not handle the current style so well.

> And changing code-style to else, mean one huge commit that will destroy git-blame :/

This problem is relatively easy to fix, this article describes how https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

Try:
Good point, thanks @Reveares !

There is one more thing I'm wondering about:
Maybe there a way to instruct `clang-format` to format now all things, but few selected ones? Reason for that: I'm not ready drop habit of  unusual code-style for brackets

CReimer:
+1 for switching to a more common code style with proper support in IDEs. C++ is difficult enough for me right now. Having to fight against CLion every time I hit Enter is painful.

lmichaelis:
> Good point, thanks @Reveares !
> 
> There is one more thing I'm wondering about: Maybe there a way to instruct `clang-format` to format now all things, but few selected ones? Reason for that: I'm not ready drop habit of unusual code-style for brackets

@Try I just found this in the clang-format docs: It is possible to use [a Git plugin](https://clang.llvm.org/docs/ClangFormat.html#git-integration) which only touches lines changed in Git. Visual Studio probably has a setting to that effect as well.

You could also just [manually run it](https://clang.llvm.org/docs/ClangFormat.html#standalone-tool) on files you're comfortable with.

CReimer:
I found out @Try s brace style is called Ratliff style. The problem is: Nobody wants that brace style.
There is only one tool which supports this brace style. It's called astyle. But it doesnt really work for me either. A lot of other stuff is not supported or just ignored

[...]

PR #885, opened 2026-01-25
Clang format

versable:
Hey @Try ,

it would be wise to offload the code style to a formatter/linter. Here is a WIP implementation with mainwindow.cpp as a test case for automatic formatting.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,75 @@
+Language: Cpp
+Standard: c++17
+
+# Indentation
+IndentWidth: 2
+TabWidth: 2
+UseTab: Never
+AccessModifierOffset: -2
+IndentCaseLabels: true
+NamespaceIndentation: None
+IndentPPDirectives: BeforeHash
+
+# Braces
+BreakBeforeBraces: Attach
+AllowShortBlocksOnASingleLine: Empty
+AllowShortFunctionsOnASingleLine: Empty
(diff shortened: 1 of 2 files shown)


Try:
Afaik current code-style is not represented in clang-format, would it be possible to format other things, while ignoring brackets `{ }` ?

versable:
@Try I'll look into it.

Try:
Linking to prior art: https://github.com/Try/OpenGothic/issues/304

LordKobra:
Regarding brackets. The current style is called "Ratliff" and is quite unconventional (i find i a little confusing, because the closing bracket stays indented). Usually either K&R or Allman brackets are used.
https://en.wikipedia.org/wiki/Indentation_style

But if you really want to keep this bracket style, clang-format does have BS_Custom style, where you can manually define the brackets the way you want.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

Edit: So i tried to get it to work, but sadly it is not possible, because `IndentBraces` doesn't properly apply to all braces.
```yaml
BreakBeforeBraces: Custom
BraceWrapping:
  AfterFunction: false
  AfterClass: false
  AfterStruct: false
  AfterEnum: false
  AfterUnion: false
  BeforeElse: false
  BeforeCatch: false
  IndentBraces: true
```
So the only way to indent braces at the moment is to use Whitesmiths brace style, which in turn disables customization for the opening bracket.

PR #897, opened 2026-02-23
Fix journal highlighting

pancakedevourer:
The journal highlighting wouldn't update before, it would just stay on the first entry, now it does :D


pancakedevourer:
Cleaned it up :D

Try:
Merged, thanks!

pancakedevourer:
Thank you very much :D

PR #917, opened 2026-04-12
downscale preview image for faster save

jairus-kope:
Hi folks,
I really like what you did here with this game porting! amazing work!
While playing it on my macos I noticed the save was taking quite a while. about 7 seconds each time. which was an issue for me since I save quite a lot..
I took a glimpse at the code this morning and after some testing I believe I found the culprit. It looks like the image preview saved in a saving slot is what causes this, as I'm running on a high resolution (3008*1692) and this creates a large pixel buffer taking about 30mb. So I went ahead and reduced the resolution size (to 640x480) to see if this will solve it and I think it did. The save is much faster now and it also got the load/save screen to load faster as the images it contains are smaller. To test this I replaced my existing saves by loading and re-saving them with the updated code.
I suppose this will affect those who run the game on high resolutions.
Feel free to review or use this PR as you see fit.
Regards, jairus.

### DIFF
--- a/game/mainwindow.cpp
+++ b/game/mainwindow.cpp
@@ -28,38 +28,6 @@
 
 using namespace Tempest;
 
-namespace {
-  Pixmap downscaleSavePreviewPixmap(Pixmap src) {
-    if(src.isEmpty())
-      return src;
-
-    constexpr uint32_t kThumbW = 1024;
-    const uint32_t w = src.w(), h = src.h();
-    if(w <= kThumbW)
-      return src;
-
-    const uint32_t nh = std::max(1u, uint32_t((size_t(h) * kThumbW) / w));
-    const size_t bpp = src.bpp();
-
-    Pixmap out(kThumbW, nh, src.format());
-    const uint8_t* s = static_cast<const uint8_t*>(src.data());
-    uint8_t* d = static_cast<uint8_t*>(out.data());
-    
-    // Copy source pixel map to reduced pixel map using Nearest-Neighbor interpolation
-    for(uint32_t y=0; y<nh; ++y) {
-      const uint32_t sy = uint32_t((size_t(y)*h) / nh);
-      const uint8_t* srcRow = s + (size_t(sy) * w * bpp);
-      uint8_t* destRow = d + (size_t(y) * kThumbW * bpp);
-        
-      for(uint32_t x=0; x<kThumbW; ++x) {
-        const uint32_t sx = uint32_t((size_t(x)*w) / kThumbW);
-        memcpy(destRow + x*bpp, srcRow + sx*bpp, bpp);
-        }
-      }
-    return out;
-    }
-  }
-
 MainWindow::MainWindow(Device& device)
   : Window(Maximized),device(device),swapchain(device,hwnd()),
     atlas(device),renderer(swapchain),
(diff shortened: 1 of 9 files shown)


Try:
Hi, @jairus-kope and thanks for PR!

This issue been raised in the past at least one time. However it wasn't concluded that small-res is explicitly better or worse, and there was no support for configs(still no support).
Although 4k is indeed overkill.

> about 7 seconds each time.

What are time-savings, with your change?

game/mainwindow.cpp
```diff
@@ -28,6 +28,34 @@
 
 using namespace Tempest;
 
+namespace {
+  Pixmap downscaleSavePreviewPixmap(Pixmap src) {
+    if(src.isEmpty() || src.format()!=TextureFormat::RGBA8)
+      return src;
+    constexpr uint32_t kThumbW = 640, kThumbH = 480;
```
Try: 480 is overkill in term of how small image. 1440p, maybe?
Note: images can be extracted as screen-shoots.

game/mainwindow.cpp
```diff
@@ -28,6 +28,34 @@
...
+    const uint32_t w = src.w(), h = src.h();
+    
+    if(w<=kThumbW || h<=kThumbH)
+      return src;
+
+    Pixmap out(kThumbW, kThumbH, TextureFormat::RGBA8);
```
Try: This code doesn't preserve proportions. 640x480 is 4:3 ratio, but most screens today is 16:9.

> TextureFormat::RGBA8

While today it's rgba8, but there is HDR support in works - so, you should not relay strongly on format. Probably downsampling image on GPU would make more sense, to take care of formats and performance.

game/mainwindow.cpp
jairus-kope: This change does not affect screen-shoots (F11 key I believe) which will keep as is.
This is only for the thumbnails in the load/save screen. images there are a small part of the screen so I wouldn't think they need a high resolution. anyways I've changed it to 1024 (maybe take this from ini settings?)
Regarding timings (I tested this manually with my phone's stopwatch so not that accurate..):
without this change it produced a 6K HDR (6016x3384), though my screen resolution is 3008x1692 - save took ~7 seconds
1440p took ~1.5 seconds
640x480 took ~0.3 seconds (keeping aspect ration of 16:9 will probably take less)
1024x768 took ~0.5 seconds
1024x576 ie. keeping 16:9 aspect ratio (using the latest code changes i pushed) - took about i guess ~0.4 seconds
Note that these changes also totally remove the lag of the load/save screen loading.

game/mainwindow.cpp
jairus-kope: Good points. I modified the code to preserve aspect ratio and also the original texture format. Though I don't have the means to test it on HDR.
yeah I agree. downsampling on GPU would probably be best. Feel free to use this as a temp fix or just PoC

game/mainwindow.cpp
Try: > This is only for the thumbnails in the load/save screen

Yes; what I've been pointing to is fact that screen-shoot can be extracted from the save as *.png.
For move precise time you can leverage `Application::tickCount` utility.

game/mainwindow.cpp
Try: Will it be OK, to you if I'll implement the feature on my own, with gpu-support and you your data only as reference?
Asking, since merging first and do full rewrite a week after is also not ideal.

jairus-kope: Sure. please go ahead. just happy to have assisted with finding and debugging this issue. I think solving it will definitely improve the gaming experience.

Try:
Since no objections - merged :)

PR #937, opened 2026-05-18
Nix dev flake

pancakedevourer:
Added a Nix dev flake, for muuuch easier compilation on NixOS (or other systems with Nix).
Also updated the README to mention using the flake :)


Try:
Hi, @pancakedevourer and thanks for PR!
Sorry, github is very laggy for me right now, I'll have a proper look in maybe day/two.

Small question for now, can flake files be located in a folder(`./linux/flake`), to avoid crowd in root directory?

pancakedevourer:
Oh sure, that's a good idea :)
Also, compiling stuff with many libraries on NixOS, without a flake, is quite hard, because it puts everything in /nix/store, and no compiler/build system will ever find them. And a flake such as this is very convienient, as one also oesn't have to install the dependencies system-wide :)

linux/flake/flake.nix
```diff
@@ -0,0 +1,33 @@
...
+            git
+            cmake
+            glslang
+            addDriverRunpath
+          ];
+          buildInputs = with pkgs; [
+            alsa-lib
+            libX11
+            libXcursor
+            libglvnd
```
Try: I don't think we need `libglvnd`. There is no opengl support after all...
PS:
just noticed, that it's mentioned in the readme - probably by mistake

Try (approved):
LGTM

Small question: Am I correct to assume that this setup is essentially set-and-forget kind, and wont require adjustments on new releases?

linux/flake/flake.nix
pancakedevourer: Lol :D, I'll remove it. Quite definitely won't require adjustments :)

Try:
Merged, thanks!

PR #941, opened 2026-05-26
Fix Gothic 1 underwater diving

flatsponge:
Right now diving in Gothic 1 fully breaks the game and the screen goes black. Even going back up breaks it and the screen is still black.

Fixed it by:
- handling Gothic 1 diving with the swim camera instead of the Gothic 2 dive camera
- stopping the camera from following dive pitch in Gothic 1
- making underwater detection use the actual dive/swim state instead of getting stuck from previous camera checks
- resetting vertical dive movement when leaving dive mode
- not drawing the water surface while the camera is underwater
- adding the proper underwater overlay/texture pass so underwater view renders normally instead of blacking out

Basically, G1 and G2 handle diving differently, but we were treating G1 like G2. This PR separates that behavior and makes underwater movement/rendering work again in G1. 

Before:

https://github.com/user-attachments/assets/b27f1516-faa6-439d-9ee7-4f7493444f38

After:

https://github.com/user-attachments/assets/1ef6d7d5-1329-4e7b-ba11-e0d65b3efb8d

### DIFF
--- a/game/camera.cpp
+++ b/game/camera.cpp
@@ -755,16 +755,15 @@ void Camera::tick(uint64_t dt) {
     auto  pl     = isFree() ? nullptr : world->player();
     auto& physic = *world->physic();
 
-    if(pl!=nullptr && pl->isSwim()) {
-      inWater = (angles.x < -8.f);
+    if(pl!=nullptr && (pl->isSwim() || pl->isDive())) {
+      inWater = pl->isDive();
       }
     else if(pl!=nullptr && (pl->isInAir() || pl->isJump()) && !pl->isDead()) {
       // NOTE: not quite correct
       inWater = physic.cameraRay(inter.target, origin).waterCol % 2;
       }
     else {
-      // NOTE: find a way to avoid persistent tracking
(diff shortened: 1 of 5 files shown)


Try:
HI, @flatsponge !

> Even going back up breaks it and the screen is still black.

I've checked this issue locally - it's division by zero, in camera. This snippet should be enough:
```
diff --git a/game/camera.cpp b/game/camera.cpp
index 9603e6a7..e5dc67b8 100644
--- a/game/camera.cpp
+++ b/game/camera.cpp
@@ -207,7 +207,9 @@ void Camera::setMode(const Camera::Mode m) {

   if(reset) {
     state.range = def.best_range;
-    userRange   = (def.best_range - def.min_range)/(def.max_range - def.min_range);
+    if(def.max_range > def.min_range+0.01)
+      userRange = (def.best_range - def.min_range)/(def.max_range - def.min_range); else
+      userRange = 0;
     state.spin  = Vec3(0);
     }
```
Basically, it means core assumptions of a path are wrong

flatsponge:
Oh! Thanks for the quick review! 🥰

PR #947, opened 2026-06-21
quick implementation of animation adjustment

bumpah:


game/game/gamescript.cpp
```diff
@@ -1479,25 +1479,35 @@ Npc* GameScript::findNpc(zenkit::DaedalusSymbol* s) {
...
-  assert(handle->user_ptr); // engine bug, if null
-  return reinterpret_cast<Npc*>(handle->user_ptr);
+  return ptr;
   }
 
 Npc* GameScript::findNpcById(const std::shared_ptr<zenkit::DaedalusInstance>& handle) {
   if(handle==nullptr)
     return nullptr;
   if(auto npc = dynamic_cast<const zenkit::INpc*>(handle.get())) {
-    assert(npc->user_ptr); // engine bug, if null
```
Try, line 1499: This is where codex can't do good enough job today: it's a contract, that pointer must be never be null. If it is - that indicated error in other part of code

PR #950, opened 2026-06-27
Ghidra parity tooling

Alino:
I used Ghidra + Ghidra MCP + claude to extract all gothic2.exe functions and map them to what already exists in this repo and check for logic parity. This might be useful.


Try:
Hi, @Alino !

Unfortunately I would have to delete this PT, since no one allowed us to do direct reverse engineering. All game code so far - 'reconstruction', basically putting together assets/scripts to achieve same gameplay

Issue #813, opened 2025-10-17
L’Hiver Mod: Monster HP sometimes increases during a fight


AlexJakeGreen:
When attacking a monster, its health occasionally increases instead of decreasing. This issue has been observed while playing with the L'Hiver mod, and it's unclear whether this behavior is caused by the mod itself.
Reproducibility is inconsistent, and it seems to happen randomly during combat -- I had to fight he wolf 3 times when recorded a video.

This behavior can be seen in the attached video at 00:20 and 00:25 timestamps -- the monster gains HP

It would be helpful to clarify whether this is expected behavior in the original engine or a bug introduced by the mod.

[save_slot_1.sav.gz](https://github.com/user-attachments/files/22964454/save_slot_1.sav.gz)

https://github.com/user-attachments/assets/39c73380-8fed-4c8d-b3ca-81502c9c6046

ps I saw a message about a similar issue in #795:
> hitting some monsters will randomly cause damage or give back energy, e.g. bloodflies. This could be connected to the LHiver mechanic of hitting an enemy in the back

I reviewed the video, and it looks to me that there were 2 hits in wolf's back. When the wolf moves to the left or right at 00:20 and 00:25, it exposes its back, and (just a guess) incoming attacks hit it from behind

Try:
Hi @AlexJakeGreen !

I've managed to reproduce this issue. Increase to HP is incurred by script `B_MM_ASSESSDAMAGE`. Decompiling of that script shows:
```
if (OTHER.ID == 0) {
  if (!NPC_CANSEENPC(SELF, OTHER) && NPC_GETDISTTONPC(SELF, OTHER) < 1850 && (NPC_ISINSTATE(SELF, ZS_ATTACK) || NPC_ISINSTATE(SELF, ZS_MM_ATTACK))) {
    if (SELF.LEVEL - OTHER.LEVEL > 20) {
      NPC_CHANGEATTRIBUTE(SELF, ATR_HITPOINTS, 5);
    } else if (SELF.LEVEL - OTHER.LEVEL > 15) {
      NPC_CHANGEATTRIBUTE(SELF, ATR_HITPOINTS, 10);
    } else if (SELF.LEVEL - OTHER.LEVEL > 10) {
      NPC_CHANGEATTRIBUTE(SELF, ATR_HITPOINTS, 15);
    } else {
    NPC_CHANGEATTRIBUTE(SELF, ATR_HITPOINTS, 20);
   };
  };
};
```

[...]

PR #953, opened 2026-07-22
Fix L’Hiver Mod: Monster HP sometimes increases during a fight

blackus3r:
This should fix the issue described here: https://github.com/Try/OpenGothic/issues/813
I've tested it and it worked with L'hiver + Uriziel Mod. 

Fix Npc_CanSeeNpc logic to use logical orientation instead of visual bone rotation

Previously, `Npc_CanSeeNpc` (via `Npc::canRayHitPoint`) relied on `visual.viewDirection()`, which calculates the viewing angle dynamically from the `BIP01` root bone. This caused the perceived line of sight to fluctuate significantly during combat animations (e.g., when a character turns sideways to dodge or stumble).

As a result, mods relying on line of sight checks during combat (like the anti-exploit healing mechanic in L'Hiver) would break, causing enemies to regenerate mid-combat because they temporarily "lost sight" of the player due to their own animation.

Using the NPC's logical `angle` instead prevents these animation-based FOV blindspots and aligns the engine's behavior correctly with original ZenGin logic.

### DIFF
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -2296,16 +2296,22 @@ void Npc::tickTimedEvt(Animation::EvCount& ev) {
   }
 
 void Npc::tickRegen(int32_t& v, const int32_t max, const int32_t chg, const uint64_t dt) {
-  uint64_t tick = owner.tickCount();
+  const uint64_t tick = owner.tickCount();
   if(tick<dt || chg==0)
     return;
-  int32_t time0 = int32_t(tick%1000);
-  int32_t time1 = time0+int32_t(dt);
+  // ATR_REGENERATE* is an interval: 1 point per 'chg' seconds.
+  // treating it as points-per-second made monsters in mods with hp-regeneration (L'Hiver)
+  // heal faster than the player can deal damage
+  const uint64_t amount = chg>0 ? uint64_t(chg) : uint64_t(-int64_t(chg));
+  const uint64_t period = amount*1000u;
+  const uint64_t prev   = tick-dt;
+  const uint64_t n      = tick/period-prev/period;
+  if(n==0)
+    return;
 
-  int32_t val0 = (time0*chg)/1000;
-  int32_t val1 = (time1*chg)/1000;
-
-  int32_t nextV = std::max(0,std::min(v+val1-val0,max));
+  const int64_t step  = chg>0 ? int64_t(n) : -int64_t(n);
+  const int64_t upper = std::max<int64_t>(0,max);
+  int32_t nextV = int32_t(std::clamp<int64_t>(int64_t(v)+step,0,upper));
   if(v!=nextV) {
     v = nextV;
     // check health, in case of negative chg
--- a/game/world/objects/npc.cpp
+++ b/game/world/objects/npc.cpp
@@ -4668,16 +4668,26 @@ bool Npc::canRayHitPoint(const Tempest::Vec3 self, const Tempest::Vec3 pos, floa
   if(qDistTo(pos)>range*range)
     return false;
 
-  static const double ref = std::cos(100*M_PI/180.0); // spec requires +-100 view angle range
+  // The vector below points from the target back to this NPC. A >=100 degree separation from
+  // that reverse vector is the intended <=80 degree front arc.
+  static const double ref = std::cos(100*M_PI/180.0);
   const DynamicWorld* w   = owner.physic();
   bool freeLos = angOverride>=180.f;
   if(freeLos) {
     return !w->ray(self, pos).hasCol;
     }
 
-  float dx  = self.x-pos.x, dz=self.z-pos.z;
+  // cone from the object position, not the animated head bone: in melee the head dives into
+  // the target and the direction to it collapses to noise. a false negative here reads to some
+  // mods as a point-blank stealth hit and triggers a monster self-heal.
+  float dx  = x-pos.x, dz = z-pos.z;
   float dir = angleDir(dx,dz);
-  float da  = float(M_PI)*(visual.viewDirection()-dir)/180.f;
+
+  // Use the logical heading during normal movement/combat. While attached to a MOBSI the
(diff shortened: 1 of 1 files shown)


Try:
Hi, @blackus3r and thanks for PR!

Use of head orientation is intended in this place. In original scripts, when npc sits on a chair/bench his face is oriented opposed to logical orientation, and then if you talk to them this script runs:
```c++
		if (Npc_CanSeeNpc(self, other))
		{
			AI_StartState 	(self, ZS_Talk, 0, ""); 	// keep talking while sit
		}	
		else 
		{
			...
			AI_StartState 	(self, ZS_Talk, 1, ""); 	// standup and turn around
		};
```

blackus3r:
Hey @Try, 
thanks for the important heads up! 
That explains why the visual orientation was originally used in this place.
To preserve this intended behavior for chairs/benches while still fixing the combat regeneration bug in the L'Hiver mod, I've added a check for interactive objects.

What do you think about this hybrid approach?

Edit: Okay, whatever. The bug is back. I need to do more testing. I'll get back to you.

Try:
I'll move PR into drafts for now.

one theory: maybe in L'Hiver script never meant to reach `Npc_CanSeeNpc` check, and was suppose to early-out earlier

blackus3r:
@Try 

I built and tested another fix last night. However, it was very late, so I'll continue testing over the next few days to see if the bug has been fixed.
In any case, I've gotten much closer to identifying the problem thanks to additional logging: 

`Npc_CanSeeNpc` computed the view-cone direction from the NPC's head bone to the target. In melee, an attacking monster's head dives into the player model — the measured distance drops to ~35 cm and the direction vector becomes numerical noise. The ±100° view-angle check then fails randomly, even with the player directly in front and a clear line of sight. L'Hiver's anti-exploit script interprets `Npc_CanSeeNpc == false` as "player ran away", heals the monster (+20 HP per trigger) and drops it out of its attack state.

Fix 1: Measure the view cone from the NPC position instead of the head bone — collision keeps positions ~1 m apart in melee, so the direction stays stable. The line-of-sight ray still starts at the head bone. Matches original engine behavior.

Fix 2 : ATR_REGENERATEHP/MANA was treated as points per second; the original engine defines it as 1 point every N seconds. A value of 20 healed 20 HP/s instead of 1 HP per 20 s. Not the cause for the wolf (attribute was 0), but would make any monster that uses it effectively unkillable.

Verified: Debug logging showed 14× script heals triggered by failed sight checks after the fix, wolf fights behave normally.

Try:
> Fix 1: Measure the view cone from the NPC position instead of the head bone
> Matches original engine behavior.

Hm, look like a hack rather: it's very counter-intuitive to use head-bone as a reference for view direction, but logical position for view origin. When you mention that it matches original behavior, did you had some specific test setup?  

> Fix 2 : ATR_REGENERATEHP/MANA was treated as points per second

That looks solid.
Should we maybe merge regen, only for now?

blackus3r:
I phrased that too strongly. I didn’t compare it against the ZenGin binary, and I don’t have a proper setup to test the original engine. So this is based on debugging and observed behavior, not on a direct verification.

What led to the change was a debug log from a wolf fight in L’Hiver + Uriziel. In melee, the attacker’s head bone can move directly into the target. That makes the head-to-target vector extremely short, so the calculated angle is basically just numerical noise. The ±100° check then fails more or less randomly.

L’Hiver’s anti-exploit script treats that failed sight check as if the player had broken line of sight or run away, so it heals the enemy. When the cone is calculated from the body position instead, the vector stays stable and the check no longer flickers.

About the head-bone/position inconsistency: in the current version, combat no longer uses the head bone for the facing direction at all. It now uses the logical angle, which is also what the AI and turn logic use as the actual facing direction.

The head bone’s `viewDirection()` is only used when `interactive() != nullptr`, for example during a seated MOBSI dialog. In that case, the animation can physically rotate the model away from its logical angle.

So in combat it is position plus logical angle. Only for MOBSI it is position plus the actual head direction. If that is still too much for one PR, I can also remove the MOBSI special case and keep only the combat change.

Try:
Sorry, but at this point you essentially introducing more and more complexity to fix case, that may not even be broken.
If you very sure, that MOBSI is special, or raycast logic, or angle logic - it has to have sufficient evidence from base game.

In current PR, ray-cast origin is still based of head-bone, while angles computes from `pos` - ground position. Ground position is specific to opengothic, and unlikely to be a thing in original game.

In my test, with `marvin+g`, it seem hostile nps can detect you even if player  is stand behind:
image

blackus3r:
The behind-detection observation is reproducible, but it is unrelated to this change. The directional Npc_CanSeeNpc(..., false) check returns false for a player placed 180° behind the Orc. The player is instead acquired by the pre-existing updateNearestEnemy() path, which calls canSenseNpc(enemy, true). That enters the freeLos early return before the modified cone-origin and heading calculations and deliberately performs only an unobstructed ray check without an angular restriction.

**reverted fix:**

https://github.com/user-attachments/assets/18f379de-1bed-4a11-9420-fc9581d65d8a



**with fix:**

https://github.com/user-attachments/assets/849e504f-6864-49da-b41c-0a8a3ef1ffcb

blackus3r:
I made my changes in my develop branch and wrote these automated tests: https://github.com/blackus3r/OpenGothic/commits/develop/

Try:
> The behind-detection observation is reproducible,

This is video of opengothic, not of original game. Problem here is: this patch makes massive assumption, about how original may function, only to workaround issue a mod, that might be have its own bugs in the first place.
In theory: build a custom mod that can log and verify all those corner cases will help progress, but no one going to do it any time soon.

blackus3r:
Yes, that's right.
To what extent are you actually allowed to analyze and replicate the original engine?

Try:
> To what extent are you actually allowed to analyze and replicate the original engine?

Mod-kit and its documentation is fine; A:B tests of how mods behave also fine.
Union - gray area.
Direct reverse engineering - strong no.

PR #957, opened 2026-07-29
CMakeLists.txt proposed fixes for building on Gentoo Linux

Anoncheg1:
Hi! I created Ebuild file for Gentoo Linux with instruction to compile package from sources.

https://github.com/Anoncheg1/anonch-overlay/blob/main/games-engines/opengothic/opengothic-1.0.3549.ebuild

it is for Dec 23 2025 https://github.com/Try/OpenGothic/commit/abf079ead8edf1c4c4a9bad11508680fdb94d7ae

It is a recommendation, not tested, every change should be applied manually and separately.

### Motivation
Linux distributions (e.g., Gentoo, Arch, NixOS) strongly prefer linking against system-installed libraries for security updates, binary deduplication, and easier maintenance. Currently, packaging OpenGothic requires maintainers to apply fragile `sed` patches to strip out bundled submodules, which is prone to breaking on upstream updates.

### Changes
This PR introduces three new CMake options, all defaulting to `OFF` to preserve the exact existing behavior for end-users building from source:
- `USE_SYSTEM_ZENKIT`: Uses system `zenkit` and `squish` libraries.
- `USE_SYSTEM_TEMPEST`: Uses system `Tempest` and `openal` libraries.
- `USE_SYSTEM_BULLET`: Uses system `bullet` libraries via `find_package(Bullet)`.

### Benefits
1. **Zero impact on default builds**: Regular users cloning the repo and running `cmake ..` will experience no changes.
2. **Clean packaging**: Distributors can now simply pass `-DUSE_SYSTEM_ZENKIT=ON -DUSE_SYSTEM_TEMPEST=ON -DUSE_SYSTEM_BULLET=ON` to CMake, eliminating the need for ebuild/PKGBUILD patching.
3. **Proper dependency chaining**: The system library targets correctly declare their `INTERFACE_LINK_LIBRARIES` (e.g., system `zenkit` links to `squish`, system `Tempest` links to `openal`), preventing "DSO missing from command line" linker errors.
4. **Code cleanup**: Removed accidental empty `if()` blocks that were left behind in the Bullet section.

*(Note: There is also a minor C++ deprecation warning in `game/world/objects/interactive.cpp` regarding `inter.state`. I recommend changing it to `inter.state_count` to ensure clean builds on modern compilers, but I have kept this PR strictly focused on CMake.)*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,17 +99,34 @@ elseif(UNIX)
 endif()
 
 # ZenKit
-set(BUILD_SQUISH_WITH_SSE2     OFF CACHE INTERNAL "")
-set(ZK_ENABLE_ASAN             OFF CACHE INTERNAL "")
-set(ZK_ENABLE_DEPRECATION      OFF CACHE INTERNAL "")
-
-add_subdirectory(lib/ZenKit)
+# CHANGE: Added USE_SYSTEM_ZENKIT option to allow linking against system-installed ZenKit.
+option(USE_SYSTEM_ZENKIT "Use system ZenKit instead of bundled submodule" OFF)
+if(USE_SYSTEM_ZENKIT)
+  find_path(ZENKIT_INCLUDE_DIR NAMES zenkit/World.hh)
+  find_library(ZENKIT_LIBRARY NAMES zenkit)
+  find_library(SQUISH_LIBRARY NAMES squish)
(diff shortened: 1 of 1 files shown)


Try:
close due to:

> It is a recommendation, not tested, every change should be applied manually and separately.

PR #968, opened 2026-08-20
fix(compatibility): include <unordered_map> where it is used

mikolajmikolajczyk:
libstdc++ includes unordered_map via functional.
libc++ does not. This causes compilation to fail on libc++ targets.


Try:
Merged, thanks!

PR #969, opened 2026-08-20
fix(compatibility): construct string_view by length, not by iterator pair

mikolajmikolajczyk:
basic_string_view(It first, End last) is C++20 and libc++ 11 does not have it. The call fails with "no matching constructor".
The length form is C++17, exists in both libraries, and denotes the same view - ptr..ptr+size is a range of length size.


Try:
Merged, thanks!

PR #970, opened 2026-08-20
fix: removed unused Read::from

mikolajmikolajczyk:
`in` is never used, and `mount_disk` on the next line opens the same path anyway.

It costs a full read of every archive, discarded. In a Gothic 2 NotR installation that is 2.51 GiB, read and thrown away on every start. I only noticed this because PS4 was hanging when executing this call.


Try:
merged, thanks!

PR #971, opened 2026-08-20
fix(sound): memset sizes the silence buffer with the wrong element type

mikolajmikolajczyk:
The element size is settled by the format.
 Tempest declares: alBufferCallbackDirectSOFT is given AL_FORMAT_STEREO16 or AL_FORMAT_MONO16.


Try:
I've changed path slightly: `int16_t` -> `out[0]`, so we don't name same type twice

Try:
Merged, thanks!

PR #976, opened 2026-08-27
Wip/spacer gizmo selection

Raven358:
**Description:**

> Initial WIP implementation for the Spacer editor.
>
> Includes:
>
> * object selection using actual mesh geometry rather than only bounding boxes;
> * Im3d transform gizmo;
> * selection outline;
> * required editor shaders and integration.
>
> This is an experimental implementation and still needs cleanup/review. A significant part of the initial code was AI-assisted/generated.
>
> Known issues:
>
> * gizmo manipulators can stretch when close to the screen edge;
> * selection outline may create visual noise on textures with an alpha channel when viewed very closely.


Try:
Hi, @Raven358 and thanks for PR!

As we discussed on discord, I'm not going to take patch as-is, but rather use it as a base to rewrite optimal implementation.
Here current state of selection (also property view is updates accordingly)
image

PR #980, opened 2026-09-03
Fix cursor visibility on startup and focus

Abendlied:
see Tempest change: Try/Tempest#98

Removed the old fullscreen-dependent cursor handling and replaced with a simple centering call.

### DIFF
--- a/common/mainwindow.cpp
+++ b/common/mainwindow.cpp
@@ -37,6 +37,7 @@ MainWindow::MainWindow(Device& device)
     mobileUi(player),
 #endif
     player(dialogs,inventory) {
+  setCursorShape(CursorShape::Hidden);
   Gothic::inst().onSettingsChanged.bind(this,&MainWindow::onSettings);
   onSettings();
 
@@ -294,11 +295,7 @@ void MainWindow::resizeEvent(SizeEvent&) {
   swapchain.reset();
   if(auto camera = Gothic::inst().camera())
     camera->setViewport(swapchain.w(),swapchain.h());
-
-  const bool fs = SystemApi::isFullscreen(hwnd());
-  auto rect = SystemApi::windowClientRect(hwnd());
-  setCursorPosition(rect.w/2,rect.h/2);
-  setCursorShape(fs ? CursorShape::Hidden : CursorShape::Arrow);
+  setCursorPosition(Point(w()/2,h()/2));
   dMouse = Point();
   }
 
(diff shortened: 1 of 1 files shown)


common/mainwindow.cpp
```diff
@@ -557,6 +554,7 @@ void MainWindow::keyUpEvent(KeyEvent &event) {
   }
 
 void MainWindow::focusEvent(FocusEvent &event) {
+  setCursorShape(event.in ? CursorShape::Hidden : CursorShape::Arrow);
```
Try: why toggle from `hidden` to `arrow`?

common/mainwindow.cpp
```diff
@@ -294,11 +295,7 @@ void MainWindow::resizeEvent(SizeEvent&) {
   swapchain.reset();
   if(auto camera = Gothic::inst().camera())
     camera->setViewport(swapchain.w(),swapchain.h());
-
-  const bool fs = SystemApi::isFullscreen(hwnd());
-  auto rect = SystemApi::windowClientRect(hwnd());
-  setCursorPosition(rect.w/2,rect.h/2);
-  setCursorShape(fs ? CursorShape::Hidden : CursorShape::Arrow);
+  setCursorPosition(Point(w()/2,h()/2));
```
Try: Game not meant to own mouse, when in window mode. Also `setCursorPosition` should not be called here in window mode, but that's different issue.

common/mainwindow.cpp
```diff
@@ -37,6 +37,7 @@ MainWindow::MainWindow(Device& device)
     mobileUi(player),
 #endif
     player(dialogs,inventory) {
+  setCursorShape(CursorShape::Hidden);
```
Try: Game meant to hide cursor only in full-screen mode.

Abendlied:
Before, OpenGothic left the native cursor visible ingame, even if starting fullscreen and even if no mouse movement was generated after the fullscreen transition.

>why toggle from hidden to arrow?

Doing it now without the unnecessary toggle.

>Game meant to hide cursor only in full-screen mode.

&
>Game not meant to own mouse, when in window mode. [...]

This fix now hides the cursor immediately after entering fullscreen. It only recenters the mouse during resize when the window is fullscreen. This should satisty your requirement for the window mode.

Tempest now handles X11 EnterNotify so that the hovered widget and its cursor state are evaluated when the pointer enters the game window.

Tested on Linux/X11:
1. Cursor hidden on fullscreen startup.
2. Cursor remains hidden during gameplay.
3. Cursor becomes visible when focus is lost (Alt-Tabbing).
4. Cursor becomes hidden again when focus returns, both with and without mouse movement.

Try:
I've merged, without Tempest change - as this two are not dependent on one another.

Thanks for PR @Abendlied !

PR #981, opened 2026-09-03
Extract gog assets

orodreth1315:
This PR enables extraction of Gothic 2 Vanilla Assets from **2.7 Vanilla OpenGothic 1.0.3010 GOG**
With the script, Users can extract G2 Classic Assets and use it for builds on other platforms such as macOS and Linux.
Furthermore, new versions of OpenGothic Vanilla on Windows can be created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,11 @@
 *.dylib
 *.dll
 
-# Executables
+# Binaries
 *.exe
 *.out
 *.app
+*.bin
 
 # IDE
 CMakeLists.txt.user*
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ OpenGothic is designed to utilize features of modern graphics hardware and APIs,
 
 #### Prerequisites
 
-Gothic 2: Night of the Raven is required as OpenGothic does not provide any built-in game assets or scripts.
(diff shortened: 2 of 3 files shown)


Try:
Hi, @orodreth1315 !

I don't think I see what problem you looking to solve with it. On Steam, while game is not listed for mac/linux - installing it is not an issue from the commandline.

> Furthermore, new versions of OpenGothic Vanilla on Windows can be created.

???

orodreth1315:
Hello, thank you for your reply.

I had this Problem: 
- I wanted to play Gothic 2 vanilla (without the night of the raven addon) on macOS. 
- Problem is: As I don‘t have the Game on CD anymore, I only had the option to download this from GOG: **2.7 Vanilla OpenGothic 1.0.3010 GOG** which is a Windows Installer.
- The script from this PR takes the exe and extracts the game assets of Gothic 2 vanilla. Note that other Gothic 2 Gold versions contain addon content which differs from the vanilla version and can‘t be used.
- With the assets in place, I was able to build and play the vanilla game on macOS.
- One can use this script to generate future versions of vanilla OpenGothic for Windows. That‘s what I meant with 

> Furthermore, new versions of OpenGothic Vanilla on Windows can be created.

Please let me know if anything else is unclear.

PS: Awesome work, I was able to complete the game on macOS.

Try:
I think this PR solve a very specific issue with your use-case. Otherwise installing original files on mac is not an issue

PR #982, opened 2026-09-06
Optimize Lanczos upscaling and stabilize near-zero weights

Solessfir:
- Reduce texture reads from 21 to 12 by combining compatible Lanczos weights with bilinear sampling.
- Stabilize weight calculation near zero.
- Leave native-resolution rendering unchanged.

Working on the Android port.
On a Galaxy S24 at 50% render scale, the measured tonemapping/upscale pass dropped from 2.32 ms to 0.78 ms.
This is a pass-level measurement, not an overall FPS gain.

### DIFF
--- a/common/graphics/renderer.cpp
+++ b/common/graphics/renderer.cpp
@@ -754,7 +754,9 @@ void Renderer::drawTonemapping(Attachment& result, Encoder<CommandBuffer>& cmd, 
   auto& pso = (settings.vidResIndex==0) ? shaders.tonemapping : shaders.tonemappingUpscale;
   cmd.setFramebuffer({ {result, Tempest::Discard, Tempest::Preserve} });
   cmd.setBinding(0, wview.sceneGlobals().uboGlobal[SceneGlobals::V_Main]);
-  cmd.setBinding(1, sceneLinear, Sampler::nearest(ClampMode::ClampToEdge)); // Lanczos upscale requires nearest sampling
+  const auto sampler = settings.vidResIndex==0 ? Sampler::nearest(ClampMode::ClampToEdge) :
+                                               Sampler::bilinear(ClampMode::ClampToEdge);
+  cmd.setBinding(1, sceneLinear, sampler);
   cmd.setPushData(p);
   cmd.setPipeline(pso);
   cmd.draw(nullptr, 0, 3);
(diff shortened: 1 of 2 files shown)


shader/upscale/lanczos.glsl
```diff
@@ -20,20 +17,34 @@ vec3 lanczosUpscale(in sampler2D img, vec2 coord) {
   coord      += -0.5 * resInv;
   vec2 ccoord = floor(coord * res) * resInv;
 
+  // Evaluate each one-dimensional weight once.
+  // Keep the original footprint, including its omitted four corner samples.
+  vec2 weights[5];
```
Try: should it be `r*2+1` here, instead of `5`?
depite, at this point algorithm is strictly bound to `r=2` anyway :)

common/graphics/renderer.cpp
```diff
@@ -754,7 +754,9 @@ void Renderer::drawTonemapping(Attachment& result, Encoder<CommandBuffer>& cmd,
   auto& pso = (settings.vidResIndex==0) ? shaders.tonemapping : shaders.tonemappingUpscale;
   cmd.setFramebuffer({ {result, Tempest::Discard, Tempest::Preserve} });
   cmd.setBinding(0, wview.sceneGlobals().uboGlobal[SceneGlobals::V_Main]);
-  cmd.setBinding(1, sceneLinear, Sampler::nearest(ClampMode::ClampToEdge)); // Lanczos upscale requires nearest sampling
+  const auto sampler = settings.vidResIndex==0 ? Sampler::nearest(ClampMode::ClampToEdge) :
```
Try, line 759: probably worth commenting, that upscale is not vanilla Lanczos anymore, to avoid confusion.

common/graphics/renderer.cpp
```diff
@@ -805,7 +807,8 @@ void Renderer::drawCMAA2(Tempest::Attachment& result, Tempest::Encoder<Tempest::
   auto& psoTone = (settings.vidResIndex==0) ? shaders.tonemapping : shaders.tonemappingUpscale;
   cmd.setFramebuffer({{result, Tempest::Discard, Tempest::Preserve}});
   cmd.setBinding(0, wview.sceneGlobals().uboGlobal[SceneGlobals::V_Main]);
-  cmd.setBinding(1, sceneLinear, Sampler::nearest());
+  const auto sampler = settings.vidResIndex==0 ? Sampler::nearest() : Sampler::bilinear(ClampMode::ClampToEdge);
+  cmd.setBinding(1, sceneLinear, sampler);
```
Try: upscale and AA should be mutually exclusive. Both algorithms are doing fundamentally same job of reconstructing more samples from less

shader/upscale/lanczos.glsl
Try: maybe `r*2+1`, instead of `5`? Despite algorithm is now hard-coded to `r=2` :)

Solessfir: Yep, I'll use r*2+1. The paired sampling is still specialized for r=2, so I'll make that explicit too.

common/graphics/renderer.cpp
Solessfir: Sure, I'll add a comment at the sampler binding.

common/graphics/renderer.cpp
Solessfir: They already are: aaEnabled requires vidResIndex==0.
I unnecessarily changed the sampler in drawCMAA2, where the upscale branch can't be reached.
I'll revert that part.

Try:
Merged, thanks!

PR #983, opened 2026-09-07
Support for additional keys

NHogweed:
Insert, Home, End, Page up, Page down. Works on Linux, not sure about Windows. See also [the Tempest part here](https://github.com/Try/Tempest/pull/104).


Try:
Merged, thanks!

PR #985, opened 2026-09-10
Handle DirectMusic loader creation failures safely

Solessfir:
When DirectMusic loader initialization fails, `DmLoader_create` can free its allocation while leaving the output pointer non-null. OpenGothic then passes that dangling pointer to `DmLoader_addResolver` and later `DmLoader_release`.

Clear the pointer on failure and register the resolver only after successful initialization. This prevents use-after-free during startup and shutdown.

Found while investigating a Gothic 1 crash on Quit: loader initialization failed, and shutdown crashed while destroying the VFS tree. The change eliminated that native crash in the Android build. It affects shared resource initialization and requires no Android code.

### DIFF
--- a/common/resources.cpp
+++ b/common/resources.cpp
@@ -98,9 +98,10 @@ Resources::Resources(Tempest::Device &device)
   DmResult rv = DmLoader_create(&dmLoader, DmLoader_DOWNLOAD);
   if(rv != DmResult_SUCCESS) {
     Log::e("Failed to created DmLoader object. Out of memory?");
+    dmLoader = nullptr;
     }
-
-  DmLoader_addResolver(dmLoader, [](void* ctx, char const* name, size_t* len) -> void* {
+  else {
+    DmLoader_addResolver(dmLoader, [](void* ctx, char const* name, size_t* len) -> void* {
       auto* slf = reinterpret_cast<Resources*>(ctx);
       auto* node = slf->gothicAssets.find(name);
 
@@ -118,7 +119,8 @@ Resources::Resources(Tempest::Device &device)
       reader->read(bytes, *len);
 
       return bytes;
-  }, this);
+    }, this);
+    }
   }
 
 void Resources::mountWork(const std::filesystem::path& path) {


common/resources.cpp
```diff
@@ -98,9 +98,10 @@ Resources::Resources(Tempest::Device &device)
   DmResult rv = DmLoader_create(&dmLoader, DmLoader_DOWNLOAD);
   if(rv != DmResult_SUCCESS) {
     Log::e("Failed to created DmLoader object. Out of memory?");
+    dmLoader = nullptr;
```
Try, line 101: I'm pretty sure, It's not intended for library code (`DmLoader_create`) to change `&dmLoader` on error.
So, correct way to fix it is to PR into DirectMusic.

common/resources.cpp
```diff
@@ -98,9 +98,10 @@ Resources::Resources(Tempest::Device &device)
...
     }
-
-  DmLoader_addResolver(dmLoader, [](void* ctx, char const* name, size_t* len) -> void* {
+  else {
+    DmLoader_addResolver(dmLoader, [](void* ctx, char const* name, size_t* len) -> void* {
```
Try, line 104: also not need - `DmLoader_*` meant to early out and return `DmResult_INVALID_ARGUMENT`, for null-argument.

Try (commented):
Hi @Solessfir!

I think it would be right to adjust code of dmusic rather. Also wonder, how did you hit this case? Did `mtx_lock` failed for you?

Solessfir:
Makes sense. I've prepared and committed the fix in dmusic locally: leave the output pointer untouched on failure and assign it only after initialization succeeds.

```diff
- DmLoader* new = *slf = Dm_alloc(sizeof *new);
+ DmLoader* new = Dm_alloc(sizeof *new);
```

Then add `*slf = new;` after the cache/list initialization, just before returning `DmResult_SUCCESS`. The dmusic Release build passes with MSVC.

You're right about the resolver guard too: once the loader pointer stays null, `DmLoader_addResolver` already handles it.

I found this while investigating an Android crash when quitting Gothic 1. The startup log reported loader creation failure, but didn't record the result code. The dangling-pointer path I found is the `mtx_init` failure path, not `mtx_lock`. I can't confirm what caused the original initialization failure from that log alone.

Solessfir:
Opened the dmusic fix here: https://github.com/GothicKit/dmusic/pull/14. It assigns the output pointer only after initialization succeeds. This replaces the caller-side workaround proposed here once OpenGothic updates its dmusic submodule.

Try:
I'll close this one for now - will update dmusic, once PR is merged

PR #987, opened 2026-09-12
Restore Ikarus runtime bindings after loading a save

Solessfir:
Loading a save restores script variables but leaves Ikarus's transient parser and engine bindings uninitialized.

Reproduced with Archolos: dialogue and facial-animation calls work in a new game, then fail after saving and reloading with `MEM_GetFuncID: Unresolvable request` and an out-of-bounds access to `MOB_CREATEITEMS.PAR1`.

Call `MEM_InitAll` after restoring variables and `HERO`, only for scripts using Ikarus. This does not rerun gameplay initialization, reset NPC routines, or change the save format.

Verified on Windows using the same save and `AI_ResetFaceAni` call before and after the fix. The call failed before and completed without those errors afterward. Windows and Android ARM64 release builds passed on the development branch before extracting this fix.

### DIFF
--- a/common/game/gamescript.h
+++ b/common/game/gamescript.h
@@ -75,6 +75,8 @@ class GameScript final {
     void         savePerc(Serialize& fout);
     void         loadPerc(Serialize& fin);
 
+    bool         usesMemoryExtensions() const { return dma!=nullptr; }
+
     inline auto& getVm() { return vm; }
     auto         questLog() const -> const QuestLog&;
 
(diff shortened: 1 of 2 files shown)


Try:
Hi, @Solessfir !

Thanks for PR, but I will have to reject it for now. There is no software design on save/load in DMA environment yet. One option is to do same as in regular games: stash whole memory state and transparently restore it on load. Other one - make it dirty to respect Ikarus API to work with archives.

Either way - it's probably too early to start on it, as first priority would be to have DMA work cleanly in use-cases without save/load. Part of it will include: implement `reinterpret_cast`, develop unit-tests, custom-compiler (for test scripts).

PR #989, opened 2026-09-15
Fix diving controls

NHogweed:
Forward/backward keys to control pitch and only jump key to move ahead, like in vanilla engine.

### DIFF
--- a/common/utils/keycodec.cpp
+++ b/common/utils/keycodec.cpp
@@ -42,6 +42,13 @@ std::initializer_list<KeyCodec::K_Key> KeyCodec::keys = {
   {Tempest::Event::K_8,        0x7f00},
   {Tempest::Event::K_9,        0x8000},
 
+  {Tempest::Event::K_Insert,   0xd200},
+  {Tempest::Event::K_Delete,   0xd300},
+  {Tempest::Event::K_Home,     0xc700},
+  {Tempest::Event::K_End,      0xcf00},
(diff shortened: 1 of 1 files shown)


Try (commented):
Hi, @NHogweed !

I've checked `Npc::walkMode` vs `Npc::isDive`. There is an issue, that `walkMode` not accounting dive/water states, while this patch only hides the issue.

NHogweed:
> this patch only hides the issue

Not sure what do you mean. Does an ordinary player complaining about underwater controls need anything more? Why it's so important to rely on `walkMode` here (and only here, unlike [a couple of lines below](https://github.com/Try/OpenGothic/blob/master/common/game/playercontrol.cpp#L883))?

Try:
> Not sure what do you mean.

You need to be careful with path you committing, and make sure that code you adding is something that make sense.
I think for now I'll take this PR, as a bug report rather, and will push a popper fix

NHogweed:
> You need to be careful with path you committing

This was a reason why I decided to make changes as little as possible to achieve the stated goal and not to break anything else (that can be observed by a player). Anyway, your "philosophically and aestethically correct" fix is here, so there's no point to argue about that anymore. But there are still cosmetic problems with diving: 

- when the hero is upside down in the water, it is noticeable that his height rotation axis is out of his model;
- rotation around height axis is not animated, unlike in vanilla game. The animation should be present when swimming as well.

PR #991, opened 2026-09-19
Custom Save Games Directory (`-saves`)

mztulip:
### 📁 Custom Save Games Directory (`-saves`)

New command-line option allowing you to redirect save files (`.sav`) and their UI thumbnails to a custom directory instead of the default game folder:

```bash
./Gothic2Notr -g "/path/to/Gothic II" -saves "/path/to/custom_saves"


Try:
Hi, @mztulip !

Sorry, but I'll prefer to not work on such features individually for now. Lets instead think about out-of-box experience as a whole:
* how installation mean to be like 
* how game/mod management mean to be like
* how .ini stack should work

Started a wider discussion here: https://github.com/Try/OpenGothic/issues/992

mztulip:
I will think about it.

